PR #6, opened 2022-09-07
fix model controller

anishasthana:
- Add InferenceService API to controller scheme
- Update model controller module name

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -2,5 +2,5 @@ domain: opendatahub.io
 layout:
 - go.kubebuilder.io/v3
 projectName: odh-model-controller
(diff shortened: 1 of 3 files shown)


anishasthana:
This PR should fix the following error

```
1.662477753609541e+09 ERROR setup unable to create controller {"controller": "InferenceService", "error": "no kind is registered for the type v1beta1.InferenceService in scheme \"go/pkg/mod/k8s.io/apimachinery@v0.24.2/pkg/runtime/scheme.go:100\""}
```

I'm in the process of testing out the newly built image now.

crobby (approved):

PR #7, opened 2022-09-07
Fix things

anishasthana:


OWNERS
```diff
@@ -0,0 +1,7 @@
+approvers:
+  - anishasthana
+  - VaishnaviHire
```
anishasthana, line 3: I'm adding Vaishnavi as an approver in the short term until we've figured out who else will be contributing to Model Mesh

crobby (approved):

PR #8, opened 2022-09-13
Allow ODH Model Controller to create SA and corresponding rbac for au…

anishasthana:


controllers/inferenceservice_controller.go
```diff
@@ -108,7 +115,8 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
 		Owns(&routev1.Route{}).
 		Owns(&corev1.ServiceAccount{}).
 		Owns(&corev1.Service{}).
-		Owns(&corev1.Secret{})
+		Owns(&corev1.Secret{}).
```
crobby, line 118: Remind me again why this was necessary.

anishasthana: So we are now creating a new cluster role binding for the modelmesh oauth proxy container to be able to delegate auth calls. I'm setting an owner reference so that the CRB is deleted when this odh-model-controller deployment is nuked

crobby: Maybe it's too early for me today, but what does Secret have to do with CRB?

crobby (approved):
I've seen this stuff work.  Works well, code is sensible.

PR #14, opened 2022-12-01
WIP: Add servicemonitor management to odh controller.

HumairAK:
Dependent on:  https://github.com/red-hat-data-services/odh-manifests/pull/279

This change introduces a new controller that can be enabled via a flag (--monitoring-ns) to the operator to monitor servingruntimes. The controller watches servingruntimes and the newly created servicemonitor in the designated namespace for changes and ensures it's consistently in its expected state.

Also removed unused files (namespace/serving runtime reconcile logic).

## Description

1) The decision to enable this via the flag is because we don't want to be opinionated about where the servicemonitor lives, otherwise users would have to worry about hardcoded upstream/downstream namespaces that they would need to adhere to 
2) Decision to make this into a separate controller is because service monitors are not dependent on inferenceservices (the other possible  choice of adding this logic), instead they are dependent on servingruntimes (as that is what they monitor), having the monitoring be a separate controller also enables us to easily extend monitoring capabilities in a decoupled way from the other logic
3) One potential costly procedure is we fetch all the servingruntimes on a cluster each reconcile loop, to ensure we always keep the servingmonitor namespace selector list up to date. But since we only reconcile upon changes to servingruntime/servicemonitors, this shouldn't be too bad. Still, some caching logic would probably nice to have in the future if this poses as a problem.

## How Has This Been Tested?
Locally tested, still need to test in OCP.

## Merge criteria:

- [X] The commits are squashed in a cohesive manner and have meaningful messages.
- [] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [] The developer has manually tested the changes and verified that the changes work


main.go
```diff
@@ -74,12 +76,17 @@ func getEnvAsBool(name string, defaultValue bool) bool {
...
 	var enableLeaderElection bool
+	var monitoringNS string
 	var probeAddr string
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+
+	flag.StringVar(&monitoringNS, "monitoring-namespace", "",
```
anishasthana, line 87: I'm wondering how we'd allow users to monitor their components in odh-core installs... If the ODH operator is what is installing the odh model controller, how are users going to override it?

anishasthana: Perhaps we default to creating them in the same namespace?

anishasthana: We settled on using kfdef parameters for configuration stuff.

HumairAK:
closing in favor of: https://github.com/opendatahub-io/odh-model-controller/pull/15

PR #20, opened 2023-03-14
Implement simple logic for automatic ServingRuntime selection

israel-hdez:
## Description
An InferenceService is allowed to omit specifying a runtime. This should lead to automatic selection of a ServingRuntime from an allowed pool of ServingRuntimes.

These changes implement the automatic selection of a ServingRuntime: if there are several runtimes available, order priority is by creation date (most recent first). If several runtimes have the same creation date, priority is by alphabetical order.

## How Has This Been Tested?
1. Creating one or more ServingRuntime resources, where one (or more) of theirs supportedModelFormats has `autoSelect: true`.
2. Then, create an InferenceService without specifying a `runtime`.
    * Test using a model format that would match an auto-selectable ServingRuntime
    * Test using a model format that would _not_ match any ServingRuntime

## Testing Instructions
Live build is here: quay.io/edgarhz/rhods-operator-live-catalog:1.23.0-7468

* Install rhods from the provided live build, or install the model controller from this PR.
* Create a ServingRuntime that has model formats with `autoSelect: true`
  * The [test-openvino-serving-runtime-1.yaml](https://github.com/opendatahub-io/odh-model-controller/blob/ab9590f6d9265f0fab31100610c35e971e354f34/controllers/testdata/deploy/test-openvino-serving-runtime-1.yaml#L29-L31) used in unit tests works.
* Create an InferenceService that does not specify a runtime
  * The [openvino-inference-service-no-runtime.yaml](https://github.com/opendatahub-io/odh-model-controller/pull/20/files#diff-f553173b111481bfa595c9980b989654e13d1879823f6a9bdc12a78189458d69) added in this PR works with the mentioned ServiceRuntime in the previous bullet.
* Check that a route is created correctly.
  * If you are using the two resources mentioned in the previous points, `oc get route -n default example-onnx-mnist-no-runtime` should return a result. This would indicate that the odh-model-controller is running OK (before this PR, it crashed).

## Merge criteria:

### DIFF
--- a/controllers/inferenceservice_route.go
+++ b/controllers/inferenceservice_route.go
@@ -17,10 +17,12 @@ package controllers
 
 import (
 	"context"
+	"reflect"
+	"sort"
+
+	"github.com/go-logr/logr"
 	predictorv1 "github.com/kserve/modelmesh-serving/apis/serving/v1alpha1"
 	inferenceservicev1 "github.com/kserve/modelmesh-serving/apis/serving/v1beta1"
-	"reflect"
-
 	routev1 "github.com/openshift/api/route/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -29,6 +31,7 @@ import (
 	"k8s.io/client-go/util/retry"
 	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


VedantMahabaleshwarkar (dismissed):
/lgtm

controllers/inferenceservice_route.go
```diff
@@ -93,25 +96,73 @@ func CompareInferenceServiceRoutes(r1 routev1.Route, r2 routev1.Route) bool {
...
+			if apierrs.IsNotFound(err) {
+				log.Info("Runtime specified in InferenceService does not exist", "runtime", *inferenceservice.Spec.Predictor.Model.Runtime)
+			}
+		}
+		return &desiredServingRuntime
+	} else {
+		runtimes := &predictorv1.ServingRuntimeList{}
+		err := r.List(ctx, runtimes, client.InNamespace(inferenceservice.Namespace))
+		if err != nil {
+			return nil
```
heyselbi, line 117: Could you add a log.Info so it shows why it errored out

heyselbi (changes requested):

controllers/inferenceservice_route.go
israel-hdez: @heyselbi Done!

VedantMahabaleshwarkar (approved):

Jooho (approved):

Jooho (approved):
/lgtm

anishasthana (approved):

PR #27, opened 2023-04-27
clean up manifests

Jooho:
## Description
manifests are outdated so this PR updated manifests

## How Has This Been Tested?
There is no specific test case for odh model controller so I only tested the controller is running well on the minikube.

## Merge criteria:

- [V] The commits are squashed in a cohesive manner and have meaningful messages.
- [V] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [V] The developer has manually tested the changes and verified that the changes work


config/manager/manager.yaml
```diff
@@ -1,60 +1,55 @@
...
-  namespace: system
   labels:
     control-plane: odh-model-controller
+    app: odh-model-controller
 spec:
   selector:
     matchLabels:
       control-plane: odh-model-controller
-  replicas: 1
+  replicas: 3
```
israel-hdez, line 12: Replicas changed to 3. Is this intentional?

Jooho: Yes, odh-manifests uses 3 for a long time.

israel-hdez (approved):
/lgtm

PR #33, opened 2023-05-04
Implement traffic splitting

israel-hdez:
Add support for configuring traffic splitting by using a Service Mesh

## Description
Traffic splitting support depends on Service Mesh. When a Service Mesh is enabled, VirtualServices are created to route traffic to ModelMesh and OpenShift Routes are no longer created.

It is possible to do infer requests to specific InferenceServices, like before/currently. Annotations and Labels are used to group more than one InferenceService to create a logical group where traffic splitting would be done, and to configure how much traffic should be routed to each service in the group.

## How Has This Been Tested?

TODO

## Merge criteria:

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -111,6 +111,7 @@ rules:
   verbs:
   - get
   - list
+  - update
   - watch
 - apiGroups:
   - serving.kserve.io
--- a/controllers/inferenceservice_controller.go
+++ b/controllers/inferenceservice_controller.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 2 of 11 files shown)


israel-hdez:
**Please node** that this PR is against a feature branch; not `main`.

aslakknutsen:
@israel-hdez You're creating a new VS resource for the splitting part, right? Considered reconfiguring the 'other' one instead? I'm not sure if we'll have an issue with this as of now, but while you can split the configuration into multiple resources and they will get merged into one when read by istio, the order of matches etc are taken into account but they become random if split.

israel-hdez:
@aslakknutsen Yes, this is creating a new VS for the splitting part.

> but while you can split the configuration into multiple resources and they will get merged into one when read by istio, the order of matches etc are taken into account but they become random if split

I indeed had some issues related to this while testing. I solved by making the match more strict/specific. 

Thinking a little bit. A VS is already created per InferenceService. For traffic splitting, at least two InferenceServices are need which would create two VS for the regular case. Then, another VS groups together both InferenceServices. So this confuses me a little bit, because there isn't a singe "other one". There are "many other ones".

So, should we merge everything together and have a single VS for the namespace? Or do we keep many VSs with very specific match rules?

aslakknutsen (approved):

PR #34, opened 2023-05-08
Update Dockerfile.develop for openshift-ci modelmesh-serving

Jooho:
## Description
Dockerfile.develop file in modelmesh-serving is updated. This is required to execute modelmesh-serving openshift ci


Jooho:
In order to trigger openshift-ci, I will create another PR

PR #49, opened 2023-06-07
RHODS-7318 : Parameterize Serviceaccount

vaibhavjainwiz:
https://issues.redhat.com/browse/RHODS-7318

**Solution Approach :**
Our main goal is to listen for user custom configmap changes and act on it. If a user provides its custom configmap(`model-serving-config`) or makes a change into it then we need to call `InferencesServiceReconciler-> reconcile` method which would create appropriate ServiceAccount and its ClusterRoleBinding. 

**Step 1:**  `model-serving-config` always installed inside the same namespace in which odh-model-controller runs so to load user custom configmap first we need to identify the namespace in which the controller runs.
**Step 2:** Add watcher for `model-serving-config` configMap in `InferencesServiceReconciler`.
**Step 3:** Load the serving configuration configMap data into memory of controller inside watch logic itself so that every instance of InferencesService don't need to load it independently.
**Step 4:** Load the serving configuration configMap data into memory of controller on controller startup.
**Step 5:** Ĭnside `InferencesServiceReconciler` use in memory config data to fetch ServiceAccount name. 

## Merge criteria:


main.go
```diff
@@ -74,6 +84,21 @@ func getEnvAsBool(name string, defaultValue bool) bool {
...
+
+	// ----- fetch controller namespace -----
+	controllerNamespace := os.Getenv(ControllerNamespaceEnvVar)
+	if controllerNamespace == "" {
+		bytes, err := os.ReadFile(KubeNamespaceFile)
+		if err != nil {
+			//TODO check kube context and retrieve namespace from there
+			setupLog.Info("Error reading Kube-mounted namespace file, reverting to default namespace",
+				"file", KubeNamespaceFile, "err", err, "default", DefaultControllerNamespace)
+			controllerNamespace = DefaultControllerNamespace
```
israel-hdez, line 96: I think this should be a `panic`. Although the default is fair, it is probably better to make no assumptions.

main.go
```diff
@@ -47,6 +50,13 @@ var (
 	setupLog = ctrl.Log.WithName("setup")
 )
 
+const (
+	ControllerNamespaceEnvVar  = "NAMESPACE"
```
israel-hdez, line 54: Can you use a more specific name for the env variable?
`NAMESPACE` doesn't give the idea that you want the namespace of the modelmesh-controller.

controllers/inferenceservice_controller.go
```diff
@@ -126,7 +131,37 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
...
+					r.Log.Error(err, "Unable to reload user configuration")
+				}
+				inferenceServicesList := &inferenceservicev1.InferenceServiceList{}
+				err = r.List(context.TODO(), inferenceServicesList)
+				if err != nil {
+					r.Log.Info("Error getting list of inference services in all namespaces")
+					return []reconcile.Request{}
+				}
+				if len(inferenceServicesList.Items) == 0 {
+					r.Log.Info("No InferenceServices found for Serving Runtime: " + o.GetName())
```
israel-hdez, line 148: Is `o.GetName()` a ServingRuntime?

controllers/inferecenceservice_serviceaccount.go
```diff
@@ -63,17 +59,17 @@ func createDelegateClusterRoleBinding(serviceAccountName string, serviceAccountN
...
 }
 
-func (r *OpenshiftInferenceServiceReconciler) reconcileSA(inferenceService *inferenceservicev1.InferenceService, ctx context.Context, newSA func(service *inferenceservicev1.InferenceService) *corev1.ServiceAccount) error {
+func (r *OpenshiftInferenceServiceReconciler) reconcileSA(inferenceService *inferenceservicev1.InferenceService, ctx context.Context, serviceAccountName string, serviceAccountNamespace string, newSA func(serviceAccountName string, serviceAccountNamespace string) *corev1.ServiceAccount) error {
 
 	// Initialize logger format
 	log := r.Log.WithValues("inferenceservice", inferenceService.Name, "namespace", inferenceService.Namespace)
 
-	desiredSA := newSA(inferenceService)
+	desiredSA := newSA(serviceAccountName, serviceAccountNamespace)
```
israel-hdez, line 67: Should we delete the old SA, if there is a new name?

israel-hdez (commented):
There seems to be 2 parts on the Jira:
* A `model-serving-config-defaults` ConfigMap is guaranteed to exist in the namespace of modelmesh controller. So, model-controller should read it to get the default service account name. This is the SA name to use if there is no user-configuration in place.
* A `model-serving-config` **can** exist in the namespace of modelmesh controller. If it exists, model-controller should check if there is an override for the service account name. If there is, then this should be the SA to use.

Current code seems to be covering the second bullet, but not the first one.

main.go
vaibhavjainwiz: Same handling was already implemented in `ModelMesh-Controller`. In `ModelMesh-Controller` they are using `NAMESPACE` env to fetch controller namespace. I keep the same in `odh-model-controller` to maintain consistency.
https://github.com/opendatahub-io/modelmesh-serving/blob/dd0229fe52d6c90703eabeecfdee287dd3ac55fd/main.go#L69

controllers/inferecenceservice_serviceaccount.go
vaibhavjainwiz: I have same doubt in this area as well. According to kserve/modelmesh-controller codebase, `ServiceAccount` should be maintained by the user. At the last user just provide the name reference of `ServiceAccount` which he want to link with ServingRuntime deployment but in opendatahub/odh-model-controller, we are creating `ServiceAccount` for user if its not already exists.

So if we follow kserve pattern then its a user responsiblity to delete old SA.

main.go
israel-hdez: I think it depends on which controller you mean. If this is for "self", the naming in OK. But after reading the code, most likely, this should refer to the `modelmesh-controller` namespace rather than $self. If this is the case, the naming should reflect that.

controllers/inferenceservice_controller.go
israel-hdez: Then, please, fix the Log message.

controllers/inferecenceservice_serviceaccount.go
israel-hdez: > According to kserve/modelmesh-controller codebase, `ServiceAccount` should be maintained by the user.

I agree with you. As an independent component, it is user responsibility.

However, my understanding is that `odh-model-controller` is adding missing things so that ODH can be used as a platform. Because of this, I'd lean towards deleting automatically. It is worth to check with other people to confirm.

vaibhavjainwiz:
> There seems to be 2 parts on the Jira:
> 
> * A `model-serving-config-defaults` ConfigMap is guaranteed to exist in the namespace of modelmesh controller. So, model-controller should read it to get the default service account name. This is the SA name to use if there is no user-configuration in place.
> * A `model-serving-config` **can** exist in the namespace of modelmesh controller. If it exists, model-controller should check if there is an override for the service account name. If there is, then this should be the SA to use.
> 
> Current code seems to be covering the second bullet, but not the first one.

The first bullet is already incoporate at below line.
https://github.com/opendatahub-io/modelmesh-serving/blob/58d9660a7a54813a66e228f84ad895e5b0d2e54b/pkg/config/config.go#L369

ConfigMap `model-serving-config-defaults` is mounted over `/config/default` location. Which refered in above code.

israel-hdez:
> The first bullet is already incoporate at below line.
> https://github.com/opendatahub-io/modelmesh-serving/blob/58d9660a7a54813a66e228f84ad895e5b0d2e54b/pkg/config/config.go#L369

But that is `modelmesh-serving`. This controller `odh-model-controller` also needs to read the `model-serving-config-defaults` ConfigMap.

vaibhavjainwiz:
> > The first bullet is already incoporate at below line.
> > https://github.com/opendatahub-io/modelmesh-serving/blob/58d9660a7a54813a66e228f84ad895e5b0d2e54b/pkg/config/config.go#L369
> 
> But that is `modelmesh-serving`. This controller `odh-model-controller` also needs to read the `model-serving-config-defaults` ConfigMap.

With my change we are using same library in odh-model-controller which we used in ModelMesh-controller.

controllers/inferenceservice_controller.go
vaibhavjainwiz: done

spolti:
This PR is one year old, I am going to close it, if it is something that still necessary, please reopen and fix the confclits.

Issue #59, opened 2023-07-24
OpenShift routes for KServe InferenceServices


israel-hdez:
In KServe, the URL where an InferenceService can be reached is reported in the same InferenceService. For example, this is a sample output when querying InferenceServices:

```
$ oc get isvc -A
NAMESPACE            NAME                      URL                                                                            READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION                       AGE
kserve-ci-e2e-test   isvc-sklearn-v2-mixed     http://isvc-sklearn-v2-mixed-kserve-ci-e2e-test.apps.openshift-cluster.com     True           100                              isvc-sklearn-v2-mixed-predictor-00001     6d6h
kserve-ci-e2e-test   isvc-tensorflow-runtime   http://isvc-tensorflow-runtime-kserve-ci-e2e-test.apps.openshift-cluster.com   True           100                              isvc-tensorflow-runtime-predictor-00001   6d6h
```

The issue is that the specified URL is not reachable in OpenShift.

**Describe the solution you'd like**

KServe assumes a wildcard DNS configuration, as mentioned here: 
* https://knative.dev/docs/install/yaml-install/serving/install-serving-with-yaml/#configure-dns (See _Real DNS_ tab)

Typically, in OpenShift, a similar DNS configuration is already set. However, while KServe assumes that the DNS configuration points to Istio Ingress Gateway, in OpenShift the DNS configuration points to OpenShift's Ingress.  This is the reason of seeing the generic "Application is not available" page when trying to access the URLs specified on the InferenceServices.

Since the URLs are hitting OpenShift's Ingress, it turns out that it is a matter of creating the right OpenShift Route, in order to properly enable the InferenceService endpoints. So, **the goal in this ticket is to create these missing Route resources**.

For example, the following route will make reachable the InferenceService `isvc-sklearn-v2-mixed` that is present in the example above:

```yaml
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  name: isvc-sklearn-v2-mixed-kserve-ci-e2e-test
  namespace: istio-system
spec:
  host: isvc-sklearn-v2-mixed-kserve-ci-e2e-test.apps.openshift-cluster.com
  port:
    targetPort: https
  tls:
    insecureEdgeTerminationPolicy: Redirect
    termination: passthrough
  to:
    kind: Service
    name: istio-ingressgateway
    weight: 100
  wildcardPolicy: None
```

NOTE that:
[...]

PR #84, opened 2023-09-18
Expose generic OpenShift routes for KServe InferenceServices

vaibhavjainwiz:
## Description
With https://github.com/opendatahub-io/odh-model-controller/issues/59 we had implemented a hot fix by exposing predictor URL directly to user but to provide proper fix we need to implement a generic openshift route for kserve inference service.

## How Has This Been Tested?
- Create a new cluster
- Setup the cluster with below Installation step
https://github.com/opendatahub-io/caikit-tgis-serving/blob/main/demo/kserve/install-manual.md
- Before executing Step 11, specify custom manifests
```
apiVersion: dscinitialization.opendatahub.io/v1
kind: DSCInitialization
metadata:
  name: example
spec:
  devFlags:
    manifestsUri: https://github.com/vaibhavjainwiz/odh-manifests/tarball/generic_route
```
- Apply Step 11.  
- Deploy/validate a Kserve model with below Step.
https://github.com/opendatahub-io/caikit-tgis-serving/blob/main/demo/kserve/deploy-remove.md

If everything work as mentioned in above step then we are good to go :)

## Expected behavior
A new route should be created in `istio-system` namespace with name same as InferenceService. If new route is created then we are good to go :)

## Merge criteria:


Jooho:
I am testing this PR but I have a couple of questions.

1. what images should be used for this test?
    when I followed your test steps, it deployed **quay.io/opendatahub/odh-model-controller:stable**
    I don't think it is your test image for this PR.
2. What is the expected result from these test steps?
    Where should I check? Route?

controllers/processors/deltaProcessor.go
```diff
@@ -0,0 +1,67 @@
+package processors
```
Jooho, line 1: I like this code and it is obvious to me but if you add some description on this file, it would be much easier to read the code.

controllers/kserve_inferenceservice_controller.go
```diff
@@ -468,6 +470,25 @@ func (r *OpenshiftInferenceServiceReconciler) createOrUpdateMetricsServiceMonito
 	return nil
 }
 
+func (r *OpenshiftInferenceServiceReconciler) DeleteKserveInferenceServiceRoute(ctx context.Context, inferenceService *kservev1beta1.InferenceService) error {
+	if r.isDeploymentModeForIsvcModelMesh(inferenceService) {
+		return nil
+	}
```
israel-hdez, line 476: Perhaps, this check shouldn't be done. It may be that the ISVC was switched from KServe mode to ModelMesh mode. 

In general, if there is an owned route which should not exist, the controller should delete it regardless of the deployment mode.

controllers/reconcilers/kisvc_route_reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/pointer"
+	"knative.dev/pkg/network"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
+)
+
+const (
+	IstioNamespace      = "istio-system"
+	IstioIngressService = "istio-ingressgateway"
```
israel-hdez, line 25: Please. create another ticket to properly parametrize this.

controllers/reconcilers/kisvc_route_reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+				},
+				TLS: &v1.TLSConfig{
+					Termination:                   v1.TLSTerminationPassthrough,
+					InsecureEdgeTerminationPolicy: v1.InsecureEdgeTerminationPolicyRedirect,
+				},
+				WildcardPolicy: v1.WildcardPolicyNone,
+			},
+			Status: v1.RouteStatus{
+				Ingress: []v1.RouteIngress{},
+			},
```
israel-hdez, line 126: Do you need to set the Status, or can you remove this fragment?

controllers/reconcilers/kisvc_route_reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+	if delta.IsRemoved() {
+		r.log.Info("Will", "delete", existingRoute.GetName())
+		if err = r.client.Delete(r.ctx, existingRoute); err != nil {
+			return
+		}
+	}
+	return nil
+}
+
+func getServiceHost(isvc *kservev1beta1.InferenceService) string {
```
israel-hdez, line 172: This seems to be a copy of KServe's logic to set the top-level status.URL.
Can we use the URL that KServe already would set for us, rather than replicating code here?

controllers/reconcilers/kisvc_route_reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+func (r *kserveInferenceServiceRouteReconciler) processDelta(desiredRoute *v1.Route, existingRoute *v1.Route) (err error) {
+	comparator := r.routeHandler.GetComparator()
+	delta := r.deltaProcessor.ComputeDelta(comparator, desiredRoute, existingRoute)
+
+	if !delta.HasChanges() {
+		r.log.Info("No delta found")
+		return nil
+	}
+
+	if delta.IsAdded() {
```
israel-hdez, line 146: Please, properly set owner references.

controllers/kserve_inferenceservice_controller.go
```diff
@@ -468,6 +470,25 @@ func (r *OpenshiftInferenceServiceReconciler) createOrUpdateMetricsServiceMonito
 	return nil
 }
 
+func (r *OpenshiftInferenceServiceReconciler) DeleteKserveInferenceServiceRoute(ctx context.Context, inferenceService *kservev1beta1.InferenceService) error {
```
israel-hdez, line 473: shouldn't this be part of kisvc_route_reconciler?

Delete logic in two places is error prone.

israel-hdez (changes requested):
Please, add unit tests.

controllers/kserve_inferenceservice_controller.go
```diff
@@ -644,5 +665,13 @@ func (r *OpenshiftInferenceServiceReconciler) ReconcileKserveInference(ctx conte
 	if err != nil {
 		return err
 	}
+
+	log.Info("Reconciling Generic Route for Kserve InferenceService")
+	kisvcRouteReconciler := reconcilers.NewKserveInferenceServiceRouteReconciler(r.Client, r.Scheme, ctx, log, inferenceService)
+	err = kisvcRouteReconciler.Reconcile()
```
israel-hdez, line 671: I haven't tested, but there seems to be a potential race condition here.

If `kserve-controller` reconciliation is delayed (i.e., heavy load), it is possible that `odh-model-controller` tries to reconcilie before `kserve-controller`. URLs will still not be set and `odh-model-controller` should re-try later.

I cannot spot any part where a retry is scheduled. So, I think will work only if order of events is _ideal_.

Please, cover the case where KServe reconciliation is delayed (for any reason). Or, please, correct me if I'm wrong and I was unable to spot the schedule of the retry.

controllers/kserve_inferenceservice_controller.go
vaibhavjainwiz: This route doesn't have owner reference because its created in `istio-system` namespace. Owner references could only be assigned within same namespaces thats why I have to explicitly write its deletion on deletion of isvc.

controllers/reconcilers/kisvc_route_reconciler.go
vaibhavjainwiz: In this case route could not have owner reference because its created in `istio-system` namespace. Owner references could only be assigned within same namespaces.

vaibhavjainwiz:
>1. what images should be used for this test?
>    when I followed your test steps, it deployed **quay.io/opendatahub/odh-model-controller:stable**
To test this PR you need to use odh-model-controller image created by openshift-ci for this corresponding PR.
`quay.io/opendatahub/odh-model-controller:pr-84`
I am not sure how we could provide overriden value of odh-model-controller image through v2 operator. @Jooho Could you please with this configuration.


> 2. What is the expected result from these test steps?
>    Where should I check? Route?
With this PR, new more generic route would be created in `istio-system` namespace which should be use to query model.

controllers/reconcilers/kisvc_route_reconciler.go
vaibhavjainwiz: I had created below ticket to parametrize these env var
https://github.com/opendatahub-io/odh-model-controller/issues/85

controllers/kserve_inferenceservice_controller.go
israel-hdez: Think of "owned", generally speaking, as a route that was created by the controller (not about the annotations).

I agree that explicit deletion is needed.

The question is around the conditional: `if r.isDeploymentModeForIsvcModelMesh(inferenceService)`.

I haven't tried the changes. But my perception is that given an `InferenceService` **without** the `serving.kserve.io/deploymentMode` annotation, the controller would correctly create a route. If later, the `serving.kserve.io/deploymentMode: ModelMesh` is added to this same InferenceService, this conditional will bypass deletion and the OpenShift Route will still exist as a remainder (which may be unwanted).

vaibhavjainwiz: @israel-hdez Do we support such type of transition from Kserve to ModelMesh or vice versa?

controllers/reconcilers/kisvc_route_reconciler.go
vaibhavjainwiz: Actually in this function we are picking the URL that kserve already set in the status of InferenceService and then drive the host name from it.

controllers/reconcilers/kisvc_route_reconciler.go
vaibhavjainwiz: Thanks @israel-hdez for pointed it out. We need to explicitly set the status. It would be inintialized by its own. I had removed this part now.

controllers/kserve_inferenceservice_controller.go
vaibhavjainwiz: done... I had moved Route deletion logic to RouteHandler.

controllers/kserve_inferenceservice_controller.go
vaibhavjainwiz: @israel-hdez nice catch.
Currently we don't have any provision in our code to send back schedule interval from function call. If I go to implement it right now then it would cause lot of changes and maybe hard to define boundaries of this partical task.

For now, I just retrun the `error` if ServiceURL is not available which would keep retrigggering the reconciler untill ServiceURL is available. 
I would raise a new ticket to support schduler functionality in our code.

controllers/processors/deltaProcessor.go
vaibhavjainwiz: done.

vaibhavjainwiz:
On my local, I follow following step for testing :

- Install the Kserve controller using odh operator
- Remove ODH operator
- Set deployment replicas of odh-model-controller to 0
- Run odh-model-controller from local in debug mode.


To support testing for reviewer, I had created a image of `odh-model-controller` from local and push it in quay.io. 
I had also updated the image reference in odh-manifest repo so if deploy their cluster with `https://github.com/vaibhavjainwiz/odh-manifests/tarball/generic_route` in DSCInitialization then everything should work.

vaibhavjainwiz:
@Jooho 
I had added **Expected behavior** section in Issue Description.

controllers/kserve_inferenceservice_controller.go
israel-hdez: > Do we support such type of transition from Kserve to ModelMesh or vice versa?
It depends on if KServe or ModelMesh block the edit via some Validating WebHook.

I can say that ModelMesh does not block the edit. But I'm unsure about KServe.

If the edit is allowed, it becomes _desired state_ and `odh-model-controller` should support it.

> I believe there would be multiple place that need this type of handling.
Probably yes.

> Could I create seperate ticket for it?
You can, although I'd think it should take priority. It is easy to forget the annotation for ModelMesh. If `odh-model-controller` is released as is, it would be seen as a bug.

vaibhavjainwiz: I had created below ticket for it with Critical Priority.
https://github.com/opendatahub-io/odh-model-controller/issues/88

controllers/reconcilers/kisvc_route_reconciler.go
israel-hdez: I was meaning that this code chooses between `status.components.predictor.url` or `status.components.transformer.url`. IMHO, it is better to simply look at `status.url`, because KServe has already set the needed one there, and you shouldn't need all this code.

controllers/kserve_inferenceservice_controller.go
vaibhavjainwiz: I had created below ticket to add scheduler interval support in odh-model-controller.
https://github.com/opendatahub-io/odh-model-controller/issues/89

controllers/reconcilers/kisvc_route_reconciler.go
vaibhavjainwiz: done

vaibhavjainwiz:
@israel-hdez I had the unit testcases. Could you please review.

controllers/kserve_inferenceservice_controller.go
```diff
@@ -468,6 +470,14 @@ func (r *OpenshiftInferenceServiceReconciler) createOrUpdateMetricsServiceMonito
 	return nil
 }
 
+func (r *OpenshiftInferenceServiceReconciler) OnDeletionOfKserveInferenceService(ctx context.Context, inferenceService *kservev1beta1.InferenceService) error {
+	log := r.Log.WithValues("InferenceService", inferenceService.Name, "namespace", inferenceService.Namespace)
+
+	log.Info("Deleting Kserve inference service generic route")
```
israel-hdez, line 476: nitpick: Same suggestion in this file about raising log-level for messages that look more like Debug messages.

israel-hdez (commented):
@vaibhavjainwiz I'm happy with the code.

I still want to test before approving. I didn't have time today.

If you want an approval sooner, maybe work with @rpancham and test it together?
Otherwise, I can try on my Tuesday.

controllers/inferenceservice_controller.go
```diff
@@ -152,3 +151,15 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
 
 	return nil
 }
+
+// general clean-up, mostly resources in different namespaces from kservev1beta1.InferenceService
+func (r *OpenshiftInferenceServiceReconciler) onDeletion(ctx context.Context, inferenceService *kservev1beta1.InferenceService) error {
+	log := r.Log.WithValues("InferenceService", inferenceService.Name, "namespace", inferenceService.Namespace)
+	log.Info("Running cleanup logic")
```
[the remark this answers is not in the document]
vaibhavjainwiz: done

controllers/kserve_inferenceservice_controller.go
vaibhavjainwiz: done

israel-hdez (commented):
I tested. So far, looks good, with some issues:
- It seems RBAC is missing in this PR. After installing and creating an ISVC I got `spec.host: Forbidden: you do not have permission to set the host field of the route`, which I fixed by modifying the Role to add something like this: https://github.com/openshift/origin/issues/14904#issuecomment-341757280.
    - I remember this RBAC would need to go in some kubebuilder comments, to let it generate them for you. I don't know how to do this.
- Similar permission denied error when trying to delete the ISVC and, thus, trying to delete the route.
- Name conflict. Having two ISVCs with same name on different namespaces, will lead to a conflict in the route.
- Schema mismatch: The ISVC can state that protocol is plain HTTP, while the route will be created for HTTPS.
  - Perhaps for our demo/sample setup, this is error config on KServe side. Yet, I think we want to honor the schema in the ISVC, just in case some user wants to use plain HTTP.

vaibhavjainwiz:
> I tested. So far, looks good, with some issues:
> 
> * It seems RBAC is missing in this PR. After installing and creating an ISVC I got `spec.host: Forbidden: you do not have permission to set the host field of the route`, which I fixed by modifying the Role to add something like this: [Route's new admission routes/custom-host doesn't work for legacy API openshift/origin#14904 (comment)](https://github.com/openshift/origin/issues/14904#issuecomment-341757280).
>
>   * I remember this RBAC would need to go in some kubebuilder comments, to let it generate them for you. I don't know how to do this.
> * Similar permission denied error when trying to delete the ISVC and, thus, trying to delete the route.

Strange these RBAC are already present in code. You should not see this error. I will debug it more.
https://github.com/opendatahub-io/odh-model-controller/blob/bf310d9baf0160cd020aaf0ec8ec737ba105d8e7/main.go#L68
https://github.com/opendatahub-io/odh-model-controller/blob/8ccb4ccc560d82e78950c646104812a2e599d6bd/config/rbac/role.yaml#L135
  
> * Name conflict. Having two ISVCs with same name on different namespaces, will lead to a conflict in the route.

I updated route name to include isvc.Name + isvc.Namespace so that It would not cause.

> * Schema mismatch: The ISVC can state that protocol is plain HTTP, while the route will be created for HTTPS.
>   
>   * Perhaps for our demo/sample setup, this is error config on KServe side (i.e. is config error on my testing). Yet, I think we want to honor the schema in the ISVC, just in case some user wants to use plain HTTP.
Are you talking about below config?
https://github.com/opendatahub-io/kserve/blob/1529b712ae536286fe939e7b6ad426305bf168f1/config/configmap/inferenceservice.yaml#L452

israel-hdez:
> Strange these RBAC are already present in code. You should not see this error. I will debug it more.
> 
> https://github.com/opendatahub-io/odh-model-controller/blob/bf310d9baf0160cd020aaf0ec8ec737ba105d8e7/main.go#L68
> 
> https://github.com/opendatahub-io/odh-model-controller/blob/8ccb4ccc560d82e78950c646104812a2e599d6bd/config/rbac/role.yaml#L135
> 

Yes, some RBAC is there. Although, it is missing `delete` privileges.
It is also missing the privilege to set custom hosts, which is a different resource.

> > * Schema mismatch: The ISVC can state that protocol is plain HTTP, while the route will be created for HTTPS.
> >   
> >   * Perhaps for our demo/sample setup, this is error config on KServe side (i.e. is config error on my testing). Yet, I think we want to honor the schema in the ISVC, just in case some user wants to use plain HTTP.
> 
> Are you talking about below config? https://github.com/opendatahub-io/kserve/blob/1529b712ae536286fe939e7b6ad426305bf168f1/config/configmap/inferenceservice.yaml#L452


Yes. That configuration is reflected in the ISVC. So, you don't need to look at the ConfigMap. Just check the schema set in the ISVC.

vaibhavjainwiz:
@israel-hdez I had incoporated all review comments. Could you please review again.
I had made the change in odh-model-controller to honor user specified url scheme but some change are also required in kserve as well because url in InferenceService status is set by KServe controller itself. I will create a seperate ticket in KServe to fix it.

israel-hdez (approved):

Issue #125, opened 2023-11-29
bug: modelmesh container have error logs when kserve runtime is running.


Jooho:
When kserve and modelmeh are running in the same namespace, modelmesh container show these errors:
~~~
{"instant":{"epochSecond":1701291690,"nanoOfSecond":966215444},"thread":"ll-elg-thread-2","level":"INFO","loggerName":"com.ibm.watson.modelmesh.ModelMesh","message":"Returning READY to readiness probe (did not find any other pods in terminating state)","endOfBatch":false,"loggerFqcn":"org.apache.logging.log4j.spi.AbstractLogger","contextMap":{},"threadId":35,"threadPriority":5}
Nov 29, 2023 9:01:37 PM io.grpc.netty.NettyServerTransport notifyTerminated
INFO: Transport failed
io.netty.handler.codec.http2.Http2Exception: Unexpected HTTP/1.x request: GET /stats/prometheus
at io.netty.handler.codec.http2.Http2Exception.connectionError(Http2Exception.java:109)
at io.netty.handler.codec.http2.Http2ConnectionHandler$PrefaceDecoder.readClientPrefaceString(Http2ConnectionHandler.java:317)
at io.netty.handler.codec.http2.Http2ConnectionHandler$PrefaceDecoder.decode(Http2ConnectionHandler.java:247)
at io.netty.handler.codec.http2.Http2ConnectionHandler.decode(Http2ConnectionHandler.java:453)
at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:529)
at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:468)
~~~

There are 3 networkpolicy in the namespace:
- allow-from-openshift-monitoring-ns
- istio-expose-route-minimal
- istio-mesh-minimal

If allow-from-openshift-monitoring-ns network policy is deleted, the error message is not showing up anymore. So I think this networkpolicy is the culprit of this issue. However, it is not 100% so it needs more debugging.

Reference:
https://github.com/orgs/opendatahub-io/projects/42?pane=issue&itemId=40292089

skonto:
By setting istio-prometheus-ignore="true"  you can avoid scraping on port 15020 happening on the modelmesh pod.
See:
```
Name:         istio-proxies-monitor
Namespace:    kserve-demo
... 
Spec:
  Namespace Selector:
[...]

PR #126, opened 2023-12-06
Remove PodMonitor from Namespace level and implement PodMonitor for e…

vaibhavjainwiz:
…ach KServe isvc.

## Description

Currently PodMonitor for KServe isvc is monitoring all pod in the namespace which causes issue in ModelMesh monitor. With this PR I had change the behaviour of PodMonitor to just monitor pods of KServe isvc. It should fix https://github.com/opendatahub-io/odh-model-controller/issues/125

## Merge criteria:

### DIFF
--- a/controllers/reconcilers/kserve_podmonitor_reconciler.go
+++ b/controllers/reconcilers/kserve_podmonitor_reconciler.go
@@ -19,28 +19,27 @@ import (
 	"context"
 	"github.com/go-logr/logr"
 	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
+	"github.com/kserve/kserve/pkg/constants"
 	"github.com/opendatahub-io/odh-model-controller/controllers/comparators"
 	"github.com/opendatahub-io/odh-model-controller/controllers/processors"
 	"github.com/opendatahub-io/odh-model-controller/controllers/resources"
 	v1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const (
-	istioPodMonitorName = "istio-proxies-monitor"
-)
-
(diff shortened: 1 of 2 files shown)


vaibhavjainwiz:
Next step is to verify the metrics. With this change both ModelMesh and KServe metrics should be available in Openshift Monitoring stack.

controllers/reconcilers/kserve_istio_podmonitor_reconciler.go
```diff
@@ -71,29 +70,7 @@ func (r *KserveIstioPodMonitorReconciler) Reconcile(ctx context.Context, log log
 }
 
 func (r *KserveIstioPodMonitorReconciler) createDesiredResource(isvc *kservev1beta1.InferenceService) (*v1.PodMonitor, error) {
-	desiredPodMonitor := &v1.PodMonitor{
```
spolti, line 74: what is the point of having this function just to return nil?

controllers/reconcilers/kserve_podmonitor_reconciler.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	return desiredPodMonitor, nil
+}
+
+func (r *KserveMetricsPodMonitorReconciler) getExistingResource(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) (*v1.PodMonitor, error) {
+	return r.podMonitorHandler.FetchPodMonitor(ctx, log, types.NamespacedName{Name: getMetricsPodMonitorName(isvc), Namespace: isvc.Namespace})
+}
+
+func (r *KserveMetricsPodMonitorReconciler) processDelta(ctx context.Context, log logr.Logger, desiredPod *v1.PodMonitor, existingPod *v1.PodMonitor) (err error) {
+	comparator := comparators.GetPodMonitorComparator()
```
spolti, line 102: Don't we have these functions already on the RHSyseng operator-utils?

maybe we could use it from there and avoid each reconciler have its own deltaProcessor.

spolti (changes requested):
Can you please add a unit test for it?

Just to make sure that the needed resources will be created/update the objects as desired when the reconciler is called.

vaibhavjainwiz:
Thank you @spolti for reviewing it but I need to close this PR because this solution didn't work. I will raise a new PR to fix this issue on bases of below thread.
https://github.com/opendatahub-io/odh-model-controller/issues/125#issuecomment-1850153594

spolti:
ok, thanks :)

PR #130, opened 2023-12-08
feat: update project level auth config with kserve hosts

aslakknutsen:
## Description

Add Authorino AuthConfig for KServe InferenceService hosts in ODHProject Namespaces.

The `InferenceService` can be annotated with `enable-auth` to trigger the creation of an `AuthConfig` that secures the `InferenceService` access. The secure access require a kubernetes token and access to `get` the `InferenceService` resources in the target `Namespace`.

If  `enable-auth` is not present, an AuthConfig for anonymous access will be created instead. 

## How Has This Been Tested?

enable-auth:

grpc|curl -H "Authorization: Bearer $(oc whoami -t)" ...

no auth:

grpc|curl ...

## Merge criteria:

# Dependencies

- opendatahub-io/opendatahub-operator#784
- #158


go.mod
```diff
@@ -5,9 +5,12 @@ go 1.19
 require (
 	github.com/go-logr/logr v1.2.4
 	github.com/kserve/kserve v0.11.0
+	github.com/kuadrant/authorino v0.15.0
+	github.com/manifestival/manifestival v0.7.2
 	github.com/onsi/ginkgo v1.16.4
 	github.com/onsi/gomega v1.27.6
 	github.com/openshift/api v3.9.0+incompatible
+	github.com/pkg/errors v0.9.1
```
spolti, line 13: Hi

any reason to make it direct dependency?

controllers/reconcilers/kserve_authconfig_reconciler.go
```diff
@@ -0,0 +1,108 @@
...
+		detector:       resources.NewKServeAuthTypeDetector(client),
+		store:          resources.NewClientAuthConfigStore(client),
+		templateLoader: resources.NewConfigMapTempalteLoader(client, resources.NewStaticTemplateLoader()),
+	}
+}
+
+func (r *KserveAuthConfigReconciler) Reconcile(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+
+	if isvc.Status.URL == nil {
+		log.V(1).Info("Inference Service not ready yet, waiting for URL")
```
spolti, line 54: imho, it should be a warning.

controllers/resources/authconfig.go
```diff
@@ -0,0 +1,172 @@
...
+
+func NewConfigMapTempalteLoader(client client.Client, fallback AuthConfigTemplateLoader) AuthConfigTemplateLoader {
+	return &configMapTemplateLoader{
+		client:   client,
+		fallback: fallback,
+	}
+}
+
+func (c *configMapTemplateLoader) Load(authType AuthType, key types.NamespacedName) (authorinov1beta2.AuthConfig, error) {
+	// TOOD: check "authconfig-template" CM in key.Namespace to see if there is a "spec" to use, construct a AuthConfig object
```
spolti, line 95: maybe open a issue to not lost track of it.

controllers/resources/authconfig.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func NewClientAuthConfigStore(client client.Client) AuthConfigStore {
+	return &clientAuthConfigStore{
+		client: client,
+	}
+}
+
+func (c *clientAuthConfigStore) Get(ctx context.Context, key types.NamespacedName) (authorinov1beta2.AuthConfig, error) {
+	authConfig := authorinov1beta2.AuthConfig{
```
spolti, line 112: this same peace of code is used on another go file, maybe create a common authconfig object that can be reused.

controllers/resources/authconfig_test.go
```diff
@@ -0,0 +1,39 @@
+package resources_test
+
+import (
```
spolti, line 3: maybe add a reconcile test?

controllers/utils/converter.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/manifestival/manifestival"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+)
+
+func ConvertToStructuredResource(yamlContent []byte, out runtime.Object, opts ...manifestival.Option) error {
+
+	s := runtime.NewScheme()
+	RegisterSchemes(s)
```
spolti, line 12: not sure if I followed, why is this needed here?

controllers/reconcilers/kserve_authconfig_reconciler.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	return nil
+}
+
+func (r *KserveAuthConfigReconciler) Remove(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+	typeName := types.NamespacedName{
+		Name:      isvc.GetNamespace(),
+		Namespace: isvc.GetNamespace(),
+	}
```
bartoszmajsak, line 106: I think you meant

```suggestion
	typeName := types.NamespacedName{
		Name:      isvc.GetName(),
		Namespace: isvc.GetNamespace(),
	}
```

controllers/reconcilers/kserve_authconfig_reconciler.go
aslakknutsen: Could make it a warning. It's part of the normal flow to reconcile a few times before the Status.URL show up

controllers/resources/authconfig.go
aslakknutsen: https://github.com/opendatahub-io/odh-model-controller/pull/130/commits/9086d691f96b408942fc80727bc7327233bfa483

controllers/resources/authconfig.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	return Anonymous, nil
+}
+
+// inclueLocalHostsIfPresent will add other internal host addresses if an internal address is present
+// *.x.svc.cluster.local
+// *.x.svc
+// *.x
+/*
+func inclueLocalHostsIfPresent(hosts []string) []string {
```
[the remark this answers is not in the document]
aslakknutsen: yeah :) https://github.com/opendatahub-io/odh-model-controller/pull/130/commits/4d7b3a8a88d9d5505b61a269bc5a5f0872ec091f

controllers/reconcilers/kserve_authconfig_reconciler.go
```diff
@@ -0,0 +1,160 @@
...
+
+func (r *KserveAuthConfigReconciler) createDesiredResource(ctx context.Context, isvc *kservev1beta1.InferenceService) (*authorinov1beta2.AuthConfig, error) {
+	typeName := types.NamespacedName{
+		Name:      isvc.GetName(),
+		Namespace: isvc.GetNamespace(),
+	}
+
+	authType, err := r.detector.Detect(ctx, isvc)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not detect AuthType for InferenceService %s", typeName)
```
israel-hdez, line 93: @aslakknutsen @bartoszmajsak The unmaintained libray... uhm :confused:

israel-hdez:
So, I was trying to make this work.
Looks like everything seems to be correctly "wired up" in my cluster. Despite so, the request still goes through even when annotating the ISVC with `enable-auth=true`.

@aslakknutsen @bartoszmajsak I may meet with you soon to see if we can debug this.

controllers/resources/template/authconfig_userdefined.yaml
```diff
@@ -0,0 +1,26 @@
...
+        authorizationHeader: {}
+      kubernetesTokenReview:
+        audiences:
+        - https://kubernetes.default.svc
+  response:
+    unauthenticated:
+      code: 302
+      headers:
+        Location:
+          selector: http://some-service?redirect_to={context.request.http.path}
```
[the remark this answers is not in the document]
aslakknutsen: Yeah, we can just remove them. Then the response will be 40x

controllers/resources/authconfig.go
```diff
@@ -0,0 +1,210 @@
...
+	base := strings.ReplaceAll(statusHost[0:strings.Index(statusHost, isvc.Namespace)-1], "-predictor", "")
+	basePre := base + "-predictor"
+
+	// caikit-example-isvc-kserve-demo.apps-crc.testing
+	hosts = append(hosts, base+statusHost[strings.Index(statusHost, isvc.Namespace)-1:])
+	// // caikit-example-isvc.kserve-demo.svc.cluster.local
+	hosts = append(hosts, base+"."+isvc.Namespace+".svc.cluster.local")
+
+	// caikit-example-isvc-predictor-kserve-demo.apps-crc.testing
+	hosts = append(hosts, basePre+statusHost[strings.Index(statusHost, isvc.Namespace)-1:])
```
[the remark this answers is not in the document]
aslakknutsen: There's only two of them, sadly. We could surly include those as well (in case we didn't infer them) as a safety. The reason for _all_ the hostnames is simply that that's what the VirtualServices setup. 

I guess we could read all the hosts from the VirtualServices instead. Then we "only" need to infer the VirtualService names.

controllers/resources/template/authconfig_userdefined.yaml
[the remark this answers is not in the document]
israel-hdez: Let's go this route for now.

controllers/resources/authconfig.go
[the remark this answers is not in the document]
israel-hdez: > There's only two of them, sadly.
So, yes, there are two. This is all what you would get in the current release.
On the upcoming release, we will have three because `status.url` will become `https://sklearn-v2-iris-kserve-test.apps-crc.testing` (without the `predictor`).

From these three, I'd think the remaining work is to infer the other shortened in-cluster hostnames.

> I guess we could read all the hosts from the VirtualServices instead.

We could... and the name of the `VirtualService` seems [to be the same name as the InferenceService](https://github.com/opendatahub-io/kserve/blob/release-v0.11.1/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go#L442).

The problem that I see around inferring the hostnames is that these does not necessarily come from a single VirtualService, and we may need to protect not only the predictor pod.

Try the Transformer example for a more complete list of hostnames that may need to be protected: https://kserve.github.io/website/latest/modelserving/v1beta1/transformer/feast/#create-the-inferenceservice.

I haven't looked at it but, AFAIK, the VS that KServe creates will switch from using `predictor` hostnames to using `transformer` hostnames. Yet, the ISVC will still contain all involved hostames on several `url` fields.

controllers/reconcilers/kserve_authconfig_reconciler.go
israel-hdez: In this repo it is going to be the first usage. Indeed, Filippe pointed that this library now would become a direct dependency [here](https://github.com/opendatahub-io/odh-model-controller/pull/130#discussion_r1420924906).

I think, I'd prefer to not introduce it as a dependency, unless we have a good reason to (directly) depend on that 3rd party library.

controllers/resources/template/authconfig_userdefined.yaml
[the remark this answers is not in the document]
aslakknutsen: fixed https://github.com/opendatahub-io/odh-model-controller/pull/130/commits/93f3c2ff7b534b399db082d10a7cf7b9647f149f

config/rbac/role.yaml
```diff
@@ -5,18 +5,6 @@ metadata:
   creationTimestamp: null
   name: odh-model-controller-role
 rules:
-- apiGroups:
```
israel-hdez, line 8: Some changes here look like a rebase issue, since were recently added.

controllers/resources/authconfig.go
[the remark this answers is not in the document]
aslakknutsen: https://github.com/opendatahub-io/odh-model-controller/pull/130/commits/7aefd5d604e7db684e462b6ef74305abaa9fc330

controllers/reconcilers/kserve_authconfig_reconciler.go
aslakknutsen: https://github.com/opendatahub-io/odh-model-controller/pull/130/commits/5c1fdf8a99e58492a17508406d6e79e7875a22bc

go.mod
aslakknutsen: https://github.com/opendatahub-io/odh-model-controller/pull/130/commits/5c1fdf8a99e58492a17508406d6e79e7875a22bc

config/rbac/role.yaml
israel-hdez: Looks like this was added manually, rather than using kubebuilder comments. Thus, `make manifests` would remove this.

I'll create a PR fixing this.

israel-hdez: https://github.com/opendatahub-io/odh-model-controller/pull/158

controllers/resources/authconfig_test.go
aslakknutsen: https://github.com/opendatahub-io/odh-model-controller/pull/130/commits/2f5b8e3537123a9ae9adb76dac39a1f2935caec4

controllers/utils/converter.go
israel-hdez: So, this is creating an empty Scheme which is needed by the decoder a few lines later.
I guess, the decoder won't work correctly if the Schemes/CRDs are not present in the Scheme.

controllers/reconcilers/kserve_authconfig_reconciler.go
israel-hdez: So, `log` does not have a warning func. Only `Info` and `Error`.
We have `V(1)` in the middle to raise verbosity, so I think this is good enough.

controllers/kserve_inferenceservice_controller_authconfig_test.go
```diff
@@ -0,0 +1,222 @@
...
+	inferenceService := createBasicISVC(namespace)
+	inferenceService.Annotations["enable-auth"] = "true"
+	Expect(cli.Create(ctx, inferenceService)).Should(Succeed())
+
+	// HMM... second routine to do updates?
+	url, _ := apis.ParseURL("http://iscv-" + namespace + "ns.apps.openshift.ai")
+	inferenceService.Status = kservev1beta1.InferenceServiceStatus{
+		URL: url,
+	}
+
```
israel-hdez, line 174: This chunk of code seems to be safe to remove, given `updateISVCStatus` is called on the relevant units.

controllers/kserve_inferenceservice_controller_authconfig_test.go
```diff
@@ -0,0 +1,222 @@
...
+func createISVCWithoutAuth(namespace string) *kservev1beta1.InferenceService {
+	inferenceService := createBasicISVC(namespace)
+	Expect(cli.Create(ctx, inferenceService)).Should(Succeed())
+
+	// HMM... second routine to do updates?
+	url, _ := apis.ParseURL("http://iscv-" + namespace + "ns.apps.openshift.ai")
+	inferenceService.Status = kservev1beta1.InferenceServiceStatus{
+		URL: url,
+	}
```
israel-hdez, line 186: Same here: seems to be safe to remove.

controllers/kserve_inferenceservice_controller_authconfig_test.go
```diff
@@ -0,0 +1,222 @@
...
+	url, _ := apis.ParseURL("http://iscv-" + isvc.Namespace + "ns.apps.openshift.ai")
+	isvc.Status = kservev1beta1.InferenceServiceStatus{
+		URL: url,
+	}
+	return cli.Status().Update(context.Background(), isvc)
+}
+
+func disableAuth(isvc *kservev1beta1.InferenceService) error {
+	delete(isvc.Annotations, "enable-auth")
+	delete(isvc.Annotations, "security.opendatahub.io/enable-auth")
```
israel-hdez, line 212: Maybe, these strings should be a `const`.

controllers/resources/template/authconfig_userdefined.yaml
```diff
@@ -0,0 +1,31 @@
...
+      credentials:
+        authorizationHeader: {}
+      kubernetesTokenReview:
+        audiences:
+{{- range .Audiences }}
+        - "{{ . }}"
+{{- end }}
+  authorization:
+    kubernetes-rbac:
+      kubernetesSubjectAccessReview:
```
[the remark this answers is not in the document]
israel-hdez: I confirm we must use the same SAR as in ModelMesh.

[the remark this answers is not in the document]
israel-hdez: I changed the SAR in commit efd73ff

controllers/reconcilers/kserve_authconfig_reconciler.go
```diff
@@ -0,0 +1,162 @@
...
+	comparator := comparators.GetAuthConfigComparator()
+	delta := r.deltaProcessor.ComputeDelta(comparator, desiredState, existingState)
+
+	if !delta.HasChanges() {
+		log.V(1).Info("No delta found")
+		return nil
+	}
+
+	if delta.IsAdded() {
+		log.V(1).Info("Delta found", "create", desiredState.GetName())
```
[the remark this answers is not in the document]
aslakknutsen: This is basically copy/paste from the other "Reconcilers" in the project. I agree it's not very useful, but at least it's consistent. :)

Personally I would have skipped this for this PR and rather raise a more general "logging" PR.

controllers/resources/template/authconfig_userdefined.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: authorino.kuadrant.io/v1beta2
+kind: AuthConfig
+metadata:
+  labels:
+    security.opendatahub.io/authorization-group: default
+spec:
+  hosts:
+  - "UPDATED.RUNTIME"
+  authentication:
+    kubernetes-user:
```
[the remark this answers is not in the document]
aslakknutsen: That's just a user defined name. We can make it Edgar if you want? :)

controllers/kserve_inferenceservice_controller_authconfig_test.go
aslakknutsen: Fixed https://github.com/opendatahub-io/odh-model-controller/pull/130/commits/a6c333a99c31f82e5f6a35d314bee22de3951685

Seeing them side by side I wonder if maybe we should have called it something like `security.opendatahub.io/authorization` instead...?

controllers/resources/template/authconfig_userdefined.yaml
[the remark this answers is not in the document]
israel-hdez: Ah, well... I thought it was meaning you want to authenticate a kubernetes user :-)

controllers/reconcilers/kserve_authconfig_reconciler.go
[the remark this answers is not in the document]
israel-hdez: OK, let's let it be.

controllers/kserve_inferenceservice_controller_authconfig_test.go
israel-hdez: It could be. Although, for enabling and disabling, I'm not sure if it is that useful to make distinction between authorization and authentication.

But you tell, I let you do the renaming if you think it is better.

aslakknutsen: I was more thinking the short vs "auth" vs "auth..." and "auth-x" vs "x-auth" order. (not authentication vs authorization)

security.opendatahub.io/authorization-group
security.opendatahub.io/enable-auth

security.opendatahub.io/authorization-group
security.opendatahub.io/authorization-enabled

israel-hdez: The `security.opendatahub.io/authorization-enabled` is the one that I see as potentially noisy for _later_. Right now, we are dealing only with authZ and the naming looks fine. But once we start looking at proper SSO, we may want to expand this with the authN part and add back to the `AuthConfig` the redirect that we removed previously. Then, this annotation would work to enable/disable both authN and authZ together, and its name will no longer fully reflect what it is doing.

aslakknutsen: is `security....io/enabled` sufficient? It hints at securing the Model. Would there be other types of 'security' that could make it confusing in the future at some point?

israel-hdez: > is security....io/enable sufficient?

Uhm... I'm OK with it, but...

> Would there be other types of 'security' 

I don't know....
The only other thing that has been a topic lately is around TLS certs. But IDK if we will use annotations or labels for that.

Issue #128, opened 2023-12-07
bug: modelmesh runtime is not accessible from exposed route


vaibhavjainwiz:
Once we resolved the https://github.com/opendatahub-io/odh-model-controller/issues/127 issue by creating a NetworkPolicy. Deployed ModelMesh InferenceService exposed the route to make prediction call.
But I am facing below issue while making prediction call.

request:
```
curl --location 'https://example-onnx-mnist-kserve-demo.apps.rosa.vajain.9yeo.p3.openshiftapps.com/v2/models/example-onnx-mnist/infer' \
--header 'Content-Type: application/json' \
--data '{
  "model_name": "example-onnx-mnist",
  "inputs": [
    {
      "name": "Input3",
      "shape": [1, 1, 28, 28],
      "datatype": "FP32",
      "data": [
        0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
        0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
        0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
        0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
        0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
        0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
        0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
        0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
        0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
[...]

Issue #127, opened 2023-12-07
bug: modelmesh controller have error logs when kserve and ModelMesh run in same namespace


vaibhavjainwiz:
When kserve and modelmeh are running in the same namespace, modelmesh controller show these errors:
```
{"level":"error","ts":"2023-12-07T11:33:47Z","msg":"Reconciler error","controller":"predictor","controllerGroup":"serving.kserve.io","controllerKind":"Predictor","Predictor":{"name":"caikit-tgis-example-isvc","namespace":"isvc_kserve-demo"},"namespace":"isvc_kserve-demo","name":"caikit-tgis-example-isvc","reconcileID":"868aa907-1733-408b-a8cd-482ac234f616","error":"failed to remove corresponding VModel for deleted Predictor kserve-demo/caikit-tgis-example-isvc: rpc error: code = Unavailable desc = last connection error: connection error: desc = \"transport: Error while dialing: dial tcp 10.128.0.84:8033: i/o timeout\"","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/root/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:329\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/root/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:274\ns...
[...]

PR #132, opened 2023-12-12
Allow ModelMesh and KServe to run in same namespace.

vaibhavjainwiz:
Solving following issue with this PR:
- https://github.com/opendatahub-io/odh-model-controller/issues/133
- https://github.com/opendatahub-io/odh-model-controller/issues/125
- https://github.com/opendatahub-io/odh-model-controller/issues/128
- https://github.com/opendatahub-io/odh-model-controller/issues/127

## How Has This Been Tested?
**Setup Env**
Please follow below instruction to setup your env:
https://gist.github.com/vaibhavjainwiz/61045727dfda727f7577a6f97ca46563

**Validate**

1. After above setup, verift both KServe and ModelMesh InferenceServices are running successfully.
2. There should not be any error in log of KServe controller/Model Controller and runtime pods.
3. Both isvc should be accessble using curl/grpc.
4. Now delete ModelMesh InferenceService, KServe isvc should be running without any error.
5. Now delete KServe isvc, all dependent resources should be deleted.

## Merge criteria:

### DIFF
--- a/controllers/inferenceservice_controller.go
+++ b/controllers/inferenceservice_controller.go
@@ -115,31 +115,12 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
 		Watches(&source.Kind{Type: &kservev1alpha1.ServingRuntime{}},
 			handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
 				r.log.Info("Reconcile event triggered by serving runtime: " + o.GetName())
-				inferenceServicesList := &kservev1beta1.InferenceServiceList{}
-				opts := []client.ListOption{client.InNamespace(o.GetNamespace())}
-
-				// Todo: Get only Inference Services that are deploying on the specific serving runtime
-				err := r.client.List(context.TODO(), inferenceServicesList, opts...)
-				if err != nil {
-					r.log.Info("Error getting list of inference services for namespace")
-					return []reconcile.Request{}
-				}
-
-				if len(inferenceServicesList.Items) == 0 {
-					r.log.Info("No InferenceServices found for Serving Runtime: " + o.GetName())
-					return []reconcile.Request{}
-				}
-
-				reconcileRequests := make([]reconcile.Request, 0, len(inferenceServicesList.Items))
-				for _, inferenceService := range inferenceServicesList.Items {
-					reconcileRequests = append(reconcileRequests, reconcile.Request{
-						NamespacedName: types.NamespacedName{
-							Name:      inferenceService.Name,
-							Namespace: inferenceService.Namespace,
-						},
-					})
-				}
-				return reconcileRequests
+				return r.getReconcileRequestsOnUpdateOfServingRuntime(o)
+			})).
+		Watches(&source.Kind{Type: &networkingv1.NetworkPolicy{}},
+			handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
+				r.log.Info("Reconcile event triggered by Network Policy: " + o.GetName())
+				return r.getReconcileRequestsOnUpdateOfServingRuntime(o)
 			}))
 	err := builder.Complete(r)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -55,6 +55,21 @@ var _ = Describe("The Openshift Kserve model controller", func() {
 
 		})
 
+		//Context("Validate Openshift Monitoring Stack", func() {
```
spolti, line 58: leftover or?

controllers/inferenceservice_controller.go
```diff
@@ -115,31 +115,12 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
...
-						NamespacedName: types.NamespacedName{
-							Name:      inferenceService.Name,
-							Namespace: inferenceService.Namespace,
-						},
-					})
-				}
-				return reconcileRequests
+				return r.getReconcileRequestsOnUpdateOfServingRuntime(o)
+			})).
+		Watches(&source.Kind{Type: &networkingv1.NetworkPolicy{}},
```
spolti, line 120: why do we need to watch the network policy here?
I guess it would be better to only notify the user if it is needed to be configured instead creating it ourselves.

controllers/reconcilers/mm_controller_networkpolicy_reconciler.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	// Process Delta
+	if err = r.processDelta(ctx, log, desiredResource, existingResource); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *MMControllerNetworkPolicyReconciler) createDesiredResource(isvc *kservev1beta1.InferenceService) (*v1.NetworkPolicy, error) {
```
spolti, line 70: this and the other reconcile have almost the same code base, can it be moved to a common function and be reused?

controllers/inferenceservice_controller.go
spolti: auto creating network policies can be dangerous sometimes, while the network policy reconcile seems only be allowing communication between namespaces, I am not sure what are the cons in this case.

vaibhavjainwiz: In our case, KServe always run in Istio monitored namespace where modelmesh doesn't support it so if we ever want to run modelmesh and kserve in same namespace then we need these network policies so that ModelMesh could work efficently.

controllers/reconcilers/mm_controller_networkpolicy_reconciler.go
vaibhavjainwiz: yes.. I also notice this pattern. I had created below Jira for it.
https://issues.redhat.com/browse/RHOAIENG-1228

controllers/kserve_inferenceservice_controller_test.go
vaibhavjainwiz: removed it.

controllers/inferenceservice_controller.go
spolti: I see, thanks.

controllers/reconcilers/mm_controller_networkpolicy_reconciler.go
spolti: cool :)
thanks.

controllers/inferenceservice_controller.go
```diff
@@ -115,31 +115,12 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
...
+			handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
+				r.log.Info("Reconcile event triggered by Network Policy: " + o.GetName())
+				return r.getReconcileRequestsOnUpdateOfServingRuntime(o)
```
davidesalerno, line 123: Is it right to call the `getReconcileRequestsOnUpdateOfServingRuntime` and not the `getReconcileRequestsOnUpdateOfNetworkPolicy` function?

controllers/inferenceservice_controller.go
```diff
@@ -149,6 +130,61 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
...
+			NamespacedName: types.NamespacedName{
+				Name:      inferenceService.Name,
+				Namespace: inferenceService.Namespace,
+			},
+		})
+	}
+	return reconcileRequests
+}
+
+func (r *OpenshiftInferenceServiceReconciler) getReconcileRequestsOnUpdateOfNetworkPolicy(o client.Object) []reconcile.Request {
```
davidesalerno, line 161: Do we really need this function? It seems to be equal, except 1 comment to the `getReconcileRequestsOnUpdateOfServingRuntime`

controllers/inferenceservice_controller.go
vaibhavjainwiz: yes, you are right. I make the wrong function call. Its not tracked in my testing earlier because both function do similiar task.
I had fixed the function call now.

controllers/inferenceservice_controller.go
vaibhavjainwiz: for now they are doing same task to return list of all InferenceServices in that namespace but we have task in our backlog to update the logic of `getReconcileRequestsOnUpdateOfServingRuntime` function. 
https://issues.redhat.com/browse/RHOAIENG-1265

controllers/reconcilers/mm_controller_networkpolicy_reconciler.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/opendatahub-io/odh-model-controller/controllers/processors"
+	"github.com/opendatahub-io/odh-model-controller/controllers/resources"
+	"k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type MMControllerNetworkPolicyReconciler struct {
```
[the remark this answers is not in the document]
vaibhavjainwiz: To fix this issue, I need to implement two NetworkPolicies :

- allow access to expose Route 
- allow calls from odh-model-controller

To make visible distinction between these two NetworkPolicies, I had create file with below names :

- mm_route_networkpolicy_reconciler.go
- mm_controller_networkpolicy_reconciler.go

I am open to change their name to be more meaningful.

vaibhavjainwiz:
Setting this PR to **do-not-merge** because NetworkPolicies created with this PR would not be required any more after https://issues.redhat.com/browse/RHOAIENG-1003 fix.

israel-hdez:
> Setting this PR to **do-not-merge** because NetworkPolicies created with this PR would not be required any more after https://issues.redhat.com/browse/RHOAIENG-1003 fix.

May this PR is still relevant? I think RHOAIENG-1003 was reverted.

spolti:
Do we still need this PR?

israel-hdez:
Closing as stale...

PR #124, opened 2023-11-27
Setup model registry serving reconciliation

lampajr:
Fixes https://github.com/opendatahub-io/model-registry/issues/174
Fixes https://github.com/opendatahub-io/model-registry/issues/175

## Description

Implement the Model Registry <--> Model Serving interaction described in issue https://github.com/opendatahub-io/model-registry/issues/104.

Setup a new Reconciler that is triggered whenever a `ServingRuntime` is create/updated/deleted and in a periodic fashion (i.e., every N seconds where N could be specified as startup flag, default `60 seconds`).

The summarized reconciliation workflow is the following:
- For each `<Namespace>/<ServingRuntime>` do:
  - Fetch all Inference services stored in the model registry and associated to the current NS (modeled as ServingEnvironment) and having as `Runtime` the current `ServingRuntime` name.
  - For each model registry `<InferenceService>` do:
    - Check if an InferenceService CR already exists for the current IS (use a custom label to keep track of these associations)
    - Check the model registry IS state
    - Then:
      - if STATE==DEPLOYED and not existing in cluster --> Create new ISVC
      - if STATE==UNDEPLOYED and not existing in cluster --> Do nothing, skip as there is no intention to deploy and there is no deployment in the cluster for that IS
      - if STATE==DEPLOYED and existing in cluster --> Check if there is any delta, if so apply updated CR
      - if STATE==UNDEPLOYED and existing in cluster --> This is an intention to undeploy the model, therefore delete the ISVC

ODH model controller keep track of those operation in the model registry itself, specifically as `ServeModel` recods (one per deployment).
These records can have different states, for now I used the following:
- RUNNING: deployed model (associated to the ISVC) is running in the namespace
- FAILED: Something went wrong creating/updating/deleting the ISVC - more details can be found in the CR itself
- UNKNOWN: Temporary and intermediate steps when the actual state of the operation is not known yet
- COMPLETE: The ISVC has been correctly removed, the deployment is removed - completed succesfully and not running anymore
- NEW: Not used for now

### Setup cluster

1. Install ODH 2.4.0
2. Install ODH component using the following DataScienceCluster CRD
```yaml
kind: DataScienceCluster
apiVersion: datasciencecluster.opendatahub.io/v1
metadata:
  name: default
  labels:
    app.kubernetes.io/name: datasciencecluster
    app.kubernetes.io/instance: default
    app.kubernetes.io/part-of: opendatahub-operator
    app.kubernetes.io/managed-by: kustomize
[...]

PR #135, opened 2024-01-04
Implement model registry inference service reconciliation

lampajr:
Fixes https://github.com/opendatahub-io/model-registry/issues/249

## Description
Implement the model registry InferenceService reconciliation, the implemented workflow is the opposite of the one proposed in https://github.com/opendatahub-io/odh-model-controller/pull/124. Here the direction goes _from_ `Cluster` _to_ `ModelRegistry`.

The new reconciler will monitor `InferenceService` CRs having pre-defined `labels`, based on those `labels` will sync the model registry by keeping track of every deployment that occurred in the cluster.
Then will update the `InferenceService` CR by linking it to the model registry record using a specific `label`.

### e2e tests

Added e2e tests under `test/e2e/` folder, currently I am running those tests using Kind cluster:
1. Setup cluster: `kind create cluster --config test/config/kind-e2e-config.yaml`
2. Create temporary dev tag image: `IMG=quay.io/$USER/odh-model-controller:$(git rev-parse HEAD)`
3. Build and tag dev image: `make IMG=$IMG docker-build`
4. Add local image to cluser: `kind load docker-image $IMG`
5. Deploy the controller: `make IMG=${IMG} deploy-dev`
6. Run e2e tests: `make e2e-test`

> NOTE: These steps could be (hopefully) easily converted in a GHA if we decide to go with that.

### Setup cluster

1. Install ODH 2.4.0
7. Install ODH component using the following DataScienceCluster CRD
```yaml
kind: DataScienceCluster
apiVersion: datasciencecluster.opendatahub.io/v1
metadata:
  name: default
  labels:
    app.kubernetes.io/name: datasciencecluster
    app.kubernetes.io/instance: default
    app.kubernetes.io/part-of: opendatahub-operator
    app.kubernetes.io/created-by: opendatahub-operator
spec:
  components:
    codeflare:
      managementState: Removed
    dashboard:
      managementState: Managed
    datasciencepipelines:
      managementState: Removed
    kserve:
      managementState: Managed
      devFlags:
        manifests:
          - contextDir: config
            sourcePath: overlays/odh
            uri: https://github.com/opendatahub-io/kserve/tarball/master
          - contextDir: config
            uri: https://github.com/lampajr/odh-model-controller/tarball/lampajr20231219_gh249_reconciler_from_isvc_test
    modelmeshserving:
      managementState: Managed
      devFlags:
        manifests:
          - contextDir: config
            sourcePath: overlays/odh
            uri: https://github.com/opendatahub-io/modelmesh-serving/tarball/main
          - contextDir: config
            uri: https://github.com/lampajr/odh-model-controller/tarball/lampajr20231219_gh249_reconciler_from_isvc_test
    ray:
      managementState: Removed
    workbenches:
      managementState: Managed
    trustyai:
      managementState: Removed
```

> NOTE: `lampajr20231219_gh249_reconciler_from_isvc_test` contains manifest changes to make use of the correct odh-model-controller image: alampare/odh-model-controller

By default this new model-registry for model-serving reconciler/controller is disabled in `odh-model-controller`, in order to properly enable it you should just start the controller with `--model-registry-enabled` flag.
`https://github.com/lampajr/odh-model-controller/tarball/lampajr20231219_gh249_reconciler_from_isvc_test` already contains this change in the configuration.

8. Install model registry operator
  1. Clone model-registry-operator repository: `git clone https://github.com/opendatahub-io/model-registry-operator.git`
  2. Install CRDs: `make KUBECTL=oc install`
  3. Deploy the operator: `make KUBECTL=oc deploy`

### Setup DS project

1. Create a Data Science project using the ODH dashboard, e.g., `demo-model-registry-20240104`
2. Setup a data connection (you could install a local minio instance in the newly create DS project)
3. Create a model server using the ODH dashboard
9. Apply the sample model registry CR from the model-registry-operator folder: `kustomize build config/samples | oc apply -f -` (make sure `oc` is pointing to the data science project create in step 1.)
10. Create a new model server using the ODH dashboard, in my case named `model-server`

### Upload models

For the sake of simplicity let's upload some `onnx` files in the local minio instance (bucket name is `models`):
- `models/mnist/v12/mnist-12.onnx`
- `models/mnist/v8/mnist-8.onnx`

### Fill up the model registry

> NOTE: IDs could change depending on the order of operations and on the number of models you are going to register.

0. Set the model registry proxy url, you might have to setup a custom manual route from port `:8080` of the model registry proxy
```bash
MR_HOSTNAME="<endpoint>"
```

1. Register an empty model, e.g, MNIST
```bash
curl --silent -X 'POST' \
  "$MR_HOSTNAME/api/model_registry/v1alpha1/registered_models" \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "description": "MNIST model for recognizing handwritten digits",
  "name": "MNIST",
  "state": "LIVE"
}'
```

2. Add a new model version
```bash
curl --silent -X 'POST' \
  "$MR_HOSTNAME/api/model_registry/v1alpha1/model_versions" \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "description": "MNIST model version n. 8",
  "name": "v8",
  "registeredModelID": "1",
  "state": "LIVE"
}'
```

3. Add the model artifact to the created model version
```bash
curl --silent -X 'POST' \
  "$MR_HOSTNAME/api/model_registry/v1alpha1/model_versions/2/artifacts" \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "description": "Model artifact for MNIST v8",
  "uri": "s3://models/mnist/v8/mnist-8.onnx",
  "state": "UNKNOWN",
  "name": "v8-model",
  "modelFormatName": "onnx",
  "modelFormatVersion": "1",
  "storageKey": "aws-connection-models",
  "storagePath": "mnist/v8",
  "artifactType": "model-artifact"
}'
```

4. [Optional] If for some reason you need to update the artifact you can run:
```bash
curl --silent -X 'PATCH' \
  "$MR_HOSTNAME/api/model_registry/v1alpha1/model_artifacts/1" \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '  {
    "description": "Model artifact for MNIST v8",
    "modelFormatName": "onnx",
    "modelFormatVersion": "1",
    "state": "UNKNOWN",
    "storageKey": "aws-connection-models",
    "storagePath": "mnist/v8",
    "uri": "s3://models/mnist/v8/mnist-8.onnx"
  }
'
```

> NOTE: `modelFormatName`, `modelFormatVersion`, `storageKey` and `storagePath` are all those information needed to create the InferenceService so they should be valid values. E.g., the `storageKey` must match the existing data connection name (which is `aws-connection-{{connection name}}`).

You can use the following curls to inspect model registry content:
```bash
# Get registered models
curl --silent -X 'GET' \
  "$MR_HOSTNAME/api/model_registry/v1alpha1/registered_models?pageSize=100&orderBy=ID&sortOrder=DESC&nextPageToken=" \
  -H 'accept: application/json' | jq '.items'

# Get model versions for a specific registered model
curl --silent -X 'GET' \
  "$MR_HOSTNAME/api/model_registry/v1alpha1/model_versions?pageSize=100&orderBy=ID&sortOrder=DESC&nextPageToken=" \
  -H 'accept: application/json' | jq '.items'

# Get model artifacts for a specific model version
curl --silent -X 'GET' \
  "$MR_HOSTNAME/api/model_registry/v1alpha1/model_versions/3/artifacts?pageSize=100&orderBy=ID&sortOrder=DESC&nextPageToken=" \
  -H 'accept: application/json' | jq '.items'

# Get all inference services stored in model registry
curl --silent -X 'GET' \
  "$MR_HOSTNAME/api/model_registry/v1alpha1/inference_services?pageSize=100&orderBy=ID&sortOrder=DESC&nextPageToken=" \
  -H 'accept: application/json' | jq '.items'
```

### Test model controller workflow

As soon as you create an `InferenceService` CR in the project, the odh-model-controller will create the `ServingEnvironment` in the model registry having the `name == namespace` (i.e., DS project), you can inspect its ID:
```bash
curl --silent -X 'GET' \
  "$MR_HOSTNAME/api/model_registry/v1alpha1/serving_environments?pageSize=100&orderBy=ID&sortOrder=DESC&nextPageToken=" \
  -H 'accept: application/json' | jq '.items'
```

1. Apply the `InferenceService` with proper links to the registered model (`IDs` might be different):
```bash
oc apply -n demo-model-registry-20240104 -f - <<EOF
apiVersion: "serving.kserve.io/v1beta1"
kind: "InferenceService"
metadata:
  name: "mnist-v8-model"
  annotations:
    "openshift.io/display-name": "mnist-v8-model"
    "serving.kserve.io/deploymentMode": "ModelMesh"
  labels:
    "modelregistry.opendatahub.io/registered-model-id": "1"
    "modelregistry.opendatahub.io/model-version-id": "2"
    "opendatahub.io/dashboard": "true"
spec:
  predictor:
    model:
      modelFormat:
        name: onnx
        version: "1"
      runtime: model-server
      storage:
        key: aws-connection-models
        path: mnist/v8/mnist-8.onnx
EOF
```

- `modelregistry.opendatahub.io/registered-model-id` identifies the registered model we would like to deploy
- `modelregistry.opendatahub.io/model-version-id` identifies the specific version to deploy, if omitted (as in this case) let's deploy the latest version for that registered model

> NOTE: the IDs must match, therefore ensure you are providing the corrects IDs.

At this point, the odh-model-controller will monitor `InferenceService` CRs and based on their `labels` will create the corresponding records in model registry

To check the model registry you can run:
```bash
curl --silent -X 'GET' \
  "$MR_HOSTNAME/api/model_registry/v1alpha1/inference_services?pageSize=100&orderBy=ID&sortOrder=DESC&nextPageToken=" \
  -H 'accept: application/json' | jq '.items'
```

Expected output:
```bash
[
  {
    "createTimeSinceEpoch": "1704383784228",
    "customProperties": {},
    "id": "4",
    "lastUpdateTimeSinceEpoch": "1704383784228",
    "modelVersionId": "3",
    "name": "mnist-v8-model/4a2b96b4-af76-4270-bd2d-1a5aed2cd6ad",
    "registeredModelId": "2",
    "runtime": "model-server",
    "servingEnvironmentId": "1",
    "desiredState": "DEPLOYED"
  }
]
```

Check the ISVC has been correctly linked to the newly created record in model registry:
```bash
oc get inferenceservices.serving.kserve.io/mnist-v8-model -o yaml
```

Expected output (note `modelregistry.opendatahub.io/inference-service-id` label):
```yaml
kind: InferenceService
metadata:
  finalizers:
  - modelregistry.opendatahub.io/finalizer
  generation: 1
  labels:
    modelregistry.opendatahub.io/inference-service-id: 4
    modelregistry.opendatahub.io/registered-model-id": 1
    modelregistry.opendatahub.io/model-version-id": 2
    opendatahub.io/dashboard: true
  name: mnist-v8-model
  namespace: demo-model-registry-20240104
spec:
  predictor:
    model:
      modelFormat:
        name: onnx
        version: "1"
      name: ""
      resources: {}
      runtime: model-server
      storage:
        key: aws-connection-models
        path: mnist/v8
```

2. Delete the `InferenceService` (you can do this using the ODH dashboard as well)
```bash
oc delete -n demo-model-registry-20240104 -f - <<EOF
apiVersion: "serving.kserve.io/v1beta1"
kind: "InferenceService"
metadata:
  name: "mnist-v8-model"
  annotations:
    "openshift.io/display-name": "mnist-v8-model"
    "serving.kserve.io/deploymentMode": "ModelMesh"
  labels:
    "modelregistry.opendatahub.io/registered-model-id": "1"
    "modelregistry.opendatahub.io/model-version-id": "2"
    "opendatahub.io/dashboard": "true"
spec:
  predictor:
    model:
      modelFormat:
        name: onnx
        version: 1
      runtime: model-server
      storage:
        key: aws-connection-models
        path: mnist/v8
EOF
```

Check record in model registry is correctly update, note the `UNDEPLOYED` state:
```bash
curl --silent -X 'GET' \
  "$MR_HOSTNAME/api/model_registry/v1alpha1/inference_services?pageSize=100&orderBy=ID&sortOrder=DESC&nextPageToken=" \
  -H 'accept: application/json' | jq '.items'
```

Expected output:
```yaml
[
  {
    "createTimeSinceEpoch": "1704383784228",
    "customProperties": {},
    "desiredState": "UNDEPLOYED",
    "id": "4",
    "lastUpdateTimeSinceEpoch": "1704384020813",
    "modelVersionId": "3",
    "name": "mnist-v8-model/4a2b96b4-af76-4270-bd2d-1a5aed2cd6ad",
    "registeredModelId": "2",
    "runtime": "model-server",
    "servingEnvironmentId": "1"
  }
]
```

> **NOTE** step `Upload models` and `Fill up the model registry` could be automated using either notebooks or pipelines, please follow a more complete e2e demo at https://github.com/tarilabs/demo20231121#demo 

## Merge criteria:


lampajr:
As agreed we decided to implement this workflow where the model-controller has a more "passive" behavior with respect to the model registry reconcilication.

As described in the PR description, the new reconciler will monitor the `InferenceService` CRs that have proper labels defined and based on those values it will sync the deployments to model registry in the form of model registry InferenceService objects.

> NOTE: under `test/` folder I setup some e2e tests that I was used to run locally (added step-by-step instructions in the description) - if you think it could make sense I can add a github workflow to automatically run them using Kind

test/data/deploy/inference-service-to-delete.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: serving.kserve.io/v1beta1
+kind: InferenceService
+metadata:
+  name: dummy-inference-service
+  namespace: default
+  labels:
+    "mr-inference-service-id": "4"
+  finalizers:
```
Jooho, line 8: This finalizer will be added to the isvc created by mr controller? The finalizer of the isvc will not be removed by mr controller, right?

test/data/deploy/inference-service-with-model-version.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: serving.kserve.io/v1beta1
+kind: InferenceService
+metadata:
+  name: dummy-inference-service
+  namespace: default
+  labels:
+    "mr-registered-model-id": "1"
+    "mr-model-version-id": "2"
```
Jooho, line 8: This usecase does not have a finalizer. when the finalizer will be added?

test/data/deploy/inference-service-without-model-version.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: serving.kserve.io/v1beta1
+kind: InferenceService
+metadata:
+  name: dummy-inference-service
+  namespace: default
+  labels:
+    "mr-registered-model-id": "1"
```
Jooho, line 7: if the model version is not added, it means that the version is `1` ?

test/data/deploy/inference-service-to-delete.yaml
lampajr: mmm not exactly, the finalizer will be added my the odh-model-controller itself (see https://github.com/opendatahub-io/odh-model-controller/pull/135/files#diff-e628c4483237f3dba685d802623c535845221a01b2bd2e3782a82ab9df48320aR120-R129) and then it will be removed by the model controller as well.

The ISVC are going to be created by the dashboard as it is right now

test/data/deploy/inference-service-without-model-version.yaml
lampajr: if the modelVersion is not added it means the latest version in that moment

Makefile
```diff
@@ -66,7 +66,11 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" POD_NAMESPACE=default go test ./... -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" POD_NAMESPACE=default go test ./controllers/... -coverprofile cover.out
+
+.PHONY: e2e-test
+e2e-test: manifests generate fmt vet ## Run e2e-tests.
+	POD_NAMESPACE=default go test ./test/e2e/...
```
spolti, line 73: Is it going to be executed only by CI?
If not, we could make it configurable, not sure if use the **default** as namesapce is a good practice.

config/crd/external/kustomization.yaml
```diff
@@ -1,3 +1,9 @@
 resources:
 - serving.kserve.io_inferenceservices.yaml
-# - route.openshift.io_routes.yaml  #minikube
+- serving.kserve.io_servingruntimes.yaml
+- route.openshift.io_routes.yaml
+- monitoring.coreos.com_podmonitors.yaml
+- monitoring.coreos.com_servicemonitors.yaml
+- maistra.io_servicemeshmemberrolls.yaml
+- maistra.io_servicemeshmembers.yaml
+- telemetry.istio.io_telemetries.yaml
```
spolti, line 9: are all these resources needed?

Makefile
lampajr: No, right now e2e tests are not going to be executed by the CI as there is no CI for e2e tests AFAICS - but the way I implemented it they should be as flexible as possible, such that we can also create a `Kind`-based github action to trigger them.

The only precondition is that the `odh-model-controller` must be deployed/running before triggering the `make e2e-test`.

config/rbac/role.yaml
```diff
@@ -169,6 +169,7 @@ rules:
   verbs:
   - get
   - list
+  - update
```
spolti, line 172: hmm, which rbac needs to be updated?

I am not about  operators updating rbac. :)

config/crd/external/kustomization.yaml
lampajr: Actually I had to add all of them to make e2e tests working in the Kind cluster (i.e., they are used just for `deploy-dev` and `deploy-e2e`)  - not sure if we want to move them directly in the overlay/e2e config folder

config/rbac/role.yaml
lampajr: The reconcile loop has to update InferenceService CRs, adding some ad-hoc labels to keep the links between the CR and the record in model registry - that's why I had to add that `update`. Is there any other way I can do that?

lampajr:
fyi, with the latest commit I rebased on main and updated the model registry library to the latest one

lampajr:
Hey @spolti @Jooho, we finally got model registry operator into the opendatahub-operator incubation branch [1] so we would like to get this merged as well. Could you please review this as soon as you have some time?

> NOTE that this new functionality is disabled by default unless the new cmd line arg is provided, so if not enabled it should not affect anything already existing

[1] https://github.com/opendatahub-io/opendatahub-operator/pull/776

main.go
```diff
@@ -169,6 +173,21 @@ func main() {
 			"monitoring for ModelServing, please provide a monitoring namespace via the (--monitoring-namespace) flag.")
 	}
 
+	if enableMRInferenceServiceReconcile {
```
israel-hdez, line 176: Rather than setting up a reconciler from scratch, I wonder if it would be better to add a child-reconciler in the existing InferenceService reconciler, in this chunk:

https://github.com/opendatahub-io/odh-model-controller/blob/062439d7e7de72dcea37ba7238c94f29f1273abe/controllers/inferenceservice_controller.go#L89-L96

test/e2e/modelregistry_controller_e2e_test.go
```diff
@@ -0,0 +1,343 @@
...
+			inferenceService, err = modelRegistryClient.UpsertInferenceService(&openapi.InferenceService{
+				Name:                 &versionName,
+				DesiredState:         openapi.INFERENCESERVICESTATE_DEPLOYED.Ptr(),
+				RegisteredModelId:    *registeredModel.Id,
+				ServingEnvironmentId: "3",
+			})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(inferenceService.GetId()).To(Equal("4"))
+
+			_, err = utils.Run(exec.Command("kubectl", "apply", "-f", ServingRuntimePath1))
```
israel-hdez, line 179: If possible, use an environment variable for `kubectl`. Later, we may run this in an environment where only `oc` may be available.

test/data/deploy/inference-service-with-model-version.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: serving.kserve.io/v1beta1
+kind: InferenceService
```
israel-hdez, line 2: I don't see a test where `mr-registered-model-id` is absent and `mr-model-version-id` is present.
If this condition is not valid, the reconciler should properly validate it and there should be a test.

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,104 @@
...
+	ctx    context.Context
+	cancel context.CancelFunc
+	cli    client.Client
+	kc     *k8sclient.Clientset
+)
+
+// Run e2e tests using the Ginkgo runner.
+// ODH model controller should be already running in the cluster where
+// these tests will run against
+func TestE2E(t *testing.T) {
```
israel-hdez, line 52: After looking at the tests, I think _kind_ to run the E2Es should not be required.

Probably, these E2Es can run with a similar setup as the one described here: https://book.kubebuilder.io/cronjob-tutorial/writing-tests.

It will be great if you can figure out how to run the tests without Kind.

config/overlays/dev/kustomization.yaml
```diff
@@ -1,3 +1,38 @@
 resources:
+  - namespace.yaml
```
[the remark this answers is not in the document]
lampajr: >I'm not sure if that was the intended outcome of make deploy-dev

Oh yeah I don't know what was the original motivation behind that, if that wasn't the original one I can revert the changes affecting `overlays/dev`

controllers/constants/constants.go
```diff
@@ -21,3 +21,12 @@ const (
...
 	IstioIngressServiceHTTPSPortName = "https"
 )
+
+// model registry
+const (
+	MLMDAddressEnv                       = "MLMD_ADDRESS"
+	ModelRegistryNamespaceLabel          = "mr-namespace"
+	ModelRegistryInferenceServiceIdLabel = "mr-inference-service-id"
+	ModelRegistryModelVersionIdLabel     = "mr-model-version-id"
+	ModelRegistryRegisteredModelIdLabel  = "mr-registered-model-id"
```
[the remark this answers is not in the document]
lampajr: yeah I agree, much more meaningful and organized this way

test/e2e/e2e_suite_test.go
lampajr: Well, IIUC how `envtest` works I cannot startup a real service to interact with (in this case the model registry service) - that's why those tests leverages on an existing cluster up and running (either with `Kind` or any other tool).

Do you have any particular idea on how I can startup and interact with a real service?

test/e2e/modelregistry_controller_e2e_test.go
lampajr: Makes sense, much more flexible

controllers/mr_inferenceservice_controller.go
```diff
@@ -0,0 +1,414 @@
...
+		is, err = mr.GetInferenceServiceById(isId)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to find InferenceService with id %s in model registry: %w", isId, err)
+		}
+	} else if okRegisteredModelId || okModelVersionId {
+		// No corresponding InferenceService in model registry, create new one
+		is, err = r.createMRInferenceService(log, mr, isvc, *servingEnvironment.Id, registeredModelId, modelVersionId)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
```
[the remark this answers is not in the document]
tarilabs: > If it is important to write a record in the MR about that a model got deployed while the MR was down, then all is good.

I'd lean into this use-case, as the purpose of the InferenceService entities in MR is a track record of audit log of how the model "was used". 

My understanding is that the reconcile loop will try until MR comes back up and so a record of InferenceService will then be tracked and marked as "archived".

controllers/mr_inferenceservice_controller.go
```diff
@@ -0,0 +1,414 @@
...
+	} else {
+		// No expected labels set in the ISVC
+		log.Error(fmt.Errorf("missing label, unable to link ISVC to Model Registry, skipping InferenceService"), "Stop ModelRegistry InferenceService reconciliation")
+		return ctrl.Result{}, nil
+	}
```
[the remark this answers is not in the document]
lampajr: That's right, this was a leftover from the initial implemention. Removed

controllers/mr_inferenceservice_controller.go
```diff
@@ -0,0 +1,414 @@
...
+func (r *ModelRegistryInferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// Initialize logger format
+	log := r.log.WithValues("ModelRegistryInferenceService", req.Name, "namespace", req.Namespace)
+
+	// Get the InferenceService object when a reconciliation event is triggered (create,
+	// update, delete)
+	isvc := &kservev1beta1.InferenceService{}
+	err := r.client.Get(ctx, req.NamespacedName, isvc)
+	if err != nil && apierrs.IsNotFound(err) {
+		log.Error(err, "Stop ModelRegistry InferenceService reconciliation")
```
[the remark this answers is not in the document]
lampajr: Done

controllers/mr_inferenceservice_controller.go
```diff
@@ -0,0 +1,414 @@
...
+				return ctrl.Result{}, err
+			}
+		}
+		return ctrl.Result{}, nil
+	} else {
+		// Update the ISVC label, set the newly created IS id if not present yet
+		desired := isvc.DeepCopy()
+		desired.Labels[constants.ModelRegistryInferenceServiceIdLabel] = *is.Id
+		delete(desired.Labels, constants.ModelRegistryRegisteredModelIdLabel)
+		delete(desired.Labels, constants.ModelRegistryModelVersionIdLabel)
```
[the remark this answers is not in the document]
lampajr: Actually there was not a valid reason, it was mainly for cleaning but I agree that keeping them is much more valuable.

controllers/mr_inferenceservice_controller.go
```diff
@@ -0,0 +1,414 @@
...
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ModelRegistryInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	builder := ctrl.NewControllerManagedBy(mgr).
+		For(&kservev1beta1.InferenceService{}).
+		Owns(&kservev1alpha1.ServingRuntime{}).
+		Owns(&corev1.Namespace{}).
+		Owns(&monitoringv1.PodMonitor{}).
```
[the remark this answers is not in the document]
lampajr: That's a good point as this reconciliation is just interested in the `InferenceService` itself, it does not manage nor monitor child/parent resources. Hence I agree on removing those owns :rocket:

controllers/mr_inferenceservice_controller.go
```diff
@@ -0,0 +1,414 @@
...
+				}
+
+				if len(inferenceServicesList.Items) == 0 {
+					r.log.Info("No InferenceServices found for Serving Runtime: " + o.GetName())
+					return []reconcile.Request{}
+				}
+
+				reconcileRequests := make([]reconcile.Request, 0, len(inferenceServicesList.Items))
+				for _, inferenceService := range inferenceServicesList.Items {
+					reconcileRequests = append(reconcileRequests, reconcile.Request{
```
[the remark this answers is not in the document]
lampajr: This was a leftover from the initial implementation where changes to the `ServingRuntime` were triggering the reconciliation loop but right now we just monitor the `InferenceService` CRs.

In conclusion I was able to simplify to just `For(&kservev1beta1.InferenceService{})`

controllers/mr_inferenceservice_controller.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+
+	if isMarkedToBeDeleted {
+		err := r.onDeletion(mr, log, isvc, is)
+		if err != nil {
+			return ctrl.Result{Requeue: true}, err
+		}
+
+		if controllerutil.ContainsFinalizer(isvc, modelRegistryFinalizer) {
+			log.Info("InferenceService going to be deleted from cluster, setting desired state to UNDEPLOYED in model registry")
```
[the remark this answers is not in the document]
lampajr: yeah I agree that this could be a bit confusing, moved this log into the `onDeletion` call

controllers/mr_inferenceservice_controller.go
```diff
@@ -0,0 +1,414 @@
...
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	mr, err := core.NewModelRegistryService(conn)
+	return mr, conn, err
+}
+
+func IgnoreDeletingErrors(err error) error {
```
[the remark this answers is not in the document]
lampajr: Good catch, I think I should fail in case of conflict and retry removing the finalizer in the next loop.

controllers/mr_inferenceservice_controller.go
```diff
@@ -0,0 +1,414 @@
...
+		return nil, nil, nil
+	}
+
+	// Setup model registry service
+	log.Info("Connecting to " + mlmdAddr)
+	conn, err := grpc.DialContext(
+		ctxTimeout,
+		mlmdAddr,
+		grpc.WithReturnConnectionError(),
+		grpc.WithBlock(),
```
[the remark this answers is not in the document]
lampajr: Hey @israel-hdez , thanks for this pointer.. that's interesting, based on the link you provided it seems that the main reasons it is discouraged is that:
1. The connection check is done at dial time ([ref](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#using-failonnontempdialerror-withblock-and-withreturnconnectionerror:~:text=The%20options%20FailOnNonTempDialError%2C%20WithBlock%2C%20and%20WithReturnConnectionError%20are%20designed%20to%20handle%20some%20of%20these%20errors%2C%20but%20they%20do%20so%20by%20relying%20on%20failures%20at%20dial%20time)) therefore there is no guarantee the connection won't fail in subsequent requests --> We are already managing this as every time we perform a request we check if that returned an error or not (either logic error or connection one)
2. Wait for connection forever ([ref](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#using-failonnontempdialerror-withblock-and-withreturnconnectionerror:~:text=if%20a%20client%20uses%20WithBlock%20to%20wait%20for%20a%20connection%20to%20be%20established%2C%20it%20may%20end%20up%20waiting%20indefinitely%20if%20the%20server%20is%20not%20responding)) --> we are already managing this case by using a context timeout to avoid infinite wait time.

wdyt?

test/data/deploy/inference-service-with-model-version.yaml
lampajr: Yeah, the condition you mentioned is not valid - refactoring a bit the code to make it much more explicit and adding a test

controllers/mr_inferenceservice_controller.go
```diff
@@ -0,0 +1,414 @@
...
+	is := &openapi.InferenceService{
+		Name:                 &isName,
+		ServingEnvironmentId: servingEnvironmentId,
+		RegisteredModelId:    registeredModelId,
+		ModelVersionId:       modelVersionIdPtr,
+		Runtime:              isvc.Spec.Predictor.Model.Runtime,
+		DesiredState:         openapi.INFERENCESERVICESTATE_DEPLOYED.Ptr(),
+	}
+
+	return mr.UpsertInferenceService(is)
```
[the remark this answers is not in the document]
lampajr: That's great catch, I wasn't considering that case - as is it would have failed trying to create an InfService in Model Registry as already existing another one with the same name.

Fixing this logic by checking the existance of the record before creating it.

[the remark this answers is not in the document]
lampajr: Fixed.

test/data/deploy/inference-service-with-model-version.yaml
lampajr: Added a test covering this case as well

config/crd/external/kustomization.yaml
VedantMahabaleshwarkar: @lampajr we have a the `overlays/dev` directory for this purpose (for local testing etc). I think adding these CRDs here is not necessary and IMO if these are needed to make local testing work they should be added to the `overlays/dev/kustomization.yaml` file and apply that overlay for local testing

config/overlays/dev/kustomization.yaml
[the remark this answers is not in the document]
VedantMahabaleshwarkar: I don't think it's a good idea to duplicate the contents of these overlays as it'll increase the difficulty of maintaining changes. For the `deploy-dev` installation, the `overlays/dev/kustomization.yaml` can call the `overlays/odh/kustomization.yaml` directly to avoid duplication

config/overlays/e2e/kustomization.yaml
```diff
@@ -0,0 +1,38 @@
+resources:
```
VedantMahabaleshwarkar, line 1: I'd want to avoid a separate overlay for e2e tests, I think we should aim to combine this with the `overlays/dev` directory since they have the same purpose

VedantMahabaleshwarkar (changes requested):

controllers/mr_inferenceservice_controller.go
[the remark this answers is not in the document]
israel-hdez: > My understanding is that the reconcile loop will try until MR comes back up and so a record of InferenceService will then be tracked and marked as "archived".

Right, that's what I understand from the code: if the registry was down while the ISVC was both created and deleted, this will retry until the registry is up again and the records in the registry can be created.

The only point that end-users may see weird is that the ISVC will be stuck in `Terminating` state until the registry recovers. But given there is no messaging layer that can help with this, I don't see another way around it.

So, let's go with this as is.

controllers/mr_inferenceservice_controller.go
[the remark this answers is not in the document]
israel-hdez: So, in the updated code, I see you removed both `WithReturnConnectionError` and `WithBlock`, and I no longer see ctxTimeout. To me is better now, as we follow the recommendations.

Just to answer your thoughts:
1. Yes, if you are handling errors, I see OK to keep `WithReturnConnectionError`.
2. The (2) was my main concern, as I don't know if `WithBlock` honors the timeout, or if it blocks indefinitely regardless of the configured timeout in the Context. The [api docs](https://pkg.go.dev/google.golang.org/grpc#WithBlock) don't clarify it, so I am unsure of the actual behavior.

test/e2e/e2e_suite_test.go
israel-hdez: As we briefly discussed via chat, my only idea is to use podman/docker to start the model registry and declaring the `MLMD_ADDRESS` environment variable to the right value before running the tests. This would simplify the setup to run the E2Es and the tests may run faster in environments like GitHub Actions.

Indeed, the main motivation for suggesting moving away from Kind was because I was thinking that a subsequent PR would create GitHub Actions workflow to run these new tests, but I think the setup depends on the platform that we are going to choose for running them. Also, I'm not sure if moving away from Kind would lead to more complex code in `BeforeEach` and `AfterEach` functions.

As an aside, to me the tests look closer to units than to E2Es, because the tests need to pre-seed the registry. The way I perceive an E2E test is by using the actual components that would interact on the whole flow. So, this was an aside motivation for suggesting a simpler setup. But this may just be my own expectation for E2Es, so this may not be valid :-)

So, this is just feedback. I don't see a mandatory change here :slightly_smiling_face:, nor this should be a blocker for merging. What is important is to have tests, and we have them! We can stay with Kind for this PR and, if needed, we can later change.

israel-hdez (commented):
@lampajr I'm good with the changes.

I haven't tried this live. I'll try it on my Friday.

I think the only pending comment is the one from Vedant about the manifests. Let me know when you are ready. If I don't see any issue in my live testing, I'll merge once you confirm this is ready to go.

config/overlays/e2e/kustomization.yaml
lampajr: with latest commit I completely removed the `e2e` overlay and I started relying on the `dev` one.

config/crd/external/kustomization.yaml
VedantMahabaleshwarkar: @lampajr looking at the manifests again, if the `crd/external` directory is not in the call chain for the actual deployment, then it's fine to add these here. We just want to avoid applying these CRDs on the cluster outside of development testing scenarios.

test/e2e/e2e_suite_test.go
VedantMahabaleshwarkar: I'd agree with Edgar's suggestion here. I'd want to stay away from prescribing an environment where the E2E tests should run and it would be good to have e2e tests be platform agnostic/configurable. But also agreed that it can be a future enhancement

lampajr: I agree with you guys, I tried to make them as generic as possible such that they could be run against an existing cluster (either using Kind or minikube or openshift-ci) - obviously I tested it using Kind only but I don't see why it shouldn't work in other "simulated" or "real" clusters as the only precondition is that the `odh-model-controller` must be running in that cluster.

On the other hand if we can find a way to make those tests running in GHA w/o the need to setup a cluster is much better, I definitely agree with this. Therefore I will create a followup ticket/issue for this enhancement!

lampajr:
Hi @israel-hdez @VedantMahabaleshwarkar , thanks a lot for all your suggestions!!

I think I was able to address all of them, I also tested again in a real cluster and it worked for me (following the instructions I provided in the PR description).

config/overlays/dev/kustomization.yaml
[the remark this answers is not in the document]
lampajr: yeah I couldn't find any reference to that `crd/external/kustomization.yaml`, at least not in this repo but I don't know if they could be referenced from other places outside this repo itself.

test/e2e/modelregistry_controller_e2e_test.go
```diff
@@ -0,0 +1,414 @@
...
+	var grpcPort *int32
+	for _, port := range mrServiceList.Items[0].Spec.Ports {
+		if port.Name == "grpc-api" {
+			grpcPort = &port.NodePort
+			break
+		}
+	}
+	Expect(grpcPort).ToNot(BeNil())
+
+	mlmdAddr = fmt.Sprintf("localhost:%d", *grpcPort)
```
israel-hdez, line 314: I had to change the code because of this connection to localhost.
This is somewhat specific to Kind. I was able to workaround by doing a port-forward, but given the tests re-deploy the registry, the port-forward needs to be re-opened before each test.

israel-hdez (approved):
I was able to try this and runs fine.
This is low risk because it is turned off by default. So, I'll merge.

I left one comment for your consideration for later improvement.

test/e2e/modelregistry_controller_e2e_test.go
lampajr: Interesting, thanks for pointing this out! I will keep track of this in a separate issue :muscle:

PR #136, opened 2024-01-10
fix: Remove NetworkPolicies created to relax SMCP policies

terrytangyuan:
## Description

Fixes https://issues.redhat.com/browse/RHOAIENG-1015. 

## How Has This Been Tested?

See https://docs.google.com/document/d/1l5FEtFRc05Nifh6IcKpd3uNH3c5fhwz0gdZtkYZL0B4/edit?usp=sharing for instructions.

## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #145, opened 2024-01-18
fix: adds network policies to unblock traffic between components 

bartoszmajsak:
## Description

This solution is based on an initially proposed [script](https://github.com/bartoszmajsak/issue-work-logs/tree/main/RHOAIENG-1003), but with slightly modified namespace selectors.

### Additional changes

- provides a composed matcher to inspect nested structs easily
- cleans up test setup for KServe
- renames variables in reconcilers to reflect actual objects used

Fixes https://issues.redhat.com/browse/RHOAIENG-1955

## How Has This Been Tested?

I included a simple integration test following the existing setup in the project.

I would like to get some help in testing it e2e, perhaps based on https://github.com/rh-aiservices-bu/fraud-detection/blob/main/setup/setup-s3.yaml which revealed the issue. Is there some other repo with e2e tests I could use it for?

## Merge criteria:


controllers/reconcilers/kserve_networkpolicy_reconciler.go
```diff
@@ -98,47 +113,128 @@ func (r *KserveNetworkPolicyReconciler) createDesiredResource(isvc *kservev1beta
...
+		},
+	}
+}
+
+// allowTrafficFromApplicationNamespaces creates combined network policy applied to pods in InferenceService namespace.
+// This set of policies allow traffic from:
+// - application namespace, where OpenDataHub and component service are deployed
+// - namespaces created by OpenDataHub where its component live
+// - traffic from other DataScienceProjects (namespaces created through dashboard)
+func (r *KserveNetworkPolicyReconciler) allowTrafficFromApplicationNamespaces(isvc *kservev1beta1.InferenceService) *v1.NetworkPolicy {
```
bartoszmajsak, line 155: @zdtsw can you help me confirm if these are all the labels we need?

controllers/reconcilers/kserve_networkpolicy_reconciler.go
```diff
@@ -30,14 +31,19 @@ import (
...
 )
 
+var definedNetworkPolicies = []string{monitoringNetworkPolicyName, openshiftIngressNetworkPolicyName, opendatahubNamespacesNetworkPolicyName}
+
 type KserveNetworkPolicyReconciler struct {
 	client               client.Client
 	scheme               *runtime.Scheme
 	networkPolicyHandler resources.NetworkPolicyHandler
 	deltaProcessor       processors.DeltaProcessor
+	appNamespace         string
```
bartoszmajsak, line 46: I need to know where this ctrl was deployed as part of RHOAI, as we need a policy allowing traffic from this ns - otherwise components like pipelines can fail. See this [comment](https://issues.redhat.com/browse/RHOAIENG-1003?focusedId=23691645&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-23691645)

israel-hdez (commented):
Codewise,  I'm not seeing anything wrong.

Tomorrow I will check if we are covered on the bad cases we found.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
```diff
@@ -98,47 +113,128 @@ func (r *KserveNetworkPolicyReconciler) createDesiredResource(isvc *kservev1beta
...
+							},
+						},
+						{
+							NamespaceSelector: &metav1.LabelSelector{
+								MatchLabels: map[string]string{"kubernetes.io/metadata.name": r.appNamespace},
+							},
+						},
+						{
+							NamespaceSelector: &metav1.LabelSelector{
+								MatchLabels: map[string]string{"opendatahub.io/generated-namespace": "true"},
```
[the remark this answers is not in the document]
bartoszmajsak: These are labels added by odh-operator to every ns it creates.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
```diff
@@ -98,47 +113,128 @@ func (r *KserveNetworkPolicyReconciler) createDesiredResource(isvc *kservev1beta
...
+func (r *KserveNetworkPolicyReconciler) allowOpenshiftIngressPolicy(isvc *kservev1beta1.InferenceService) *v1.NetworkPolicy {
+	return &v1.NetworkPolicy{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "NetworkPolicy",
+			APIVersion: "networking.k8s.io/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      openshiftIngressNetworkPolicyName,
+			Namespace: isvc.Namespace,
+			Labels:    map[string]string{"opendatahub.io/related-to": "RHOAIENG-1003"},
```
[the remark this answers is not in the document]
bartoszmajsak: Yeah, I thought it can help in deprecation / upgrade moving fwd. Won't harm I guess.

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -19,50 +19,61 @@ import (
...
+			Expect(cli.Create(ctx, testNamespace)).Should(Succeed())
 
 			inferenceServiceConfig := &corev1.ConfigMap{}
-			err := convertToStructuredResource(InferenceServiceConfigPath1, inferenceServiceConfig)
-			Expect(err).NotTo(HaveOccurred())
-			Expect(cli.Create(ctx, inferenceServiceConfig)).Should(Succeed())
+			Expect(convertToStructuredResource(InferenceServiceConfigPath1, inferenceServiceConfig)).To(Succeed())
+			if err := cli.Create(ctx, inferenceServiceConfig); err != nil && !errors.IsAlreadyExists(err) {
+				Fail(err.Error())
+			}
```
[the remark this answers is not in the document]
bartoszmajsak: Hey, that's cool. I learned something today :) Totally forgot about `WithTransform` existence. 

Generally speaking I'm on the verge as far as "easy to understand/more readable" is concerned.

controllers/reconcilers/kserve_metrics_servicemonitor_reconciler.go
```diff
@@ -102,33 +102,33 @@ func (r *KserveMetricsServiceMonitorReconciler) getExistingResource(ctx context.
 	return r.serviceMonitorHandler.FetchServiceMonitor(ctx, log, types.NamespacedName{Name: getMetricsServiceMonitorName(isvc), Namespace: isvc.Namespace})
 }
 
-func (r *KserveMetricsServiceMonitorReconciler) processDelta(ctx context.Context, log logr.Logger, desiredService *v1.ServiceMonitor, existingService *v1.ServiceMonitor) (err error) {
+func (r *KserveMetricsServiceMonitorReconciler) processDelta(ctx context.Context, log logr.Logger, desiredServiceMonitor *v1.ServiceMonitor, existingServiceMonitor *v1.ServiceMonitor) (err error) {
```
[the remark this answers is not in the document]
bartoszmajsak: De nada! :)

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -87,11 +98,64 @@ var _ = Describe("The Openshift Kserve model controller", func() {
...
+			// then
+			By("ensuring that the controller has created required network policies")
+			networkPolicies := &v1.NetworkPolicyList{}
+			Eventually(func() []v1.NetworkPolicy {
+				err := cli.List(ctx, networkPolicies, client.InNamespace(inferenceService.Namespace))
+				if err != nil {
+					Fail(err.Error())
+				}
+				return networkPolicies.Items
+			}, time.Second*1, interval).Should(HaveLen(3))
```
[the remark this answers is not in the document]
bartoszmajsak: Copy-pastie mistake. Guilty as charged. I actually reworked it this morning in f4880772598413fe6f4621e67e076d409d08f43c, thanks for spotting it.

bartoszmajsak:
> Tomorrow I will check if we are covered on the bad cases we found.

Is there some way we could automate this?

controllers/reconcilers/kserve_networkpolicy_reconciler.go
```diff
@@ -98,47 +113,128 @@ func (r *KserveNetworkPolicyReconciler) createDesiredResource(isvc *kservev1beta
...
+		},
+		Spec: v1.NetworkPolicySpec{
+			PodSelector: metav1.LabelSelector{},
+			Ingress: []v1.NetworkPolicyIngressRule{
+				{
+					From: []v1.NetworkPolicyPeer{
+						{
+							NamespaceSelector: &metav1.LabelSelector{
+								MatchLabels: map[string]string{"network.openshift.io/policy-group": "ingress"},
```
ReToCode, line 139: Just for curiosity, would not traffic always come via istio-ingressgateway anyway? Or do you create additional/other routes directly pointing to the ISVC?

controllers/reconcilers/kserve_networkpolicy_reconciler.go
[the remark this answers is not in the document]
ReToCode: Can the user create the namespace himself somehow? Or use an existing one to deploy ISVC to?

controllers/reconcilers/kserve_networkpolicy_reconciler.go
bartoszmajsak: It's for other coexisting components which create their own routes but are not part or istio.

ReToCode (commented):
I like the approach!

controllers/reconcilers/kserve_networkpolicy_reconciler.go
```diff
@@ -30,14 +31,19 @@ import (
 )
 
 const (
-	networkPolicyName = "allow-from-openshift-monitoring-ns"
+	monitoringNetworkPolicyName            = "allow-from-openshift-monitoring-ns"
```
zdtsw, line 34:
```suggestion
	uwmNWPName                           = "allow-from-uwm-ns"
	openshiftIngressNWPName      = "allow-from-openshift-ingress-ns"
	opendatahubOwnedNWPName            = "allow-from-opendatahub-ns"
```

controllers/reconcilers/kserve_networkpolicy_reconciler.go
zdtsw: these should have all namespaces covered: either created by ODH operator or DS projects done by UI.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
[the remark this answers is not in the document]
zdtsw: @israel-hdez `opendatahub.io/dashboard` is only applied on the DS projects.
if we want traffic going a bit wider by operator, application, and notebook's, better to have the later 2 added.
normally, the 3rd one should have the 2nd one covered, but in case, the label does not exsit (....)

controllers/reconcilers/kserve_networkpolicy_reconciler.go
vaibhavjainwiz: To fetch the value of ctrl namespace we could directly use `constants.KServeNamespace`. We could remove this extra function from utils.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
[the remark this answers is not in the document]
vaibhavjainwiz: These type of labels or comment on code doesn't hurt the functionality but it pollute the overall code quality. I would say we could create Jira for its deprecation in future and add it to backlog so that it doesn't miss in going forward.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
bartoszmajsak: I can't find `constants.KServeNamespace` in the code base, but I wonder how constant can handle something which is dynamic in its nature. I mean it can be `rhods-applications` or `opendatahub` for example.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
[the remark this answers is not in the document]
bartoszmajsak: Let me give you a bit more elaborated reason for the existence of this label. At some point we will provide a better-suited solution when it comes to network policies (that's actually how we started). That would mean not creating defaults on OSSM side, but also then these will go away (as they're solely to mitigate restrictions imposed by OSSM default ones). 

Considering the upgrade path with existing Data Science Projects and with those policies already existing we should take care of removing them. One way I can think of is to delete all the instances  of these special `NetworkPolicies` during the odh-operator upgrade process. Using labels makes it very straightforward to fetch all at once. Another solution (done in #136 ) is to keep their names in the code  and ensure they're removed as part of the reconciliation loop. 

> pollute the overall code quality

Can you elaborate a bit more on this? Would like to understand how I can make it better. One thing coming to mind is to extract it to a `const`.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
bartoszmajsak: I can guess what `uwm` stands for but would like to know the reason for this change :) 

Important to note is that it's already in use, so renaming it would result in adding new policy under "uwm" name with the same rules.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
vaibhavjainwiz: `constants` is coming from "github.com/kserve/kserve/pkg/constants"

bartoszmajsak: Ah found it. `getEnvOrDefault("POD_NAMESPACE", "kserve")` so in fact a `var`, that's what got me confused :) 

If we can assume this env variable is always there that will do. Thanks for sharing.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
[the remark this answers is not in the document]
bartoszmajsak: @ReToCode I guess user can do anything, but I am not sure this is supported scenario. Those policies take into account only the scenarios defined in godoc for this func:

```
// - application namespace, where OpenDataHub and component services are deployed
// - namespaces created by OpenDataHub where components live
// - traffic from other DataScienceProjects (namespaces created through dashboard)
```

I am not sure if the scenario where user creates a namespace on their own is something we should support / cover here. @danielezonca could you confirm?

controllers/reconcilers/kserve_networkpolicy_reconciler.go
bartoszmajsak: Fixed in 7e30338

Thanks for sharing, I didn't know about its existence. Only about the `POD_NAMESPACE` env variable as such.

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -19,56 +19,66 @@ import (
 	"context"
 	kservev1alpha1 "github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
```
skonto, line 20: I would recommend to separate imports for better readability as a standard best practice.
Also sorting affects package initialization, first third party then local ones etc.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
```diff
@@ -19,6 +19,7 @@ import (
 	"context"
 	"github.com/go-logr/logr"
 	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
+	"github.com/kserve/kserve/pkg/constants"
```
skonto, line 22: I would recommend to separate imports for better readability as a standard best practice.
Also sorting affects package initialization, first third party then local ones etc.

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -19,56 +19,66 @@ import (
...
 	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
-	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/v2"
```
skonto, line 22: Should not this be a separate PR or you need some functionality?

controllers/reconcilers/kserve_networkpolicy_reconciler.go
bartoszmajsak: @skonto Yeah I was having the same thoughts. In other projects I use `gci`, so it autoformats things:

```
  gci:
    sections:
      - standard
      - default
      - blank
      - prefix(github.com/opendatahub-io/opendatahub-operator)
      - blank
      - dot
    skip-generated: false
    custom-order: true
```

though that would be too much for the scope of this PR. @Jooho if you folks agree I can donate stuff like that in another PR.

> sorting affects package initialization

TIL :)

controllers/kserve_inferenceservice_controller_test.go
bartoszmajsak: It could be, but as you see it's a rather minimal change due to backward comp of API. Also, note that this version was already pulled in as indirect dep.

It first came when I wanted to use `Ordered` decorator with `BeforeAll`, but gave up on that idea

bartoszmajsak: Though it comes in a separate commit (80af622ae9c8a09f199dfd602875d) so I can slice it if that's preferred. WDYT @Jooho @vaibhavjainwiz?

controllers/reconcilers/kserve_networkpolicy_reconciler.go
skonto: The net policy defined later on has a ns selector `openshift-user-workload-monitoring"` and that means it applies to uwm so people reading this name will expect to apply for openshift-monitoring ns. It is more about consistency I think.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
skonto: Yes it is more like a comment for a separate PR.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
vaibhavjainwiz: KServeNamespace                  = getEnvOrDefault("POD_NAMESPACE", "kserve")

controllers/reconcilers/kserve_networkpolicy_reconciler.go
zdtsw: if i recall it correctly, from 2.5 release, serving started to use `user-workload-monitoring` namespace for shipping their metrics to. pre-2.5 it was the standard `openshift-monitoring` namespace. 
also I remember there is a ticket on serving team to "cleanup the old monitoring stack" which should remove resource created in old releases, including this "allow-from-openshift-monitoring-ns" NWP  ( the one created by modelmesh not by knatvie)

controllers/reconcilers/kserve_networkpolicy_reconciler.go
vaibhavjainwiz: this env always initialize
```
          env:
            - name: POD_NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
                  ```

File: controllers/reconcilers/kserve_networkpolicy_reconciler.go (hunk 4 above)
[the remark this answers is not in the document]
Comment (vaibhavjainwiz) [reply controllers/reconcilers/kserve_networkpolicy_reconciler.go:121]: If promote https://github.com/opendatahub-io/odh-model-controller/pull/136 to handling deletion of NP.

File: controllers/reconcilers/kserve_networkpolicy_reconciler.go (hunk 9 above)
Comment (skonto) [reply controllers/reconcilers/kserve_networkpolicy_reconciler.go:34]: > if i recall it correctly, from 2.5 release, serving started to use user-workload-monitoring namespace for shipping their metrics to

Not really we haven't changed anything. 

> also I remember there is a ticket on serving team to "cleanup the old monitoring stack" which should remove resource created in old releases, including this "allow-from-openshift-monitoring-ns" NWP ( the one created by modelmesh not by knatvie)

Have a pointer?

File: controllers/reconcilers/kserve_networkpolicy_reconciler.go (hunk 4 above)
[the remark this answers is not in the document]
Comment (bartoszmajsak) [reply controllers/reconcilers/kserve_networkpolicy_reconciler.go:121]: I think we differ on how the cleanup should be handled, so I hope we can work out a consensus with other reviewers :)

Thinking of it more, the logic should be implemented here, as it's this controller that owns it. I still vouch however to **not** keep the policy names in the code but leverage the concept of a label to handle such cases in a more generic fashion. I also wonder if we do need this as part of the reconciliation loop instead, maybe a one-off call during the upgrade is enough.

The name of the label is something to discuss further, as we all know it is one of the hardest problems in software engineering :)

File: controllers/reconcilers/kserve_networkpolicy_reconciler.go [hunk 13]
```diff
@@ -98,47 +104,128 @@ func (r *KserveNetworkPolicyReconciler) createDesiredResource(isvc *kservev1beta
...
+		},
+	}
 }
 
-func (r *KserveNetworkPolicyReconciler) processDelta(ctx context.Context, log logr.Logger, desiredPod *v1.NetworkPolicy, existingPod *v1.NetworkPolicy) (err error) {
+// allowTrafficFromApplicationNamespaces creates combined network policy applied to pods in InferenceService namespace.
+// This set of policies allow traffic from:
+// - application namespace, where OpenDataHub and component services are deployed
+// - namespaces created by OpenDataHub where components live
+// - traffic from other DataScienceProjects (namespaces created through dashboard)
```
bdattoma, line 145: User can create workloads in other OCP projects which were not originally created from Dashboard.  it's a recent feature.

I can't tell how much it is used, but still, it's a scenario. Would this be a problem? 

image

bartoszmajsak: @bdattoma Then it boils down to the question if it is possible to know that namespace X from where the traffic is originating is part of the Open Data Hub. Is there any label that would help us answer this question? If not, then it would be blocked even with 2.5 and KServe only, due to e.g. already existing `allow-openshift-monitoring-ns` policy.

controllers/suite_test.go
```diff
@@ -199,3 +213,49 @@ func convertToStructuredResource(path string, out interface{}) (err error) {
...
 	return nil
 }
+
+var letters = []rune("abcdefghijklmnopqrstuvwxyz")
+
+func appendRandomNameTo(prefix string) string {
+	return concatToMax(63, prefix, generateString(16))
+}
+
+// generateString generates random alphabetical name which can be used for example as application or namespace name.
```
skonto, line 223: In some of our Knative tests we use: 
```
export SYSTEM_NAMESPACE="${SYSTEM_NAMESPACE:-$(uuidgen | tr 'A-Z' 'a-z')}"
```
However this is not a script another idea would be to use something like what K8s has (https://pkg.go.dev/k8s.io/apiserver/pkg/storage/names):
```
func (simpleNameGenerator) GenerateName(base string) string {
	if len(base) > MaxGeneratedNameLength {
		base = base[:MaxGeneratedNameLength]
	}
	return fmt.Sprintf("%s%s", base, utilrand.String(randomLength))
}
```
See https://pkg.go.dev/k8s.io/apimachinery/pkg/util/rand of how that is used with some prefix as the base.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
bartoszmajsak: Can anyone shed some light here? @danielezonca @zdtsw

bdattoma: I think, but should be confirmed, that pods created via dashboard have the `"opendatahub.io/dashboard": "true"` label. I've checked with a workbench created from UI but running into a OCP project (not created via RHOAI Dashboard).

So, could we use a such pod selector in the NP?

```
podSelector:
        matchLabels:
          "opendatahub.io/dashboard": "true"
```

bdattoma: hm I see now that Modelmesh model pods, for example, don't have the same label ^. It has something else like `modelmesh-service=modelmesh-serving`

controllers/reconcilers/kserve_networkpolicy_reconciler.go
zdtsw: @danielezonca you remember any jira was created to do the cleanup work from old monitoring stack?  i can only find https://issues.redhat.com/browse/RHOAIENG-293 but from the last linked PR, does not like it has touched NWP

controllers/reconcilers/kserve_networkpolicy_reconciler.go
zdtsw: i think the best thing is to filter on which namespace, and have the open for all pods running in these namespaces first, then we can set more strict match on the pod level or even ports etc. 

the question in this comment is, if user can select "All available projects" from the UI dropdown, then it is much more namespace can be used as a datascience project(?) (in my dev cluster, i can even select istio-system and knative-eventing,....)
P.S:  i did not know this is available till now :woman_shrugging:

controllers/reconcilers/kserve_networkpolicy_reconciler.go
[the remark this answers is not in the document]
skonto: I guess there is no isolation between different notebooks also I guess a project admin could just misuse this right?

controllers/reconcilers/kserve_networkpolicy_reconciler.go
bdattoma: > then it is much more namespace can be used as a datascience project(?)


correct.

If having a generic solution here is complicated and not wroth the effort, we could probably make it working for the proper DS Projects (i.e., from UI), and release note a workaround in the case the user is using a generic OCP project. I don't link much as approach, but:
- I'm unsure about likelihood to face the scenario (new feature, I think landed in 2.4 or 2.5). I didn't follow it, so don't know the requirement behind it
- This is supposed to be a short-term fix in our code right? if we are confident a long term solution can arrive in the next or 2 releases, we might be ok taking this risk

Probably it's worth to include PMs as well in the decision.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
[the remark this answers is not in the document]
danielezonca: > I am not sure if the scenario where user creates a namespace on their own is something we should support / cover here. @danielezonca could you confirm?

I think it will be possible in the future but the focus now is only "DS project" (aka namespaces with proper annotation)

controllers/reconcilers/kserve_networkpolicy_reconciler.go
danielezonca: I think @VedantMahabaleshwarkar is still working on that.
@VedantMahabaleshwarkar Please have a look here to make sure we align the changes

controllers/reconcilers/kserve_networkpolicy_reconciler.go
danielezonca: This PR might be revisit in the future if we can remove the NP entirely so let's focus on "DS project" for now.
For "advanced" use cases where the user creates a namespace explicitly we can document the necessary NP

bdattoma: sounds good, do we already know what would be the NP in case of "advanced" use case? So we can provide it in the docs.

controllers/suite_test.go
bartoszmajsak: Thanks @skonto! I shamelessly copied the code in 75c6e00, as pulling in the whole apiserver dependency sounds a bit like overkill.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
danielezonca: Asked Adam to confirm https://issues.redhat.com/browse/RHOAIENG-1955?focusedId=23833955&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-23833955 but I would keep "DS project" filter in this PR

controllers/reconcilers/kserve_networkpolicy_reconciler.go
Jooho: Right. He is still working on this.
This is [@VedantMahabaleshwarkar PR](https://github.com/opendatahub-io/odh-model-controller/pull/146). @vaibhavjainwiz and I think using init-container to remove old monitoring stack is not a good idea. So 
 @VedantMahabaleshwarkar and @vaibhavjainwiz are discussing how to implement to remove old monitoring stack.

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -87,11 +97,62 @@ var _ = Describe("The Openshift Kserve model controller", func() {
...
+				if err != nil {
+					Fail(err.Error())
+				}
+				return networkPolicies.Items
+			}, timeout, interval).Should(
+				ContainElements(
+					withMatchingNestedField("ObjectMeta.Name", Equal("allow-from-openshift-monitoring-ns")),
+					withMatchingNestedField("ObjectMeta.Name", Equal("allow-openshift-ingress")),
+					withMatchingNestedField("ObjectMeta.Name", Equal("allow-from-opendatahub-ns")),
+				),
```
terrytangyuan, line 126: Are there other fields we need to check? Perhaps do a deep-equal check?

controllers/kserve_inferenceservice_controller_test.go
bartoszmajsak: Separate PR #147 

Reverted here in ad03002

controllers/reconcilers/kserve_networkpolicy_reconciler.go
terrytangyuan: Yes, a separate PR would be good for build/chores

controllers/reconcilers/kserve_networkpolicy_reconciler.go
```diff
@@ -98,47 +104,128 @@ func (r *KserveNetworkPolicyReconciler) createDesiredResource(isvc *kservev1beta
...
+func (r *KserveNetworkPolicyReconciler) allowOpenshiftIngressPolicy(isvc *kservev1beta1.InferenceService) *v1.NetworkPolicy {
+	return &v1.NetworkPolicy{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "NetworkPolicy",
+			APIVersion: "networking.k8s.io/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      openshiftIngressNetworkPolicyName,
+			Namespace: isvc.Namespace,
+			Labels:    map[string]string{"opendatahub.io/related-to": "RHOAIENG-1003"},
```
terrytangyuan, line 121: Can you add some (code) comments on this label?

controllers/kserve_inferenceservice_controller_test.go
bartoszmajsak: Not sure how deeply we want to go with contracts here, I just wanted to make sure that we have created those policies. Based on my knowledge we are anyway not able to check in evntest if they actually work on a network level so my thinking was we can actually stop here. WDYT?

controllers/reconcilers/kserve_networkpolicy_reconciler.go
VedantMahabaleshwarkar: @Jooho @zdtsw @danielezonca My PR for https://issues.redhat.com/browse/RHOAIENG-293 is to remove the lingering resources related to the monitoring stack. i. e the Prometheus operator deployments and operator pods. It does not touch Networkpolicies as the old monitoring stack itself did not create any Network Policies

controllers/kserve_inferenceservice_controller_test.go
israel-hdez: FWIW,

IMHO, checking only for presence is just fine in this case, because it is almost a static NP. The only thing that varies is the namespace. If the NP would be different based on properties of something (for example, based on properties of the `InferenceService`), that would be worth additional assertions. But for these instances, I'm more than happy with just checking presence.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
VedantMahabaleshwarkar: The `allow-from-openshift-monitoring-ns` NetworkPolicy was added for Kserve during the work to enable Kserve metrics in UWM so I have not touched that in my PR

controllers/reconcilers/kserve_networkpolicy_reconciler.go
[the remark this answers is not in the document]
VedantMahabaleshwarkar: I agree with @israel-hdez about not having cleanup login inside a reconcile loop, especially because IMO cleanup logic should only run once. 
For example, to clean up resources related to the deprecated monitoring stack, I added the logic to a temporary initContainer for the odh-model-controller deployment which I felt was a better way of deleting those resources. This can differ for this PR though, as the number of NetworkPolicies and what namespaces they reside in is not known at startup time - the cleanup logic would have to determine that. 
For my PR the list of resources to be deleted is already known

controllers/reconcilers/kserve_networkpolicy_reconciler.go
israel-hdez: Commenting about the original suggested change, I don't think it is worth to do the renaming of the NP. In existing releases, the NP is already created with `allow-from-openshift-monitoring-ns` name. Since we want to fully move away from creating NPs, spending effort on a rename (i.e. migrating from `allow-from-openshift-monitoring-ns` -> to `allow-from-uwm-ns`) is probably not worth the time.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
israel-hdez: IMHO the most important use-case is the one from @cfchase. If I understood it correctly: assuming a namespace belongs to a tenant, services within the namespace should be reachable between them transparently, regardless if they belong or not to the mesh.

Also, we should have NPs to let infrastructure components to do their job. This would mean to allow pods in the `AppNamespace` to access pods in the namespaces where a KServe `InferenceService` exists (regardless if it is a DSProject or not). I think this PR already covers this, right?.

To me, traffic flowing to/from other DSProject is just a "plus" here. Assuming that the discussion here is about cross-namespace traffic between DSProjects and non-DSProjects, I'd agree that for the time being documenting it as a known issue with the workaround would be OK. However, this PR should still properly enable traffic between services in the same namespace where an `InferenceService` exists, even if the namespace in play is not a DSProject.

controllers/reconcilers/kserve_networkpolicy_reconciler.go
israel-hdez: I'm no longer sure about the intended goal. Now, on the code another dev seein `constants.KServeNamespace` would think that it would return KServe controller namespace, while it is odh-model-controller namespace.

When installing via the operator, it is the same. But for development, it is not. Not sure if this will be confusing.
At least, drop a comment that such value would refer to odh-model-controller namespace, rather than KServe controller namespace.

israel-hdez: As an aside: we are due to update the KServe version in go.mod.

controllers/kserve_inferenceservice_controller_test.go
terrytangyuan: No strong preference

VedantMahabaleshwarkar:
@bartoszmajsak I just merged https://github.com/opendatahub-io/odh-model-controller/pull/148 which fixed an issue on the main branch. Please make sure to rebase this PR to include the changes before merging

PR #146, opened 2024-01-19
add temp Job to delete old monitoring stack

VedantMahabaleshwarkar:
## Fixes: 
https://issues.redhat.com/browse/RHOAIENG-293 

## Context: 

The `model-monitoring` stack was deprecated in RHODS 2.5. This means that starting from 2.5, the RHODS Operator does not create the resources for `model-monitoring` during new installs. However, for upgrades from 2.4, the existing resources are not deleted. This PR adds a Job for RHODS 2.6 that deletes the resources from the `model-monitoring` stack that are no longer needed. 
Resources being deleted : 
- Deployment `rhods-prometheus-operator` 
- StatefulSet (containing 3 pods) `prometheus-rhods-model-monitoring` 
- ServiceMonitor `modelmesh-federated-metrics`

Note: This Job is a temporary addition for RHODS 2.6 ONLY, and will be reverted in 2.7 as it will not be needed anymore. 

## PR changes : 
- Added initContainer to odh-model-controller deployment to delete the resources mentioned above
- Added permissions to odh-model-controller-role to allow it the necessary privileges to perform the deletion
 
#### Prerequisites 
- Install RHODS 2.4
- We install 2.4 because we want to test if the initContainer correctly deletes the resources created by the old monitoring stack. 
- I used RC3 build for 2.4 because I could not find a way to install the 2.4 version of the product from operatorhub. 
```
Catalog Source Images for 2.4 RC3 
Index image v4.11: registry-proxy.engineering.redhat.com/rh-osbs/iib:625499
Index image v4.12: registry-proxy.engineering.redhat.com/rh-osbs/iib:625502
Index image v4.13: registry-proxy.engineering.redhat.com/rh-osbs/iib:625517
Index image v4.14: registry-proxy.engineering.redhat.com/rh-osbs/iib:625527
Index image v4.15: registry-proxy.engineering.redhat.com/rh-osbs/iib:625532
``` 
[Instructions to install RC builds on clusters (doesn't work on ROSA):](https://docs.google.com/document/d/1DcnCV500nMrJFknWJXC6XAaylKlUY6XB4swltwiImKQ/edit#heading=h.uhu58j7scukc)   

#### Testing Steps
- Spin down RHODS Operator deployment to 0 in NS `redhat-ods-operator` 
   - We do this because we are testing the deletion of resources that are no longer watched by the operator starting from 2.5.0
   - For more context, the operator PR is [here](https://github.com/red-hat-data-services/rhods-operator/pull/151)
- Modify role `odh-model-controller-role` in NS `redhat-ods-applications` as per the rbac changes in PR 
   - Add `get`, `list`, `delete` permissions in apiGroup `apps` for resources `deployments`, `statefulsets` 
   - Add `get` permissions for apiGroup `project.openshift.io` for resources `projects` 
- Manually create a Job from the `remove-deprecated-monitoring.yaml` file in the PR
- Manually verify the following resources no longer exist in the cluster in the NS `redhat-ods-monitoring`
   - Deployment `rhods-prometheus-operator`
   - Statefulset `prometheus-rhods-model-monitoring`
   - ServiceMonitor `modelmesh-federated-metrics` 
 
- Success! :) 

### DIFF
--- a/config/overlays/odh/kustomization.yaml
+++ b/config/overlays/odh/kustomization.yaml
@@ -2,6 +2,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - ../../default
+  - remove-deprecated-monitoring.yaml
 
 patchesStrategicMerge:
   - odh_model_controller_manager_patch.yaml
--- a/config/overlays/odh/remove-deprecated-monitoring.yaml
+++ b/config/overlays/odh/remove-deprecated-monitoring.yaml
@@ -1,0 +1,71 @@
+
+kind: Job
+apiVersion: batch/v1
+metadata:
+  name: remove-deprecated-monitoring
+  namespace: redhat-ods-applications
+spec:
+  template:
+    spec:
+      restartPolicy: OnFailure
+      serviceAccountName: odh-model-controller
+      terminationGracePeriodSeconds: 30
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 999
+      containers:
+        - name: oc-cli
+          image: >-
+            registry.redhat.io/openshift4/ose-cli@sha256:25fef269ac6e7491cb8340119a9b473acbeb53bc6970ad029fdaae59c3d0ca61
+          command: ["/bin/bash", "-c", "--"]
+          args:
+          - |
+            #check if namespace redhat-ods-monitoring exists 
+            monitoringnamespaceexists=$(oc get project redhat-ods-monitoring | grep redhat-ods-monitoring || echo "false")
+            if [[ $monitoringnamespaceexists != "false" ]]; then
+
+              # Initialize an array to store errors
+              errors=()
+
+              #delete prometheus operator deployment
+              prometheusoperatordeploymentexists=$(oc get deployments -o name -n redhat-ods-monitoring | grep rhods-prometheus-operator || echo "false")
+              if [[ $prometheusoperatordeploymentexists != "false" ]]; then
+                echo "deleting prometheus operator deployment for deprecated model monitoring stack"
+                oc delete deployment rhods-prometheus-operator -n redhat-ods-monitoring || errors+=("Failed to delete prometheus operator deployment")
+              fi 
+
+              #delete prometheus statefulset
+              prometheusstatefulsetexists=$(oc get statefulsets -o name -n redhat-ods-monitoring | grep prometheus-rhods-model-monitoring || echo "false")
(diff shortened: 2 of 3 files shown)


israel-hdez (dismissed):
If I understand correctly opendatahub-io/opendatahub-operator#703, the ownership of the no longer existent components belong to the `opendatahub-operator`. So, I suggest to move this logic (either in same way or in another equivalent form) to the `opendatahub-operator`.

The `odh-model-controller` should not be responsible for fixing install, nor upgrade issues.

VedantMahabaleshwarkar:
@israel-hdez 
>If I understand correctly https://github.com/opendatahub-io/opendatahub-operator/pull/703, the ownership of the no longer existent components belong to the opendatahub-operator. So, I suggest to move this logic (either in same way or in another equivalent form) to the opendatahub-operator.
The odh-model-controller should not be responsible for fixing install, nor upgrade issues. 

I'd say that since the model monitoring stack was a component specific to model serving, model serving related repos should own the logic to also delete it during upgrade. 
https://github.com/opendatahub-io/opendatahub-operator/pull/703 takes care of no longer deploying those components as that logic has existed since we had "odh-manifests". However, model serving is the logical owner of the deprecated monitoring stack and hence should take care of deleting it.

israel-hdez:
> However, model serving is the logical owner of the deprecated monitoring stack and hence should take care of deleting it.

Looking at this as _working groups_, I agree with this statement. However, I'm more on the thinking that the component that created the resources is the one that should do the cleanup. Model serving WG is still the owner, but the implementation should be on `opendatahub-operator`.

Furthermore, we should guarantee clean-up even if the serving stack is fully disabled in any given cluster. In such case, if resources went stale for any reason, moving/running this code at the operator is what makes the most sense.

Well, I'm going to be strict in that fixing installation issues is not a concern of this repository/component. I'll tag other people to review and if they agree that this is the good repo to host the fix, I'll switch my review to an approval: @terrytangyuan @Jooho @danielezonca .

VedantMahabaleshwarkar:
>Well, I'm going to be strict in that fixing installation issues is not a concern of this repository/component. I'll tag other people to review and if they agree that this is the good repo to host the fix, I'll switch my review to an approval: @terrytangyuan @Jooho @danielezonca .

Discussed offline with @Jooho @terrytangyuan and @israel-hdez , for now we're ok with having this PR in this repo

heyselbi:
@israel-hdez @VedantMahabaleshwarkar @Jooho I manually tested on OSD cluster v4.12 with RC build. And all tests passed

PR #158, opened 2024-01-25
Remove temporary job to delete old monitoring stack, and fix kubebuilder RBAC

israel-hdez:
## Description
This does two things:
* Reverts #146, since we no longer need the temporary job in `main` branch.
* Fixes broken `make manifests`.
   * The `make manifests` is broken, because some privileges were manually added to `config/rbac/role.yaml` which should be an auto-generated file.
   * To fix, this PR adds the missing `kubebuilder:rbac` comments in `main.go`. Then `make manifests` is ran to properly re-generate the `role.yaml` file.
   * The version of `controller-gen` was bumped, to be able to include a header file making clear that the `role.yaml` file should not be changed manually.

## How Has This Been Tested?
The revert of #146 removed a `Job` (manifest change), which can be seen clearly in the PR. It also changed the `role.yaml` file (another manifest change). Since the revert just involves changes to manifests, only build correctness was verified by running: ´kustomize build config/base/´ and verifying no errors.

The fix for `make manifests` involved regenerating the `role.yaml` file. To verify that privileges of the role remain the same:

* Apply the `role.yaml` from `main` branch: `oc apply -f https://raw.githubusercontent.com/opendatahub-io/odh-model-controller/main/config/rbac/role.yaml`
* Get role description: `oc describe clusterrole odh-model-controller-role > role-before.txt`
* Apply the `role.yaml` from this PR: `oc apply -f config/rbac/role.yaml`. If should apply correctly, which would indicate a valid YAML.
* Get new role description: `oc describe clusterrole odh-model-controller-role > role-after.txt`
* Compare the two descriptions to check that they are equivalent: `gvimdiff role-before role-after.txt`. You will see something like this (_before_ at left, _after_ at right):

image

You will notice that 3 lines at the bottom are being removed from the role. That's OK because those privileges were needed only by the temporary `Job` that does cleanup of the old monitoring stack. Since the `Job` is being removed (the revert) those privileges are no longer needed.

You will also notice that the privileges for the `ServiceAccounts` is simply relocated with the same re-ordered privileges.

## Merge criteria:

- [X] The commits are squashed in a cohesive manner and have meaningful messages.
  - NOTE: It is not squashed: 1 commit for the revert, and 1 to fix `make manifests`.
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [X] The developer has manually tested the changes and verified that the changes work


VedantMahabaleshwarkar (approved):
/lgtm

PR #164, opened 2024-02-16
support no reconciliation for rawdeployments

VedantMahabaleshwarkar:
## Description

Fixes : https://issues.redhat.com/browse/RHOAIENG-3243 
This PR disables the current reconciliation process for Kserve InferenceServices if the deployment mode is `RawDeployment`. 

#### Testing Prerequisites 
- Install ODH Operator
- Create the following DSC 
```
spec:
  components:
    codeflare:
      managementState: Removed
    kserve:
      managementState: Managed
      serving:
        ingressGateway:
          certificate:
            type: SelfSigned
        managementState: Managed
        name: knative-serving
    modelregistry:
      managementState: Removed
    trustyai:
      managementState: Removed
    ray:
      managementState: Removed
    kueue:
      managementState: Removed
    workbenches:
      managementState: Removed
    dashboard:
      managementState: Managed
    modelmeshserving:
      devFlags:
        manifests:
          - contextDir: config
            sourcePath: overlays/odh
            uri: 'https://github.com/opendatahub-io/modelmesh-serving/tarball/main'
          - contextDir: config
            sourcePath: ''
            uri: >-
              https://github.com/VedantMahabaleshwarkar/odh-model-controller/tarball/rawdeployments_test
      managementState: Managed
    datasciencepipelines:
      managementState: Removed
```

#### Scenario 1 : RawDeployment mode is specified in the ISVC 
- Create an ISVC for Kserve and explicitly use RawDeployment mode in the ISVC 
   - This can be done by adding the following annotation to the ISVC `serving.kserve.io/deploymentMode: RawDeployment` 
- Check ODH model controller logs for the following 
```
INFO controllers.InferenceService Reconciling InferenceService for Kserve in mode RawDeployment {"InferenceService": "ISVC_NAME", "namespace": "ISVC_NAMESPACE"}
DEBUG controllers.InferenceService No Reconciliation to be done for inferenceservice as it is using RawDeployment mode {"InferenceService": "ISVC_NAME", "namespace": "ISVC_NAMESPACE", "inferenceservice": "ISVC_NAME"}
```

#### Scenario 2 : ISVC does not explicitly set deployment mode but Kserve default mode is RawDeployment 
- Scale ODH Operator down to 0 (to prevent reconciliation of configmap `inferenceservice-config`) 
- Edit configmap `inferenceservice` in NS `opendatahub` as follows : 
```
  deploy: |-
    {
      "defaultDeploymentMode": "RawDeployment"
    }
``` 
- Create the same ISVC as scenario 1 but without the annotation for RawDeployment 
- Verify same ODH Model Controller logs as scenario 1

## Merge criteria:

### DIFF
--- a/controllers/reconcilers/kserve_inferenceservice_reconciler.go
+++ b/controllers/reconcilers/kserve_inferenceservice_reconciler.go
@@ -148,7 +148,7 @@ func (r *KserveInferenceServiceReconciler) DeleteKserveMetricsResourcesIfNoKserv
 		if err != nil {
 			return err
 		}
-		if isvcDeploymentMode == utils.Serverless {
+		if isvcDeploymentMode != utils.Serverless {
 			inferenceServiceList.Items = append(inferenceServiceList.Items[:i], inferenceServiceList.Items[i+1:]...)
 		}
 	}
--- a/controllers/reconcilers/kserve_inferenceservice_reconciler.go
+++ b/controllers/reconcilers/kserve_inferenceservice_reconciler.go
@@ -57,8 +57,12 @@ func NewKServeInferenceServiceReconciler(client client.Client, scheme *runtime.S
 	}
 }
 
-func (r *KserveInferenceServiceReconciler) Reconcile(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+func (r *KserveInferenceServiceReconciler) ReconcileRawDeployment(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+	log.V(1).Info("No Reconciliation to be done for inferenceservice as it is using RawDeployment mode")
+	return nil
+}
 
+func (r *KserveInferenceServiceReconciler) ReconcileServerless(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
 	//  Resource created per namespace
 	log.V(1).Info("Verifying that the default ServiceMeshMemberRoll has the target namespace")
 	if err := r.istioSMMRReconciler.Reconcile(ctx, log, isvc); err != nil {
@@ -140,7 +144,11 @@ func (r *KserveInferenceServiceReconciler) DeleteKserveMetricsResourcesIfNoKserv
 
 	for i := len(inferenceServiceList.Items) - 1; i >= 0; i-- {
 		inferenceService := inferenceServiceList.Items[i]
-		if utils.IsDeploymentModeForIsvcModelMesh(&inferenceService) {
+		isvcDeploymentMode, err := utils.GetDeploymentModeForIsvc(ctx, r.client, &inferenceService)
+		if err != nil {
+			return err
+		}
+		if isvcDeploymentMode == utils.Serverless {
 			inferenceServiceList.Items = append(inferenceServiceList.Items[:i], inferenceServiceList.Items[i+1:]...)
 		}
 	}
--- a/controllers/reconcilers/mm_inferenceservice_reconciler.go
+++ b/controllers/reconcilers/mm_inferenceservice_reconciler.go
@@ -17,6 +17,7 @@ package reconcilers
 
 import (
 	"context"
+
 	"github.com/go-logr/logr"
 	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
 	"github.com/opendatahub-io/odh-model-controller/controllers/utils"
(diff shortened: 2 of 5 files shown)


controllers/utils/utils.go
```diff
@@ -1,21 +1,81 @@
...
+			return Serverless, nil
+		} else {
+			return RawDeployment, nil
+		}
 	}
-	return false
+	return "", nil
+}
+
+func GetOperatorNamespace() (string, error) {
```
aslakknutsen, line 70: Isn't that the same as env.POD_NAMESPACE ?

spolti: indeed, every pod have the NAMESPACE env set on it.
Vendat, is there a specific reason to read it form pod's file system?

VedantMahabaleshwarkar: refactored to use os.getenv("POD_NAMESPACE")

spolti:
@israel-hdez can you please take a look?

controllers/reconcilers/kserve_inferenceservice_reconciler.go
```diff
@@ -140,7 +144,11 @@ func (r *KserveInferenceServiceReconciler) DeleteKserveMetricsResourcesIfNoKserv
 
 	for i := len(inferenceServiceList.Items) - 1; i >= 0; i-- {
 		inferenceService := inferenceServiceList.Items[i]
-		if utils.IsDeploymentModeForIsvcModelMesh(&inferenceService) {
+		isvcDeploymentMode, err := utils.GetDeploymentModeForIsvc(ctx, r.client, &inferenceService)
+		if err != nil {
+			return err
+		}
+		if isvcDeploymentMode == utils.Serverless {
```
israel-hdez, line 151: Isn't the logic inverted here?
```suggestion
		if isvcDeploymentMode != utils.Serverless {
```

controllers/reconcilers/mm_inferenceservice_reconciler.go
```diff
@@ -67,7 +68,11 @@ func (r *ModelMeshInferenceServiceReconciler) DeleteModelMeshResourcesIfNoMMIsvc
 
 	for i := len(inferenceServiceList.Items) - 1; i >= 0; i-- {
 		inferenceService := inferenceServiceList.Items[i]
-		if !utils.IsDeploymentModeForIsvcModelMesh(&inferenceService) {
+		isvcDeploymentMode, err := utils.GetDeploymentModeForIsvc(ctx, r.client, &inferenceService)
+		if err != nil {
+			return err
+		}
+		if isvcDeploymentMode == utils.ModelMesh {
```
israel-hdez, line 75: Also here, the logic looks inverted.
```suggestion
		if isvcDeploymentMode != utils.ModelMesh {
```

controllers/utils/utils.go
```diff
@@ -1,22 +1,66 @@
...
-	if exists && value == inferenceServiceDeploymentModeAnnotationValue {
-		return true
+	if exists {
+		switch value {
+		case string(ModelMesh):
+			return ModelMesh, nil
+		case string(Serverless):
+			return Serverless, nil
+		case string(RawDeployment):
+			return RawDeployment, nil
```
israel-hdez, line 39: Being this an uncontrolled annotation, I would recommend to add a default:
```suggestion
		case string(RawDeployment):
			return RawDeployment, nil
		default:
			return "", fmt.Errorf("the deployment mode '%s' of the Inference Service is invalid", value)
```

controllers/utils/utils.go
```diff
@@ -1,22 +1,66 @@
...
+		}
+		defaultDeploymentMode := deployData["defaultDeploymentMode"]
+		if defaultDeploymentMode == string(Serverless) {
+			return Serverless, nil
+		} else {
+			return RawDeployment, nil
+		}
 	}
-	return false
+	return "", nil
```
israel-hdez, line 63: Once you add a default to the switch up there, you can remove this line.

The callers of t his new function does not seem to take into account that an empty string would be returned. The behaviors seems to be undetermined in such case. So, let's not return an unexpected value.

controllers/utils/utils.go
```diff
@@ -1,22 +1,66 @@
...
+		}, inferenceServiceConfigMap)
+		if err != nil {
+			return "", fmt.Errorf("error getting configmap 'inferenceservice-config'. %w", err)
+		}
+		var deployData map[string]interface{}
+		if err = json.Unmarshal([]byte(inferenceServiceConfigMap.Data["deploy"]), &deployData); err != nil {
+			return "", fmt.Errorf("error retrieving value for key 'deploy' from configmap %s. %w", KserveConfigMapName, err)
+		}
+		defaultDeploymentMode := deployData["defaultDeploymentMode"]
+		if defaultDeploymentMode == string(Serverless) {
```
israel-hdez, line 57: ModelMesh can also be a default deployment mode. See https://github.com/kserve/kserve/blob/3b15ef554aa829d0f36fec3f95a543afc4578872/config/configmap/inferenceservice.yaml#L428-L434. Let's cover that, rather than defaulting to Raw.

In fact, the defaultDeploymentMode can only have those 3 valid values (Serverless, Raw, MM). KServe code would complain if it is not set or has some other value ([ref here](https://github.com/kserve/kserve/blob/3b15ef554aa829d0f36fec3f95a543afc4578872/pkg/apis/serving/v1beta1/configmap.go#L172)). We should do the same and return an error if there is some unexpected value, rather than using a default value.

VedantMahabaleshwarkar: added a switch for `defaultDeploymentMode` to handle all 3 possible options and error on any other value

israel-hdez (approved):
Looks good!

The logic looks good to me, under the assumption that we don't support users flipping the configs; i.e. changing the DeploymentMode annotation of an ISVC would be unsupported and _also_ changing the defaultDeploymentMode of an existent ODH setup would be unsupported.

Please, note that I haven't tried the code. Approving to expedite merging.

israel-hdez:
@VedantMahabaleshwarkar Please, cherry-pick to `release-0.11.1` branch.

PR #166, opened 2024-02-22
[Feature] Integrated OpenDataHub Centralized Custom Certificate

Jooho:
## Description

[OpenDataHub Centralized Cert PR](https://github.com/opendatahub-io/opendatahub-operator/pull/848) has been merged.
KServe can now integrate with the centralized certificate as a result of this change.
This PR facilitates the update of the storage-config secret, which includes the custom certificate.
More details about the issue can be found [here](https://issues.redhat.com/browse/RHOAIENG-1010).

## How Has This Been Tested?

**Setup**
*Manual*
1. install ossm/serverless operators
2. create this catalogsource 
~~~
cat << EOF|oc apply -f -
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: rhods-catalog-dev
  namespace: openshift-marketplace
spec:
  displayName: Red Hat OpenShift Data Science
  publisher: RHODS Development Catalog
  image: quay.io/opendatahub/opendatahub-operator-catalog:v2.8.0-incubation
  sourceType: grpc
EOF
~~~
3. install rhods v2.8.0-incubation operator
4. create dsci 
5. update dsci
~~~
kubectl patch DSCInitialization default-dsci --type='json' -p='[{"op": "add", "path": "/spec/trustedCABundle", "value": {"managementState": "Managed", "customCABundle": ""}}]'
~~~

6. create dsc (Managed for dashboard/kserve)
7. deploy ssl minio for test
~~~
# SSL Minio 
export MINIO_DEMO_HOME=/tmp/demo
mkdir $MINIO_DEMO_HOME
cd $MINIO_DEMO_HOME
git clone git@github.com:Jooho/jhouse_openshift.git
cd jhouse_openshift/Minio/minio-tls-kserve/
source env.sh
./1.setup.sh 
./2.generate-cert.sh 

./3.deploy-minio.sh 
cp ${DEMO_HOME}/minio_certs/root.crt   ${DEMO_HOME}/minio_certs/cabundle.crt
cp ${DEMO_HOME}/minio_certs/cabundle.crt $MINIO_DEMO_HOME/.
~~~
8. create a DS namespace 
~~~
oc new-project kserve-demo 
oc label ns/kserve-demo opendatahub.io/dashboard="true"
~~~

9. Deploy Sample Caikit ISVC for KServe ([script](https://github.com/opendatahub-io/caikit-tgis-serving/tree/main/demo/kserve/scripts/test)

*Script*
**(NOTE) This was tested on fedora only.**
~~~
git clone   git@github.com:Jooho/loopy.git
cd loopy
make init
cat cluster_info.sh
CLUSTER_CONSOLE_URL=https://console-openshift-console.XXXX
CLUSTER_API_URL=https://api.XXX:443
CLUSTER_ADMIN_ID=admin
CLUSTER_ADMIN_PW=password
CLUSTER_TOKEN=sha256~XXX
CLUSTER_TYPE=ROSA

./loopy playbooks run setup-rhoai-for-global-cert -i ./cluster_info.sh 
~~~
**Set root.crt path**
~~~
ls /tmp/ms_cli
20240222_XXX

#replace the XXX to yours
export crt_content=/tmp/ms_cli/20240222_XXX/artifacts/8-cert-generate/root.crt
~~~

**TEST**
*1. Create a dataconnection*

*1-1. without global cert*
~~~
#Create a dataconnection
cat <<EOF |oc apply -f -
kind: Secret
apiVersion: v1
metadata:
  name: aws-connection-minio
  labels:
    opendatahub.io/dashboard: 'true'
    opendatahub.io/managed: 'true'
  annotations:
    opendatahub.io/connection-type: s3
    openshift.io/display-name: minio
stringData:
  AWS_ACCESS_KEY_ID: admin
  AWS_DEFAULT_REGION: us-east-1
  AWS_S3_BUCKET: modelmesh-example-models
  AWS_S3_ENDPOINT: https://minio.minio.svc:9000
  AWS_SECRET_ACCESS_KEY: password
EOF

# It is the usual storage-config secret
oc get secret storage-config -o yaml
~~~

*1-2. with global cert*
~~~
# Clear data connection
oc delete secret aws-connection-minio

# Update global cert in dsci
kubectl patch dscinitialization default-dsci --type='json' -p='[{"op":"replace","path":"/spec/trustedCABundle/customCABundle","value":"'"$(awk '{printf "%s\\n", $0}' $crt_content)"'"}]'

# new configmap created for kserve
oc get cm odh-kserve-custom-ca-bundle 

# Create a dataconnection
cat <<EOF |oc apply -f -
kind: Secret
apiVersion: v1
metadata:
  name: aws-connection-minio
  labels:
    opendatahub.io/dashboard: 'true'
    opendatahub.io/managed: 'true'
  annotations:
    opendatahub.io/connection-type: s3
    openshift.io/display-name: minio
stringData:
  AWS_ACCESS_KEY_ID: admin
  AWS_DEFAULT_REGION: us-east-1
  AWS_S3_BUCKET: modelmesh-example-models
  AWS_S3_ENDPOINT: https://minio.minio.svc:9000
  AWS_SECRET_ACCESS_KEY: password
EOF

# storage-config will have certificate information
oc get secret storage-config -o yaml
~~~

**2. Update global cert**
~~~
#update global custom cert
kubectl patch dscinitialization default-dsci --type='json' -p='[{"op":"replace","path":"/spec/trustedCABundle/customCABundle","value":"ABC"}]'
# 2-1. updated kserve cm 
oc get cm odh-kserve-custom-ca-bundle -o yaml
# 2-2. updated storage-config
oc get secret storage-config -o yaml
~~~

**3. KServe test**
~~~
kubectl patch dscinitialization default-dsci --type='json' -p='[{"op":"replace","path":"/spec/trustedCABundle/customCABundle","value":"'"$(awk '{printf "%s\\n", $0}' $crt_content)"'"}]'
oc delete pod --all -n kserve-demo
~~~
**4. ModelMesh test**
~~~
oc label namespace kserve-demo modelmesh-enabled=true --overwrite=true

oc apply -f https://raw.githubusercontent.com/Jooho/jhouse_openshift/main/Kserve/docs/Common/manifests/openvino-serving-runtime.yaml -n kserve-demo
oc patch servingruntime/ovms-1.x -p '{"spec":{"replicas":1}}' --type=merge -n kserve-demo
cat <<EOF| oc apply -f -
kind: InferenceService
metadata:
  name: example-onnx-mnist
  namespace: kserve-demo
  annotations:
    serving.kserve.io/deploymentMode: ModelMesh
spec:
  predictor:
    model:
      modelFormat:
        name: onnx
      runtime: ovms-1.x
      storage:
        key: aws-connection-minio
        path: onnx/mnist.onnx
EOF

# You can see that it successfully pull a model from ssl minio       
oc logs deploy/modelmesh-serving-ovms-1.x -c ovms-adapter
~~~

**Clean up demo**
~~~
oc delete isvc,servingruntime,pod --all -n kserve-demo --force
oc delete ns kserve-demo minio --force --grace-period=0
oc delete all --all --force --grace-period=0 -n opendatahub
oc patch DataScienceCluster rhoai -p '{"metadata":{"finalizers":[]}}' --type=merge
oc delete dsc,dsci --all --force --grace-period=0

oc delete subs serverless-operator -n rhoai-serverless
oc delete subs servicemeshoperator -n openshift-operators 
oc delete subs opendatahub-operator -n openshift-operators

oc delete csv serverless-operator.v1.31.1 servicemeshoperator.v2.4.5 opendatahub-operator.v2.8.0-incubation -n openshift-operators
oc delete catalogsource rhods-catalog-dev -n openshift-marketplace
oc delete pod --all -n rhoai-serverless --force --grace-period=0
oc delete validatingwebhookconfiguration openshift-operators.servicemesh-resources.maistra.io
oc delete mutatingwebhookconfiguration openshift-operators.servicemesh-resources.maistra.io 
oc patch smmr default -p '{"metadata":{"finalizers":[]}}' --type=merge -n istio-system
oc delete ns opendatahub redhat-ods-monitoring rhoai-serverless istio-system 
~~~

## Merge criteria:


controllers/storageconfig_controller_test.go
```diff
@@ -154,7 +200,7 @@ func waitForSecret(cli client.Client, namespace, secretName string, timeout time
 	}
 }
 
-// compareSecrets checks if two Secret data are equal, if not return false
+// // compareSecrets checks if two Secret data are equal, if not return false
```
spolti, line 203:
```suggestion
// compareSecrets checks if two Secret data are equal, if not return false
```

controllers/kserve_customcacert_controller.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+// reconcileConfigMap watch odh global ca cert and it will create/update/delete kserve custom cert configmap
+func (r *KServeCustomCACertReconciler) reconcileConfigMap(configmap *corev1.ConfigMap,
+	ctx context.Context) error {
+	// Initialize logger format
+	log := r.Log
+
+	odhCustomCertData := configmap.Data[constants.ODHCustomCACertFileName]
+	if odhCustomCertData != "" {
```
dtrifiro, line 53: Can this if chain be flattened? This looks pretty convoluted

Jooho: I agree. I made it more clear.

controllers/kserve_customcacert_controller.go
```diff
@@ -0,0 +1,190 @@
...
+	configData := map[string]string{constants.KServeCACertFileName: odhCustomCertData}
+	newCaCertConfigMap := getDesiredCaCertConfigMapForKServe(kserveCustomCACertConfigMapName, configmap.Namespace, configData)
+
+	kserveCustomCertConfigMap := &corev1.ConfigMap{}
+	if err := r.Get(ctx, types.NamespacedName{Name: kserveCustomCACertConfigMapName, Namespace: configmap.Namespace}, kserveCustomCertConfigMap); err != nil {
+		if !apierrs.IsNotFound(err) {
+			return err
+		}
+		log.Info(fmt.Sprintf("Creating KServe custom cert ConfigMap (%s) because it detected the creation of the opendatahub global cert ConfigMap (%s)", kserveCustomCACertConfigMapName, odhGlobalCACertConfigMapName))
+		if err := r.Create(ctx, newCaCertConfigMap); err != nil {
```
vaibhavjainwiz, line 76: If lifecycle of `kserveCustomCACertConfigMap` depends on `odhGlobalCACertConfigMap` then could we define the parent child relationship between them.

controllers/kserve_customcacert_controller.go
```diff
@@ -0,0 +1,190 @@
...
+		WithEventFilter(reconcileOpenDataHubGlobalCACertConfigMap())
+	err := builder.Complete(r)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func getDesiredCaCertConfigMapForKServe(configmapName string, namespace string, caCertData map[string]string) *corev1.ConfigMap {
+	desiredConfigMap := &corev1.ConfigMap{
```
vaibhavjainwiz, line 181: Could we add some annotation here to represent that this configmap is maintain my `odh-model-controller`

controllers/kserve_customcacert_controller.go
```diff
@@ -0,0 +1,190 @@
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *KServeCustomCACertReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// Create a builder that only watch OpenDataHub global certificate ConfigMap
+	builder := ctrl.NewControllerManagedBy(mgr).
```
vaibhavjainwiz, line 170: No event would be listened if any user directly update/delete the `kserveCustomCACertConfigMap`.

controllers/kserve_customcacert_controller.go
```diff
@@ -0,0 +1,190 @@
...
+	err := r.Get(ctx, req.NamespacedName, configmap)
+	if err != nil && apierrs.IsNotFound(err) {
+		log.Info("Opendatahub global cert ConfigMap not found")
+		configmap.Namespace = req.Namespace
+	} else if err != nil {
+		log.Error(err, "Unable to fetch the ConfigMap")
+		return ctrl.Result{}, err
+	}
+
+	err = r.reconcileConfigMap(configmap, ctx)
```
vaibhavjainwiz, line 160: we shoud pass the `log` instace that we created above into `reconcileConfigMap` method. I am asking it because in above line we created a new log instance by appending the ConfigMap name and namespace details. So we pass same reference then every log would print these details by default.

controllers/kserve_customcacert_controller.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// reconcileOpenDataHubGlobalCertConfigMap filters out all ConfigMaps that are not the OpenDataHub global certificate ConfigMap.
+func reconcileOpenDataHubGlobalCACertConfigMap() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			objectName := e.Object.GetName()
+			return checkOpenDataHubGlobalCertCAConfigMapName(objectName)
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
```
vaibhavjainwiz, line 131: We don't need to act on DeleteFunc if we could define the owner reletaionship between `odhGlobalCACertConfigMap` and `kserveCustomCACertConfigMap`

controllers/kserve_customcacert_controller.go
```diff
@@ -0,0 +1,190 @@
...
+			objectName := e.Object.GetName()
+			return checkOpenDataHubGlobalCertCAConfigMapName(objectName)
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
```
vaibhavjainwiz, line 135: Could you please explain on which type of event we are trying to act on here. I had never used GenericFunc before.

controllers/kserve_customcacert_controller.go
Jooho: @dtrifiro could you review it again?

vaibhavjainwiz (commented):
Current reconciliation logic looks fine but could I request you to please write in similiar pattern that we follow in other reconciler like below. It help us to maintain uniformity in code and its more robust.

https://github.com/opendatahub-io/odh-model-controller/blob/ba123e578471dabfd6154cb6a9588342abb2662c/controllers/reconcilers/kserve_istio_peerauthentication_reconciler.go#L15-L14

vaibhavjainwiz:
@Jooho I am not following whole centralized cert story from begining, Could I ask Who is responsible to create `odhGlobalCACertConfigMap`? I mean does some operator create it or user will create it manually?

If its created by some other operator like `odh-operator` then ideally that operator should maintain the `kserveCustomCACertConfigMap` as well.

controllers/kserve_customcacert_controller.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+		log.Info(fmt.Sprintf("Updating KServe custom cert ConfigMap due to changes in the opendatahub global cert ConfigMap (%s)", odhGlobalCACertConfigMapName))
+		if err := r.Update(ctx, newCaCertConfigMap); err != nil {
+			return err
+		}
+		if err := r.deleteStorageSecret(ctx, configmap.Namespace); err != nil {
+			log.Error(err, "Failed to delete the storage-config secret to update the custom cert")
+			return err
+		}
+		log.V(1).Info("Deleted the storage-config Secret to update the custom cert")
```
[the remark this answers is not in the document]
Jooho: I agreed. What do you think if we delete the storage-config secret when it has `"opendatahub.io/managed: "true"` label?  That means the secret is managed by odh-model-controller.

controllers/kserve_customcacert_controller.go
Jooho: I added label `"opendatahub.io/managed": "true"`

controllers/kserve_customcacert_controller.go
Jooho: added a logic for it

controllers/kserve_customcacert_controller.go
Jooho: Source is refactoryed so I will close this comemnt

controllers/kserve_customcacert_controller.go
Jooho: odhGlobalCACertConfigMap will be created by opendtahub operator and odh-model-controller reply on platform operator.

controllers/kserve_customcacert_controller.go
Jooho: updated.

controllers/kserve_customcacert_controller.go
Jooho: You are right. GenericEvent can include various types of events such as creations, deletions, or updates of Kubernetes API objects. However, it already has those events(create/update/delete) so I think we don't need this.
I removed the event.

controllers/kserve_customcacert_controller.go
Jooho: The `kserveCustomCACertConfigMap` is managed by odh-model-controller but `odhGlobalCACertConfigMap` is managed by opendatahub-operator. When `odhGlobalCACertConfigMap` is removed by operatorhub-operator, odh-model-controller should catch the deletion event to delete `kserveCustomCACertConfigMap`

Jooho:
> Current reconciliation logic looks fine but could I request you to please write in similiar pattern that we follow in other reconciler like below. It help us to maintain uniformity in code and its more robust.
> https://github.com/opendatahub-io/odh-model-controller/blob/ba123e578471dabfd6154cb6a9588342abb2662c/controllers/reconcilers/kserve_istio_peerauthentication_reconciler.go#L15-L14

@vaibhavjainwiz  I agree. Therefore, I have accepted your recommendation and made significant changes to the source code accordingly. Please review.

> Could I ask Who is responsible to create odhGlobalCACertConfigMap? I mean does some operator create it or user will create it manually? If its created by some other operator like odh-operator then ideally that operator should maintain the kserveCustomCACertConfigMap as well.

The `odhGlobalCACertConfigMap` is created by the opendatahub operator. Each component owner must use this centralized certificate appropriately for their respective components. This is because each component applies certificates in very different ways. Although kserve and modelmesh both apply certificates, their methods are different from each other. The `kserveCustomCACertConfigMap` is created specifically for utilizing kserve's functionality, therefore, it needs to be managed by us.

controllers/utils/utils.go
```diff
@@ -25,3 +29,18 @@ func IsNil(i any) bool {
 func IsNotNil(i any) bool {
 	return !IsNil(i)
 }
+
+func GetNamespaceName() (string, error) {
```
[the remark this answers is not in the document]
Jooho: Yea. I just left it for future use like a util function but i can remove it

controllers/kserve_customcacert_controller.go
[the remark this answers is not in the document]
vaibhavjainwiz: Does `storage-config` secret created by `odh-model-controller` in any case? 
So far the way I tested the KServe is by manually create the `storage-config` secret in Kserve namespace.

If there are some cases where `storage-config` secret created by `odh-model-controller` then I am agree with you to delete them by checking on `"opendatahub.io/managed: "true"` label.

controllers/utils/utils.go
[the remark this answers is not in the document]
vaibhavjainwiz: We could use `constants.KServeNamespace` to fetch the namespace of controller. This constant initialize on application startup by reading below env 
```
KServeNamespace                  = getEnvOrDefault("POD_NAMESPACE", "kserve")
```
Value of this env is further define in Deployment of controller itself like below:
```
          env:
            - name: POD_NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
 ```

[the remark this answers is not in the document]
Comment (vaibhavjainwiz) [reply controllers/utils/utils.go:33]: I think we don't need `GetNamespaceName()` function.

File: controllers/storageconfig_controller.go [hunk 11]
```diff
@@ -59,8 +60,10 @@ func newStorageSecret(dataConnectionSecretsList *corev1.SecretList) *corev1.Secr
 		dataConnectionElement["default_bucket"] = string(secret.Data["AWS_S3_BUCKET"])
 		dataConnectionElement["bucket"] = string(secret.Data["AWS_S3_BUCKET"])
 		dataConnectionElement["region"] = string(secret.Data["AWS_DEFAULT_REGION"])
-		if secret.Data["AWS_CA_BUNDLE"] != nil {
-			dataConnectionElement["certificate"] = string(secret.Data["AWS_CA_BUNDLE"])
+
+		if odhCustomCertData != "" {
+			dataConnectionElement["certificate"] = odhCustomCertData
+			dataConnectionElement["cabundle_configmap"] = constants.KServeCACertConfigMapName
```
danielezonca, line 66: Is this a breaking change compared to previous behaviour? What if a user already has a secret with `AWS_CA_BUNDLE`?

controllers/storageconfig_controller.go
```diff
@@ -91,11 +94,23 @@ func (r *StorageSecretReconciler) reconcileSecret(secret *corev1.Secret,
 	if err != nil {
 		if apierrs.IsNotFound(err) {
 			log.Info("No data connections found in namespace ", secret.Namespace)
+			return nil
```
danielezonca, line 97: This return changes quite a lot of the existing logic, it seems good to me (aka if there are no data connection it is fine to exit the method) but I find strange that the existing code is working: as far as I see the `get` for `foundStorageSecret` should fail and the flow should enter in the create branch.
Long story short I would expect the existing code to create an empty (?) secrets even if there is no data connection.
Am I missing something?
Btw I'm fine with the change, it is mainly to understand how it is changing the behaviour

controllers/storageconfig_controller.go
```diff
@@ -91,11 +94,23 @@ func (r *StorageSecretReconciler) reconcileSecret(secret *corev1.Secret,
...
 		}
```
danielezonca, line 98: I would add an else condition to log the error (and return?): currently any error different from `not found` is ignored

danielezonca (commented):
Overall the PR looks good to me, just a couple of comments to clarify impact on existing code

controllers/kserve_customcacert_controller.go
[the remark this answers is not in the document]
Jooho: Yes, it is.
`storage-config` can be manually created but in ODH/RHOAI, it will be managed by data connection and `odh-model-controller` creates the secret.

controllers/utils/utils.go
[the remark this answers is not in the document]
Jooho: great!. I removed the code.

controllers/storageconfig_controller.go
Jooho: The existing code was for a self-signed certificate in ModelMesh, but it was not officially supported because the UI was not implemented. Therefore, replacing the code has no impact.

controllers/storageconfig_controller.go
Jooho: >Long story short I would expect the existing code to create an empty (?) secrets even if there is no data connection.
You are right! This is a bug and I found this during the test. Creating an empty `storage-config` does not make sense because a user can create it manually.  

With the return, it will not create an empty `storage-config` when dataconnection does not exist.

controllers/storageconfig_controller.go
Jooho: You are right. I added `else`.

vaibhavjainwiz:
changes looks good to me as well. I had reviewed it by dry run on code.

Jooho:
@danielezonca I found the root cause why the old logic was working. I fixed it with https://github.com/opendatahub-io/odh-model-controller/pull/167 and it will be included.

Jooho:
This pr had to be squeezed and merged. I will do it manually.

controllers/storageconfig_controller.go
danielezonca: Yes with the new return it is correct indeed, I was referring to the old / previous Logic 👍

PR #168, opened 2024-02-23
[Fix] for integration ca cert, role change needed

Jooho:
https://github.com/opendatahub-io/odh-model-controller/pull/166 needs this change.

This PR does not need to be tested because it gives more permission so it would not impact others.


spolti:
hmm, is it possible to ltgm your own pr?

PR #153, opened 2024-01-22
RHOAIENG-1229 : Error in odh-model-controller logs on deletion of Mod…

vaibhavjainwiz:
…elMesh isvc

Issue : [RHOAIENG-1229](https://issues.redhat.com//browse/RHOAIENG-1229)

## Setup Env
Please follow below instruction to setup your env:
https://gist.github.com/vaibhavjainwiz/61045727dfda727f7577a6f97ca46563

## How Has This Been Tested?
- Delete the deployed ModelMesh Inference
- Verify `modelmesh-serving-sa` ServiceAccount should be deleted.

## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

Issue #73, opened 2023-08-24
Switch from patching ServiceMeshMemberRoll to creating namespaced ServiceMeshMember


israel-hdez:
For KServe, the odh-model-controller patches the `ServiceMeshMemberRoll` resource to add/remove namespaces from the Mesh.

However, the recommended way to add/remove namespaces to the mesh is by creating/deleting a namespaced `ServiceMeshMember` resource (without the `Roll` suffix). The odh-model-controller should switch to the recommended method.

israel-hdez:
Moved to: https://issues.redhat.com/browse/RHOAIENG-1043

PR #172, opened 2024-02-27
Error in odh-model-controller logs on deletion of ModelMesh isvc

vaibhavjainwiz:
cherry pick https://github.com/opendatahub-io/odh-model-controller/pull/153

## Merge criteria:

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -7,10 +7,21 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - serviceaccounts
+  verbs:
+  - create
+  - get
+  - list
+  - patch
+  - update
+  - watch
+  - delete
+- apiGroups:
+  - ""
+  resources:
   - endpoints
   - namespaces
   - pods
-  - serviceaccounts
   - services
   verbs:
   - create


config/rbac/role.yaml
```diff
@@ -4,13 +4,24 @@ kind: ClusterRole
 metadata:
   name: odh-model-controller-role
 rules:
+- apiGroups:
```
aslakknutsen, line 7: This file is generated from the  kubebuilder rules in main.go. Any change to the RBAC rules should be applied to main.go. 

https://github.com/opendatahub-io/odh-model-controller/blob/f6fb33a819b8c91c8c08b9612ca8e9ad9282ed84/main.go#L56-L73

aslakknutsen (changes requested):

config/rbac/role.yaml
Jooho: @vaibhavjainwiz I agreed with @aslakknutsen . Line #73 should be updated.

Jooho (commented):
updating manifests directly is not the right way because it will be regenerated by kubebuilder.

israel-hdez:
@vaibhavjainwiz Is this still needed?

config/rbac/role.yaml
spolti: @vaibhavjainwiz do you have bandwidth to do this?

spolti:
closed as stale.

PR #177, opened 2024-03-06
[RHOAIENG-4191] - odh-model-controller should tolerate missing Authorino

spolti:
This commit allows to run ODH without Authorino when KServe is Removed.

## How Has This Been Tested?
- Install the ODH Operator 2.8 through Operator Catalog.
- Create Default DSCI
- Create a DSC with the following content (KServe -> servinng - disabled):
```yaml
kind: DataScienceCluster
apiVersion: datasciencecluster.opendatahub.io/v1
metadata:
  labels:
    app.kubernetes.io/created-by: opendatahub-operator
    app.kubernetes.io/instance: default
    app.kubernetes.io/name: datasciencecluster
    app.kubernetes.io/part-of: opendatahub-operator
  name: default-dsc
spec:
  components:
    codeflare:
      managementState: Removed
    kserve:
      managementState: Managed
      serving:
        ingressGateway:
          certificate:
            type: SelfSigned
        managementState: Removed
        name: knative-serving
    modelregistry:
      managementState: Removed
    trustyai:
      managementState: Removed
    ray:
      managementState: Removed
    kueue:
      managementState: Removed
    workbenches:
      managementState: Removed
    dashboard:
      managementState: Managed
    modelmeshserving:
      devFlags:
        manifests:
          - contextDir: config
            sourcePath: ''
            uri: >-
              https://github.com/spolti/odh-model-controller/tarball/RHOAIENG-4191-test
      managementState: Managed
    datasciencepipelines:
      managementState: Removed
```
- Note that the KServe - serving component is disabled, it should not log any error, instead, this message should be logged:
```
controllers.InferenceService    kserve service mesh component is enabled, Authorino is required.
```
- Next step, update the DCS setting `KServe`/serving to `Managed`, install the Service Mesh Operator and  **don't install the Authorino operator**, and watch the logs for the error:

```
2024-03-06T19:25:57Z	INFO	controllers.InferenceService	kserve component is enabled, Authorino is required
2024-03-06T19:26:33Z	ERROR	controller-runtime.source	if kind is a CRD, it should be installed before calling Start	{"kind": "AuthConfig.authorino.kuadrant.io", "error": "no matches for kind \"AuthConfig\" in version \"authorino.kuadrant.io/v1beta2\""}
sigs.k8s.io/controller-runtime/pkg/source.(*Kind).Start.func1.1
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/source/source.go:143
k8s.io/apimachinery/pkg/util/wait.runConditionWithCrashProtectionWithContext
	/opt/app-root/src/go/pkg/mod/k8s.io/apimachinery@v0.27.0/pkg/util/wait/wait.go:154
k8s.io/apimachinery/pkg/util/wait.waitForWithContext
	/opt/app-root/src/go/pkg/mod/k8s.io/apimachinery@v0.27.0/pkg/util/wait/wait.go:207
k8s.io/apimachinery/pkg/util/wait.poll
	/opt/app-root/src/go/pkg/mod/k8s.io/apimachinery@v0.27.0/pkg/util/wait/poll.go:260
k8s.io/apimachinery/pkg/util/wait.PollImmediateUntilWithContext
	/opt/app-root/src/go/pkg/mod/k8s.io/apimachinery@v0.27.0/pkg/util/wait/poll.go:200
sigs.k8s.io/controller-runtime/pkg/source.(*Kind).Start.func1
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/source/source.go:136

```

## Merge criteria:

### DIFF
--- a/controllers/inferenceservice_controller.go
+++ b/controllers/inferenceservice_controller.go
@@ -152,17 +152,19 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
 			}))
 
 	// check if kserve is enabled, otherwise don't require Authorino.
-	enabled, err := utils.VerifyIfComponentIsEnabled(context.TODO(), r.client, "kserve")
+	enabled, err := utils.VerifyIfComponentIsEnabled(context.TODO(), r.client, utils.KserveAuthorinoComponent)
+	logMessage := "kserve service mesh component is enabled, Authorino is required"
 	if err != nil {
-		r.log.V(1).Error(err, "could not determine if kserve is enabled, default is enabled")
+		r.log.V(1).Error(err, "could not determine if kserve have service mesh enabled")
 		enabled = true
+		logMessage = "could not determine if kserve have service mesh enabled, Authorino integration will be enabled"
 	}
 
 	if enabled {
 		builder.Owns(&authorinov1beta2.AuthConfig{})
-		r.log.Info("kserve component is enabled, Authorino is required")
+		r.log.Info(logMessage)
 	} else {
-		r.log.Info("kserve component is disabled, ignoring Authorino requirement")
+		r.log.Info("kserve serving component is disabled, ignoring Authorino requirement")
 	}
 
 	err = builder.Complete(r)
--- a/controllers/inferenceservice_controller.go
+++ b/controllers/inferenceservice_controller.go
@@ -17,7 +17,6 @@ package controllers
 
 import (
 	"context"
-
 	"github.com/go-logr/logr"
 	kservev1alpha1 "github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
 	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
@@ -122,7 +121,6 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
 		Owns(&networkingv1.NetworkPolicy{}).
 		Owns(&monitoringv1.ServiceMonitor{}).
 		Owns(&monitoringv1.PodMonitor{}).
-		Owns(&authorinov1beta2.AuthConfig{}).
 		Watches(&source.Kind{Type: &kservev1alpha1.ServingRuntime{}},
 			handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
 				r.log.Info("Reconcile event triggered by serving runtime: " + o.GetName())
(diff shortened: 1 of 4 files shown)


controllers/inferenceservice_controller.go
```diff
@@ -152,7 +150,22 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
 				}
 				return reconcileRequests
 			}))
-	err := builder.Complete(r)
+
+	// check if kserve is enabled, otherwise don't require Authorino.
```
VedantMahabaleshwarkar, line 154: will this work without watching for changes to the DSC CR? If the management state of Kserve is changed after the initial install, odh model controller might have an incorrect assumption about the state of kserve

VedantMahabaleshwarkar: also iirc, authorino integration is `opt-in`, right? i.e our components should work when Kserve is enabled but Authorino might not be installed? 
cc @israel-hdez

israel-hdez: Authorino is not opt-in. However, in `RawDeployment` mode (i.e. when KServe is installed without Service Mesh nor Knative), Authorino will be missing.

These changes are good, but does not seem to cover `RawDeployment` mode.

> will this work without watching for changes to the DSC CR?

Being this manager setup, I don't think we can re-setup the controller once manager loop has started. So, changing the managementState of KServe or ServiceMesh/deploymentMode would require a restart of all replicas of odh-model-controller.

spolti: The RawDeployment might be tricky in this case, the setup will happens before any reconcile, and usually users will install the ISVC after creating the DSC.
With that said, users might be requiring to have the isvc pre-defined already.

For RawDeployment, I guess we can do it by adding flags through annotations or config-map configuration to enable or disable authorino integration.

spolti: okays, updated.

spolti: test steps updated as well.

controllers/inferenceservice_controller.go
```diff
@@ -152,7 +150,24 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
...
+	enabled, err := utils.VerifyIfComponentIsEnabled(context.TODO(), r.client, utils.KserveAuthorinoComponent)
+	logMessage := "kserve service mesh component is enabled, Authorino is required"
+	if err != nil {
+		r.log.V(1).Error(err, "could not determine if kserve have service mesh enabled")
+		enabled = true
+		logMessage = "could not determine if kserve have service mesh enabled, Authorino integration will be enabled"
+	}
+
+	if enabled {
```
israel-hdez, line 163: Almost! More feedback:
The DSC can be:
```yaml
apiVersion: datasciencecluster.opendatahub.io/v1
kind: DataScienceCluster
metadata:
  name: default-dsc
spec:
  components:
    kserve: 
      managementState: "Removed",
      serving: 
        ingressGateway: 
          certificate: 
            type: SelfSigned
        name: "knative-serving",
        managementState: "Managed"
```
Notice kserve.managementState is `Removed` while serving.managementState is `Managed`. If KServe is removed, I think odh-model-controller don't need to care about authorino.

So, turns out we would need to check `kserve.managementState == Managed && serving.managementState != Removed`, to require authorino here.

I think I'm bad at explaining stuff.

spolti: both enabled, right.

okay.

israel-hdez (approved):
/lgtm

PR #181, opened 2024-03-13
Prevent creation of KSVC if its namespace is not in the Mesh

israel-hdez:
## Description

When a KServe InferenceService is created, odh-model-controller enrolls the ISVC namespace to the mesh, if needed. It was found that sometimes the namespace enrollment process may take a little time to be processed.

Since KServe and odh-model-controllers are (to certain extent) independent of each other, sometimes KServe controller and Knative serving controller are faster and the pod from the KSVC may be created faster than the mesh enrollment process. This would lead to the KSVC pod not having an Istio sidecar.

Since the Istio authorization rules are evaluated on the sidecar of target service, the missing sidecar on the model/ksvc pod would mean that the traffic would bypass the authorization rules and this impacts ODH KServe authorization (ref: https://istio.io/latest/docs/ops/best-practices/security/#server-first-tcp-protocols-are-not-supported).

In an effort to prevent that situation, this is adding a new validating webhook that would block creation of the Knative Service until the namespace is acknowledged to be a member of the service mesh.

Fixes https://issues.redhat.com/browse/RHOAIENG-2542

## How Has This Been Tested?
Simple sanity check: ensuring a model can be deployed.

## Merge criteria:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=odh-model-controller-role,headerFile="hack/manifests_boilerplate.yaml.txt" crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) rbac:roleName=odh-model-controller-role,headerFile="hack/manifests_boilerplate.yaml.txt" crd paths="./..." output:crd:artifacts:config=config/crd/bases
 
 external-manifests: 
 	go get github.com/kserve/modelmesh-serving
--- a/config/webhook/kustomization.yaml
+++ b/config/webhook/kustomization.yaml
@@ -1,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - manifests.yaml
+  - service.yaml
--- a/controllers/reconcilers/kserve_istio_smmr_reconciler.go
+++ b/controllers/reconcilers/kserve_istio_smmr_reconciler.go
@@ -30,10 +30,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const (
-	serviceMeshMemberRollName = "default"
-)
-
 type KserveIstioSMMRReconciler struct {
 	client         client.Client
 	scheme         *runtime.Scheme
@@ -72,7 +68,7 @@ func (r *KserveIstioSMMRReconciler) Reconcile(ctx context.Context, log logr.Logg
 }
 
 func (r *KserveIstioSMMRReconciler) createDesiredResource(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) (*v1.ServiceMeshMemberRoll, error) {
-	desiredSMMR, err := r.smmrHandler.FetchSMMR(ctx, log, types.NamespacedName{Name: serviceMeshMemberRollName, Namespace: constants.IstioNamespace})
+	desiredSMMR, err := r.smmrHandler.FetchSMMR(ctx, log, types.NamespacedName{Name: constants.ServiceMeshMemberRollName, Namespace: constants.IstioNamespace})
 	if err != nil {
 		return nil, err
 	}
@@ -92,7 +88,7 @@ func (r *KserveIstioSMMRReconciler) createDesiredResource(ctx context.Context, l
 	} else {
 		desiredSMMR = &v1.ServiceMeshMemberRoll{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      serviceMeshMemberRollName,
+				Name:      constants.ServiceMeshMemberRollName,
 				Namespace: constants.IstioNamespace,
 			},
 			Spec: v1.ServiceMeshMemberRollSpec{
@@ -106,7 +102,7 @@ func (r *KserveIstioSMMRReconciler) createDesiredResource(ctx context.Context, l
 }
 
 func (r *KserveIstioSMMRReconciler) getExistingResource(ctx context.Context, log logr.Logger) (*v1.ServiceMeshMemberRoll, error) {
-	return r.smmrHandler.FetchSMMR(ctx, log, types.NamespacedName{Name: serviceMeshMemberRollName, Namespace: constants.IstioNamespace})
+	return r.smmrHandler.FetchSMMR(ctx, log, types.NamespacedName{Name: constants.ServiceMeshMemberRollName, Namespace: constants.IstioNamespace})
 }
 
 func (r *KserveIstioSMMRReconciler) processDelta(ctx context.Context, log logr.Logger, desiredSMMR *v1.ServiceMeshMemberRoll, existingSMMR *v1.ServiceMeshMemberRoll) (err error) {
@@ -143,5 +139,5 @@ func (r *KserveIstioSMMRReconciler) processDelta(ctx context.Context, log logr.L
 }
 
 func (r *KserveIstioSMMRReconciler) RemoveMemberFromSMMR(ctx context.Context, isvcNamespace string) error {
-	return r.smmrHandler.RemoveMemberFromSMMR(ctx, types.NamespacedName{Name: serviceMeshMemberRollName, Namespace: constants.IstioNamespace}, isvcNamespace)
+	return r.smmrHandler.RemoveMemberFromSMMR(ctx, types.NamespacedName{Name: constants.ServiceMeshMemberRollName, Namespace: constants.IstioNamespace}, isvcNamespace)
 }
--- a/controllers/utils/init.go
+++ b/controllers/utils/init.go
@@ -11,6 +11,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	authv1 "k8s.io/api/rbac/v1"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	knservingv1 "knative.dev/serving/pkg/apis/serving/v1"
 	maistrav1 "maistra.io/api/core/v1"
 
 	"k8s.io/apimachinery/pkg/runtime"
@@ -31,6 +32,7 @@ func RegisterSchemes(s *runtime.Scheme) {
 	utilruntime.Must(istiosecurityv1beta1.AddToScheme(s))
 	utilruntime.Must(telemetryv1alpha1.AddToScheme(s))
 	utilruntime.Must(maistrav1.SchemeBuilder.AddToScheme(s))
+	utilruntime.Must(knservingv1.AddToScheme(s))
 
 	// The following are related to Service Mesh, uncomment this and other
 	// similar blocks to use with Service Mesh
--- a/controllers/webhook/ksvc_validator.go
+++ b/controllers/webhook/ksvc_validator.go
@@ -1,0 +1,109 @@
+package webhook
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
+	"k8s.io/apimachinery/pkg/runtime"
+	types2 "k8s.io/apimachinery/pkg/types"
+	knservingv1 "knative.dev/serving/pkg/apis/serving/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/opendatahub-io/odh-model-controller/controllers/constants"
+	"github.com/opendatahub-io/odh-model-controller/controllers/resources"
+)
+
+// +kubebuilder:webhook:admissionReviewVersions=v1,path=/validate-serving-knative-dev-v1-service,mutating=false,failurePolicy=fail,groups="serving.knative.dev",resources=services,verbs=create,versions=v1,name=validating.ksvc.odh-model-controller.opendatahub.io,sideEffects=None
+
+type ksvcValidator struct {
+	client client.Client
+}
+
+func NewKsvcValidator(client client.Client) *ksvcValidator {
+	return &ksvcValidator{client: client}
+}
+
+func (v *ksvcValidator) ValidateCreate(ctx context.Context, obj runtime.Object) error {
+	log := logf.FromContext(ctx).WithName("KsvcValidatingWebhook")
+	ksvc, ok := obj.(*knservingv1.Service)
+	if !ok {
+		return fmt.Errorf("expected a Knative Service but got a %T", obj)
+	}
+
+	log = log.WithValues("namespace", ksvc.Namespace, "ksvc", ksvc.Name)
+	log.Info("Validating Knative Service")
+
+	// If there is an explicit intent for not having a sidecar, skip validation
+	ksvcTemplateMeta := ksvc.Spec.Template.GetObjectMeta()
+	if ksvcTemplateMeta != nil {
+		if templateAnnotations := ksvcTemplateMeta.GetAnnotations(); templateAnnotations[constants.IstioSidecarInjectAnnotationName] == "false" {
+			log.V(1).Info("Skipping validation of Knative Service because there is an explicit intent to exclude it from the Service Mesh")
+			return nil
+		}
+	}
+
+	// Only validate the KSVC if it is owned by KServe controller
+	ksvcMetadata := ksvc.GetObjectMeta()
+	if ksvcMetadata == nil {
+		log.V(1).Info("Skipping validation of Knative Service because it does not have metadata")
+		return nil
+	}
+	ksvcOwnerReferences := ksvcMetadata.GetOwnerReferences()
+	if ksvcOwnerReferences == nil {
+		log.V(1).Info("Skipping validation of Knative Service because it does not have owner references")
+		return nil
+	}
+	isOwnedByKServe := false
+	for _, owner := range ksvcOwnerReferences {
+		if owner.Kind == constants.InferenceServiceKind {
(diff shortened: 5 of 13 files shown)


controllers/webhook/ksvc_validator.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+// +kubebuilder:webhook:admissionReviewVersions=v1,path=/validate-serving-knative-dev-v1-service,mutating=false,failurePolicy=fail,groups="serving.knative.dev",resources=services,verbs=create,versions=v1,name=validating.ksvc.odh-model-controller.opendatahub.io,sideEffects=None
+
+type ksvcValidator struct {
+	client client.Client
+}
+
+func (v *ksvcValidator) ValidateCreate(ctx context.Context, obj runtime.Object) error {
+	log := logf.FromContext(ctx).WithName("KsvcValidatingWebhook")
```
israel-hdez, line 28: @bartoszmajsak @retocode @skonto May I have a review from you?
It doesn't need to be a deep review. I simply want to know if this approach makes sense to you.

ReToCode: I can't speak for the istio side of things (maybe there is another way to enforce this?), but from OSS perspective, having a validating webhook makes sense to me.

Makefile
```diff
@@ -34,7 +34,7 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=odh-model-controller-role,headerFile="hack/manifests_boilerplate.yaml.txt" crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```
bartoszmajsak, line 37: why is `webhook` removed here, but added elsewhere (e.g. `config/default/kustomization.yaml`)?

config/webhook/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - manifests.yaml
+  - service.yaml
```
bartoszmajsak, line 6: nit 
```suggestion
  - service.yaml
  
```

controllers/webhook/ksvc_validator.go
```diff
@@ -0,0 +1,113 @@
...
+		return fmt.Errorf("rejecting creation of Knative service %s on namespace %s because the ServiceMeshMemberRoll does not exist", ksvc.Name, ksvc.Namespace)
+	}
+
+	log = log.WithValues("smmr.namespace", smmr.Namespace, "smmr.name", smmr.Name)
+
+	if smmr.Status.Members == nil {
+		log.Info("Rejecting Knative service because the service mesh does not have any namespaces")
+		return fmt.Errorf("rejecting creation of Knative service %s on namespace %s because the namespace is yet not a member of the service mesh", ksvc.Name, ksvc.Namespace)
+	}
+	for _, memberNamespace := range smmr.Status.Members {
```
bartoszmajsak, line 94: I am not sure if it won't be better to check `ConfiguredMembers` and `PendingMembers` instead. I think `Members` is what is expected to be enrolled - so all known members disregards their statuses.

I also wonder if you don't want to check the actual state of things in `MemberStatuses` slice to get the full picture.

bartoszmajsak (changes requested):
With rejecting creation you kind of follow eventual consistency, so I think that is ok per se, especially considering the fact that those two controllers are independent of each other.

I have one thing for you to validate, and a few nits. Other than that the code looks good. Big kudos for spending time on writing tests!

> [!IMPORTANT]
> Keep in mind that at some point we will switch from creating SMM (or directly manipulating SMMR which we shouldn't be doing in the first place) to label selectors and at this point we will need to make sure we can handle this case as well. I do not know if SMMR is internally updated if that's the case, but I will figure it out.

controllers/utils/init.go
```diff
@@ -31,6 +32,7 @@ func RegisterSchemes(s *runtime.Scheme) {
 	utilruntime.Must(istiosecurityv1beta1.AddToScheme(s))
 	utilruntime.Must(telemetryv1alpha1.AddToScheme(s))
 	utilruntime.Must(maistrav1.SchemeBuilder.AddToScheme(s))
+	utilruntime.Must(knservingv1.AddToScheme(s))
```
spolti, line 35: are we enforcning knative?

Makefile
israel-hdez: I found that this generator would reset the `config/webhook/manifests.yaml` file to the bare template bundled in controller-gen, which seems to be quite partial and not final.

Since the `manifests` target is also ran in other targets (like `test`), it would be resetting the manifest constantly. So, I thought it was OK to remove it and run the webhook generator on demand when changing/adding webhooks.

controllers/webhook/ksvc_validator.go
israel-hdez: I think this is good feedback.
I changed it to use `ConfigredMembers` since that is what is documented here: https://docs.openshift.com/rosa/service_mesh/v2x/ossm-create-mesh.html

config/webhook/kustomization.yaml
israel-hdez: Fixed

controllers/utils/init.go
israel-hdez: Nope. This is, indeed, important for RawDeployment.
I think it is OK that the schema is added, regardless. However, it do is important that the webhook is not configured.

I fixed in my latest commit.

controllers/webhook/ksvc_validator.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+	}
+
+	// Only validate the KSVC if it is owned by KServe controller
+	ksvcMetadata := ksvc.GetObjectMeta()
+	if ksvcMetadata == nil {
+		log.V(1).Info("Skipping validation of Knative Service because it does not have metadata")
+		return nil
+	}
+	ksvcOwnerReferences := ksvcMetadata.GetOwnerReferences()
```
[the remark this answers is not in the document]
israel-hdez: TBH, I don't know.

Makefile
spolti: That's expected, no?
Webhook are configured with annotations in the source code (// comments for kubebuilder)

controllers/webhook/ksvc_validator.go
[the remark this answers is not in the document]
spolti: seems that, it comes from ksvc, ok := obj.(*knservingv1.Service)
Can this bring everything that is present in the cluster?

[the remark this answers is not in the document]
israel-hdez: That would be weird, IMO. At least for KServe case, I expect a single entry.
If you are using Knative directly, there would be no entries at all.

Users or other controllers can change it. It definitely can grew out of proportion, but still it would be intentional.

See [kubernetes docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/9) fore more info.

Makefile
israel-hdez: Yes, but the annotations don't let you customize everything.
* There is no way to customize the `metadata.name` that I changed to `validating.odh-model-controller.opendatahub.io`.
* It would remove the `annotations.service.beta.openshift.io/inject-cabundle` for OpenShift serving certificates.

Those are the two most important changes that I did.

So, I think the generator just gives you a YAML as a starting point. It even didn't generate the [required Service](https://github.com/opendatahub-io/odh-model-controller/pull/181/files#diff-4f466ecfd8295868474a2d73456f805b20b06d59ce1b5649777ee1c2371b9085R1), nor did the [changes to the manager](https://github.com/opendatahub-io/odh-model-controller/pull/181/files#diff-ce0d597a9e5f20a8fe61c3ba5e6185a6b90cfef2c00bb5b5ef3e97b15b79f480R77) that are required for the webhook to actually work.

VedantMahabaleshwarkar (approved):

controllers/reconcilers/kserve_istio_smmr_reconciler.go
```diff
@@ -72,7 +68,7 @@ func (r *KserveIstioSMMRReconciler) Reconcile(ctx context.Context, log logr.Logg
 }
 
 func (r *KserveIstioSMMRReconciler) createDesiredResource(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) (*v1.ServiceMeshMemberRoll, error) {
-	desiredSMMR, err := r.smmrHandler.FetchSMMR(ctx, log, types.NamespacedName{Name: serviceMeshMemberRollName, Namespace: constants.IstioNamespace})
+	desiredSMMR, err := r.smmrHandler.FetchSMMR(ctx, log, types.NamespacedName{Name: constants.ServiceMeshMemberRollName, Namespace: constants.IstioNamespace})
```
z103cb, line 71: A small nitpick on the choice of name: `ServiceMeshMemberRollName` shouldn't it be `ServiceMeshMemberRoleName`? What does Roll stand for in this context ?

main.go
```diff
@@ -172,6 +176,24 @@ func main() {
 			"reconciliation for InferenceService, please provide --model-registry-inference-reconcile flag.")
 	}
 
+	kserveWithMeshEnabled, kserveWithMeshEnabledErr := utils.VerifyIfComponentIsEnabled(context.TODO(), mgr.GetClient(), utils.KserveAuthorinoComponent)
```
z103cb, line 179: why not use `context.Background()` ?

main.go
```diff
@@ -172,6 +176,24 @@ func main() {
...
+		ksvcValidatorWebhookSetupErr := builder.WebhookManagedBy(mgr).
+			For(&knservingv1.Service{}).
+			WithValidator(webhook.NewKsvcValidator(mgr.GetClient())).
+			Complete()
+		if ksvcValidatorWebhookSetupErr != nil {
+			setupLog.Error(err, "unable to setup Knative Service validating Webhook")
+			os.Exit(1)
+		}
+	} else {
+		setupLog.Info("Skipping setup of Knative Service validating Webhook.")
```
bartoszmajsak, line 194: If you really have to log such info perhaps it's good to add "why" it is skipped.

controllers/webhook/ksvc_validator.go
```diff
@@ -0,0 +1,113 @@
...
+		return fetchSmmrErr
+	}
+	if smmr == nil {
+		log.Info("Rejecting Knative service because the ServiceMeshMemberRoll does not exist", "smmr.namespace", constants.IstioNamespace, "smmr.name", constants.ServiceMeshMemberRollName)
+		return fmt.Errorf("rejecting creation of Knative service %s on namespace %s because the ServiceMeshMemberRoll does not exist", ksvc.Name, ksvc.Namespace)
+	}
+
+	log = log.WithValues("smmr.namespace", smmr.Namespace, "smmr.name", smmr.Name)
+
+	if smmr.Status.ConfiguredMembers == nil {
```
bartoszmajsak, line 90: I am not sure about this check - this would mean none are configured, but there might be some pending or failing. I would rather check a list of all if you really need that. 

I wonder if for-loop below is not sufficient here.

controllers/webhook/ksvc_validator.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	for _, memberNamespace := range smmr.Status.ConfiguredMembers {
+		if memberNamespace == ksvc.Namespace {
+			log.V(1).Info("The Knative service is accepted")
+			return nil
+		}
+	}
+
+	log.Info("Rejecting Knative service because its namespace is not a member of the service mesh")
+	return fmt.Errorf("rejecting creation of Knative service %s on namespace %s because the namespace is not a configured member of the mesh", ksvc.Name, ksvc.Namespace)
```
bartoszmajsak, line 102: Not sure you really need a log here - the error you are returning seems to be sufficient and I assume it will be logged somehow?

In addition the info message is misleading. It can be pending or failing if I'm reading this right.

Makefile
bartoszmajsak: Can't you ensure this is defined in `kustomization.yaml` though? Or is it all overwritten?

bartoszmajsak: > I thought it was OK to remove it and run the webhook generator on demand when changing/adding webhooks.

That's based on the assumption that one would always remember to do so :)

israel-hdez: > > I thought it was OK to remove it and run the webhook generator on demand when changing/adding webhooks.
> 
> That's based on the assumption that one would always remember to do so :)

I guess it is OK to forget. If you add a new webhook, it would not work if you forget, so I would expect you would notice it is not working.

> Can't you ensure this is defined in `kustomization.yaml` though? Or is it all overwritten?
You mean to move the edits to the `kustomization.yaml` as patches?
I think I can do that, as long as I can successfully re-name it - I don't recall seeing a patch that would do renaming, as you typically use the name as a selector for patching.

Although, I'm not sure what's the best practice here:
* On KServe repo, [their webhooks](https://github.com/kserve/kserve/blob/master/config/webhook/manifests.yaml) are customized, and they also have removed the generator from their Makefile.
* On opendatahub-operator, they do keep [the autogenerated manifest](https://github.com/opendatahub-io/opendatahub-operator/blob/06256fc6bb2dfbd98e42ea95b0885598d14e2278/config/webhook/manifests.yaml), and they [do add annotations in their kustomization.yaml file](https://github.com/opendatahub-io/opendatahub-operator/blob/06256fc6bb2dfbd98e42ea95b0885598d14e2278/config/webhook/kustomization.yaml#L5-L6). With this they can keep the generator. However, one important point is that odh-operator have preserved [the usage of the namePrefix](https://github.com/opendatahub-io/opendatahub-operator/blob/06256fc6bb2dfbd98e42ea95b0885598d14e2278/config/default/kustomization.yaml#L7-L12) in the entrypoing kustomization.yaml, while here in odh-model-controller, at some point it was chosen to stop using it and that forces us to manage our names. 

I'd appreciate any guidance here.

controllers/reconcilers/kserve_istio_smmr_reconciler.go
israel-hdez: `Roll` is the right name :)
See https://docs.openshift.com/container-platform/4.15/service_mesh/v2x/ossm-create-mesh.html#ossm-about-adding-projects-using-smmr_ossm-create-mesh.

main.go
israel-hdez: No reason.
I can change it, if you think it is better.

controllers/webhook/ksvc_validator.go
israel-hdez: This is just some defensive coding. If you are confident that the `range` won't crash over a `nil` array/slice, I will trust you and remove this `if`.

controllers/webhook/ksvc_validator.go
israel-hdez: The error is not written to the controller logs. The error is returned to the client that tried to create the resource. So, the client chooses what to do with the message. Because of this, I chose to add the log message to make it clear what is happening in odh-model-controller side.

About `pending` or `failing`, I'm not sure if it is worth to go that deep here :-) Stating it is _not configured_ should is enough for me to make the user go and diagnose it, in case it stays on one of those states for longer than expected.

main.go
israel-hdez: I agree. I'll add a reason.

main.go
israel-hdez: I went ahead, and changed it.

main.go
israel-hdez: Fixed... That said, I'm no longer sure how much we can skip this. 
The conditional is in prevention of RawDeployment setup, where Knative is missing. However, AFAIK, there is no way to introduce equivalent conditionals around the `ValidatingWebhookConfiguration` resource.

So, if the user has Knative for other uses, the `ValidatingWebhookConfiguration` would take effect and would break the cluster, given the controller would have the logic turned off.

I'll tag @VaishnaviHire and @zdtsw for their advice.

controllers/webhook/ksvc_validator.go
bartoszmajsak: I am not confident, this is the spec , so don't trust me, trust the spec :) `len()` of nil slice is `0` :)

https://go.dev/play/p/cMDxMCSru4G

main.go
bartoszmajsak: I think it boils down to how we should deal with ctx propagation in general here. Setting it expliclty to `.TODO()` signals that we want to revise the whole approach and I think you can enable linter to catch those easily at the point the context propagation from reconcile down the chain is set. Not sure you are planning to do it, but I think that would be good to address this at some point.

About `context.Background()` - IIRC it is intended to be used at the highest level and is meant for background tasks. It essentially bypasses the benefits of context usage.

israel-hdez: Well... I don't plan to review later. And given this is a top-level call, I think `context.Background()` would be OK.

israel-hdez:
I just squashed the commits.
@spolti @VedantMahabaleshwarkar @bartoszmajsak Please, give it a final pass.

Makefile
spolti: if it is removing already done stuff, maybe it better to keep this way, but create a issue to keep track of it and try to rely on kubebuilder to manage the webhooks.

controllers/webhook/ksvc_validator.go
bartoszmajsak: Ok, I thought this is handled by operator-sdk, but if that is not the case logging is helpful here.

> About pending or failing, I'm not sure if it is worth to go that deep here :-) Stating it is not configured is enough for me to make the user go and diagnose it, in case it stays on one of those states for longer than expected.

I was coming from the angle that "not configured" is a bit ambiguous here. Does it mean that someone forgot to put the proper label on the namespace so it is not really seen by the mesh OR it cannot be configured by the mesh because... other errors? But I agree it might be going to deep.

bartoszmajsak (approved):
LGTM

Makefile
israel-hdez: https://issues.redhat.com/browse/RHOAIENG-4752

PR #183, opened 2024-04-02
feat(authz): soft opt-in for authz capability

bartoszmajsak:
This change handles "soft opt-in" for authorization capability handled by Authorino.

The solution is based on Conditions exposed in DSCInitialization resource, as provided in https://github.com/opendatahub-io/opendatahub-operator/pull/949.

When the condition of type `CapabilityServiceMeshAuthorization` has a reason `MissingOperator` it will assume authorino/authorization module is not present and therefore does not handle AuthConfigs - leaving models not secured as it was for the previous versions. This approach allows to softly opt-in for Authorization without blocking the upgrade. Any other reason means that ODH Operator is in not happy state and will eventually reconcile so that Authorino will start working.

The logic consists of the following steps:

 * during the controller bootstrap it will check separately if ServiceMesh is enabled and Authorino is present (latter by inspecting the status.condition)
 * when defining watches (using builder instance for OpenshiftReconciler) it will conditionally set a watch for `AuthConfig` 
 * it guards AuthConfig reconciler from executing based on the "authorization capability". This prevents constant failures due to `AuthConfig` not being defined for the cluster.

> [!NOTE]
> As it is the case with original logic brought in #181, it will require restart of the controller pod to set a watch on AuthConfig resource when Authorino is installed in the cluster.

Custom image: `quay.io/bmajsak/odh-model-controller:authz-soft-opt-in`

### DIFF
--- a/controllers/inferenceservice_controller.go
+++ b/controllers/inferenceservice_controller.go
@@ -21,6 +21,7 @@ import (
 	kservev1alpha1 "github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
 	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
 	authorinov1beta2 "github.com/kuadrant/authorino/api/v1beta2"
+	"github.com/opendatahub-io/odh-model-controller/controllers/constants"
 	"github.com/opendatahub-io/odh-model-controller/controllers/reconcilers"
 	"github.com/opendatahub-io/odh-model-controller/controllers/utils"
 	routev1 "github.com/openshift/api/route/v1"
@@ -150,26 +151,24 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
 				return reconcileRequests
 			}))
 
-	// check if kserve is enabled, otherwise don't require Authorino.
-
-	isAuthorinoRequired, err := utils.VerifyIfComponentIsEnabled(context.TODO(), r.client, utils.KserveAuthorinoComponent)
-	if err != nil {
-		r.log.V(1).Error(err, "could not determine if kserve have service mesh enabled")
+	kserveWithMeshEnabled, kserveWithMeshEnabledErr := utils.VerifyIfComponentIsEnabled(context.Background(), mgr.GetClient(), utils.KServeWithServiceMeshComponent)
+	if kserveWithMeshEnabledErr != nil {
+		r.log.V(1).Error(kserveWithMeshEnabledErr, "could not determine if kserve have service mesh enabled")
 	}
 
-	if isAuthorinoRequired {
+	authorinoEnabled, capabilityErr := utils.VerifyIfCapabilityIsEnabled(context.Background(), r.client, constants.CapabilityServiceMeshAuthorization, utils.AuthorinoEnabledWhenOperatorNotMissing)
+	if capabilityErr != nil {
+		r.log.V(1).Error(capabilityErr, "could not determine if Authorino is enabled")
+	}
+
+	if kserveWithMeshEnabled && authorinoEnabled {
+		r.log.Info("KServe is enabled and Authorino is registered, enabling Authorization capability")
 		builder.Owns(&authorinov1beta2.AuthConfig{})
-		r.log.Info("kserve is enabled with Service Mesh, Authorino is a requirement")
 	} else {
-		r.log.Info("didn't find kserve with service mesh, Authorino is not a requirement")
+		r.log.Info("didn't find kserve with service mesh, and Authorino is not a requirement")
 	}
 
-	err = builder.Complete(r)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return builder.Complete(r)
 }
 
 // general clean-up, mostly resources in different namespaces from kservev1beta1.InferenceService
--- a/controllers/inferenceservice_controller.go
+++ b/controllers/inferenceservice_controller.go
@@ -162,10 +162,12 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
 	}
 
 	if kserveWithMeshEnabled && authorinoEnabled {
-		r.log.Info("KServe is enabled and Authorino is registered, enabling Authorization capability")
+		r.log.Info("KServe with Service Mesh is enabled and Authorino is registered, enabling Authorization")
 		builder.Owns(&authorinov1beta2.AuthConfig{})
+	} else if kserveWithMeshEnabled {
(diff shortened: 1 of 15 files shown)


spolti (approved):

spolti:
Interesting failure on CI: `go.mod:3: invalid go version '1.22.0': must match format 1.23`
Not sure from where it is coming from, our go.mod is correct.


I guess you need to sign your commit :)

bartoszmajsak:
> Interesting failure on CI: `go.mod:3: invalid go version '1.22.0': must match format 1.23` Not sure from where it is coming from, our go.mod is correct.
> 
> I guess you need to sign your commit :)

That's fixed in #184 :)

About the DCO... on it. Though I'm signing commits with pgp (and they appear as `verified` here) :)

bartoszmajsak:
I think I should have make it draft from the get-go, as I would like to get your input on https://github.com/opendatahub-io/odh-model-controller/pull/183#issuecomment-2032109964.

controllers/reconcilers/kserve_authconfig_reconciler.go
```diff
@@ -56,6 +57,10 @@ func NewKserveAuthConfigReconciler(client client.Client, scheme *runtime.Scheme)
 }
 
 func (r *KserveAuthConfigReconciler) Reconcile(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+	if !utils.IsGroupRegisteredInScheme(r.client, authorinov1beta2.GroupVersion) {
+		log.V(1).Info("Skipping authorization reconciliation, authorization capability is not set up")
+		return nil
+	}
```
zdtsw, line 63: wanna make sure i understand this part:
so if odh-model-controller is only used with modelserving (as no kserve) what would happen here?

bartoszmajsak: Are you concerned about the log being confusing? This reconciler is only active when KServe is enabled so I think we should be good here.

spolti: Shouldn't we check `isAuthorinoEnabled` here as well or something similar?
AFAIK we check if the Service Mesh is enabled to allow KServe raw deployments

zdtsw: we do not have an option to set authornio to Removed or Unmanaged in operator
as long as it is installed it will be enough.

controllers/inferenceservice_controller.go
```diff
@@ -159,8 +159,13 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
 	}
 
 	if isAuthorinoRequired {
-		builder.Owns(&authorinov1beta2.AuthConfig{})
-		r.log.Info("kserve is enabled with Service Mesh, Authorino is a requirement")
+		r.log.Info("kserve is enabled with Service Mesh")
+		if utils.IsGroupRegisteredInScheme(r.client, authorinov1beta2.GroupVersion) {
+			r.log.Info("Authorino is registered, enabling Authorization capability")
+			builder.Owns(&authorinov1beta2.AuthConfig{})
+		} else {
```
Jooho, line 166: If Knative/Istio are installed on the cluster but Authorino is not installed, the odh-model-controller do not enable the authConfig reconciler, as it relies on Authorino for certain authentication functionalities. While this log shows "Authorino capability disabled", it's important to inform the user the warning message when the user attempts to enable authentication but Authorino is not installed. It should clearly explain that Authorino is required for certain authentication functionalities.

What do you think?

controllers/reconcilers/kserve_authconfig_reconciler.go
Jooho: I agreed with @spolti because authorino can be installed for other reasons by users. so we should double check it. [VerifyIComponentIsEnabled](https://github.com/opendatahub-io/odh-model-controller/blob/5fc41fb3ce6730aa628bb9a588d34a3e442d0a86/controllers/utils/utils.go#L80) func that @spolti implemented can verify if kerve serverless installed or not that require Authorino operator or not. Why don't you add the func IsGroupRegisteredInScheme into this func and use VerifyIComponentIsEnabled func?

controllers/inferenceservice_controller.go
spolti: by doing this, even though KServe requires it and authorino is missing, it will not require it leading to issuer later, right?

controllers/utils/utils.go
```diff
@@ -87,17 +85,23 @@ func VerifyIfComponentIsEnabled(ctx context.Context, cli client.Client, componen
 	// there must be only one dsc
 	if len(objectList.Items) == 1 {
 		fields := []string{"spec", "components", componentName, "managementState"}
-		if componentName == KserveAuthorinoComponent {
+		if componentName == KServeWithServiceMeshComponent {
 			// For KServe, Authorino is required when ServiceMesh is enabled
 			// By Disabling ServiceMesh for RawDeployment, it should reflect on disabling
 			// the Authorino integration as well.
```
bartoszmajsak, line 91: Nit: comment needs to be adjusted

spolti: To serving,not service mesh :)
I did in the other PR :)

controllers/reconcilers/kserve_authconfig_reconciler.go
bartoszmajsak: I changed the logic, PTAL at new version

controllers/inferenceservice_controller.go
bartoszmajsak: I changed the logic, PTAL at new version

controllers/inferenceservice_controller.go
```diff
@@ -150,26 +151,24 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
...
 	}
 
-	if isAuthorinoRequired {
+	if kserveWithMeshEnabled && authorinoEnabled {
+		r.log.Info("KServe is enabled and Authorino is registered, enabling Authorization capability")
 		builder.Owns(&authorinov1beta2.AuthConfig{})
-		r.log.Info("kserve is enabled with Service Mesh, Authorino is a requirement")
 	} else {
-		r.log.Info("didn't find kserve with service mesh, Authorino is not a requirement")
+		r.log.Info("didn't find kserve with service mesh, and Authorino is not a requirement")
```
bartoszmajsak, line 168:
```suggestion
		r.log.Info("didn't find KServe with Service Mesh, and Authorino is not installed. Skipping Authorization.")
```

Jooho (approved):
/lgtm

PR #195, opened 2024-04-11
Feat: Opt-In route for Kserve

VedantMahabaleshwarkar:
## Merge criteria:

### DIFF
--- a/controllers/reconcilers/kserve_route_reconciler.go
+++ b/controllers/reconcilers/kserve_route_reconciler.go
@@ -18,12 +18,14 @@ package reconcilers
 import (
 	"context"
 	"fmt"
+	"strconv"
+
 	"github.com/go-logr/logr"
 	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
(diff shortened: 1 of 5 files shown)


controllers/reconcilers/kserve_route_reconciler.go
```diff
@@ -73,17 +88,62 @@ func (r *KserveRouteReconciler) Reconcile(ctx context.Context, log logr.Logger,
 	return nil
 }
 
+func (r *KserveRouteReconciler) DoesIsvcNeedAnnotationMigration(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) (bool, error) {
```
israel-hdez, line 91: Maybe, the very first check is to verify if the route annotation exists. If it does, you can immediately return that no migration is needed and save a query to the API Server.

israel-hdez (commented):
Code wise, I think this approach is OK. I have a few technical comments, but I will leave that for later once this is ready for a deeper review.

That said, I'm not sure if this is the right migration path in preparation to add support to control route creation. We can discuss offline.

PR #200, opened 2024-04-29
Refactor kserve metrics solution to work with prometheus annotations

spolti:
chore:  Make sure that the vLLM metrics are getting correctly exposed when the
        prometheus annotations are set.
	Fixes [RHOAIENG-6264] - Ensure Metrics work in vLLM

Signed-off-by: Spolti <fspolti@redhat.com>

## Description

This change removes the hardcode monitoring Service, ServiceMonitor and PeerAuthentication resources in favor of a centralized configuration using the PodMonitor and Istio's ServiceMonitor resources to fit our needs and make them available in the aggregated metrics endpoint: http://<pod-ip>:15020/stats/prometheus

## How Has This Been Tested?

To test it follow the following steps (Can be tested with OVMS ServingRuntime as well):

- Have a cluster with GPU nodes available (only if testing with vLLM)
- Deploy the following ServingRuntime (If testing with vLLM): Available in the jira
- Deploy the following ISVC (If testing with vLLM): Available in the jira
- Use the Following DSC:
```yaml
spec:
  components:
    codeflare:
      managementState: Removed
    kserve:
      devFlags:
        manifests:
          - contextDir: config
            sourcePath: ''
            uri: 'https://github.com/spolti/odh-model-controller/tarball/RHOAIENG-6264-test'
      managementState: Managed
      serving:
        ingressGateway:
          certificate:
            type: SelfSigned
        managementState: Managed
        name: knative-serving
    trustyai:
      managementState: Removed
    ray:
      managementState: Removed
    kueue:
      managementState: Removed
    workbenches:
      managementState: Removed
    dashboard:
      managementState: Removed
    modelmeshserving:
      managementState: Removed
    datasciencepipelines:
      managementState: Removed
```

When everything is deployed, check the metrics in the Prometheus console, it should show the `vllm` prefixed metrics:
Screenshot 2024-04-29 at 16 21 59

## Merge criteria:

### DIFF
--- a/controllers/reconcilers/kserve_istio_peerauthentication_reconciler.go
+++ b/controllers/reconcilers/kserve_istio_peerauthentication_reconciler.go
@@ -50,7 +50,7 @@ func NewKServeIstioPeerAuthenticationReconciler(client client.Client) *KserveIst
 
 // TODO remove this reconcile loop in future versions
 func (r *KserveIstioPeerAuthenticationReconciler) Reconcile(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
-	log.V(1).Info("Reconciling PeerAuthentication for target namespace, checking if there are resource for deletion")
+	log.V(1).Info("Reconciling PeerAuthentication for target namespace, checking if there are resources for deletion")
 	// Create Desired resource
 	desiredResource, err := r.createDesiredResource(isvc)
 	if err != nil {
--- a/controllers/reconcilers/kserve_istio_peerauthentication_reconciler.go
+++ b/controllers/reconcilers/kserve_istio_peerauthentication_reconciler.go
@@ -22,10 +22,7 @@ import (
 	"github.com/opendatahub-io/odh-model-controller/controllers/comparators"
 	"github.com/opendatahub-io/odh-model-controller/controllers/processors"
 	"github.com/opendatahub-io/odh-model-controller/controllers/resources"
-	"istio.io/api/security/v1beta1"
-	istiotypes "istio.io/api/type/v1beta1"
 	istiosecv1beta1 "istio.io/client-go/pkg/apis/security/v1beta1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -51,9 +48,9 @@ func NewKServeIstioPeerAuthenticationReconciler(client client.Client) *KserveIst
 	}
 }
 
+// TODO remove this reconcile loop in future versions
 func (r *KserveIstioPeerAuthenticationReconciler) Reconcile(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
-	log.V(1).Info("Reconciling PeerAuthentication for target namespace")
-
+	log.V(1).Info("Reconciling PeerAuthentication for target namespace, checking if there are resource for deletion")
 	// Create Desired resource
 	desiredResource, err := r.createDesiredResource(isvc)
 	if err != nil {
@@ -79,25 +76,7 @@ func (r *KserveIstioPeerAuthenticationReconciler) Cleanup(ctx context.Context, l
 }
 
 func (r *KserveIstioPeerAuthenticationReconciler) createDesiredResource(isvc *kservev1beta1.InferenceService) (*istiosecv1beta1.PeerAuthentication, error) {
-	desiredPeerAuthentication := &istiosecv1beta1.PeerAuthentication{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      peerAuthenticationName,
-			Namespace: isvc.Namespace,
-		},
-		Spec: v1beta1.PeerAuthentication{
-			Selector: &istiotypes.WorkloadSelector{
-				MatchLabels: map[string]string{
-					"component": "predictor",
-				},
-			},
-			Mtls: &v1beta1.PeerAuthentication_MutualTLS{Mode: 3},
-			PortLevelMtls: map[uint32]*v1beta1.PeerAuthentication_MutualTLS{
-				8086: {Mode: 2},
-				3000: {Mode: 2},
-			},
-		},
-	}
-	return desiredPeerAuthentication, nil
+	return nil, nil
 }
 
 func (r *KserveIstioPeerAuthenticationReconciler) getExistingResource(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) (*istiosecv1beta1.PeerAuthentication, error) {
--- a/controllers/reconcilers/kserve_istio_podmonitor_reconciler.go
+++ b/controllers/reconcilers/kserve_istio_podmonitor_reconciler.go
@@ -90,8 +90,9 @@ func (r *KserveIstioPodMonitorReconciler) createDesiredResource(ctx context.Cont
 			Selector: metav1.LabelSelector{
 				MatchExpressions: []metav1.LabelSelectorRequirement{
 					{
-						Key:      "istio-prometheus-ignore",
-						Operator: metav1.LabelSelectorOpDoesNotExist,
+						Key:      "component",
+						Operator: metav1.LabelSelectorOpIn,
+						Values:   []string{"predictor", "explainer", "transformer"},
 					},
 				},
 			},
--- a/controllers/reconcilers/kserve_istio_podmonitor_reconciler.go
+++ b/controllers/reconcilers/kserve_istio_podmonitor_reconciler.go
@@ -17,11 +17,14 @@ package reconcilers
 
 import (
 	"context"
+	"fmt"
 	"github.com/go-logr/logr"
 	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
+	"github.com/kuadrant/authorino/pkg/log"
 	"github.com/opendatahub-io/odh-model-controller/controllers/comparators"
 	"github.com/opendatahub-io/odh-model-controller/controllers/processors"
 	"github.com/opendatahub-io/odh-model-controller/controllers/resources"
+	"github.com/opendatahub-io/odh-model-controller/controllers/utils"
 	v1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -53,7 +56,7 @@ func (r *KserveIstioPodMonitorReconciler) Reconcile(ctx context.Context, log log
 	log.V(1).Info("Creating Istio PodMonitor for target namespace")
 
 	// Create Desired resource
-	desiredResource, err := r.createDesiredResource(isvc)
+	desiredResource, err := r.createDesiredResource(ctx, isvc)
 	if err != nil {
 		return err
 	}
@@ -76,7 +79,12 @@ func (r *KserveIstioPodMonitorReconciler) Cleanup(ctx context.Context, log logr.
 	return r.podMonitorHandler.DeletePodMonitor(ctx, types.NamespacedName{Name: istioPodMonitorName, Namespace: isvcNs})
 }
 
-func (r *KserveIstioPodMonitorReconciler) createDesiredResource(isvc *kservev1beta1.InferenceService) (*v1.PodMonitor, error) {
+func (r *KserveIstioPodMonitorReconciler) createDesiredResource(ctx context.Context, isvc *kservev1beta1.InferenceService) (*v1.PodMonitor, error) {
+	dsciName, err := utils.GetDSCIName(ctx, r.client)
+	if err != nil {
+		log.V(1).Error(err, "Error getting DSCI name, default value will be used.")
+	}
+
 	desiredPodMonitor := &v1.PodMonitor{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      istioPodMonitorName,
@@ -95,6 +103,50 @@ func (r *KserveIstioPodMonitorReconciler) createDesiredResource(isvc *kservev1be
 				{
 					Path:     "/stats/prometheus",
 					Interval: "30s",
+					RelabelConfigs: []*v1.RelabelConfig{
+						{
+							Action:       "keep",
+							SourceLabels: []v1.LabelName{"__meta_kubernetes_pod_container_name"},
+							Regex:        "istio-proxy",
+						},
+						{
+							Action:       "keep",
+							SourceLabels: []v1.LabelName{"__meta_kubernetes_pod_annotationpresent_prometheus_io_scrape"},
+						},
+						{
+							Action:       "replace",
+							Regex:        "(\\d+);(([A-Fa-f0-9]{1,4}::?){1,7}[A-Fa-f0-9]{1,4})",
+							Replacement:  "[$2]:$1",
+							SourceLabels: []v1.LabelName{"__meta_kubernetes_pod_annotation_prometheus_io_port", "__meta_kubernetes_pod_ip"},
+							TargetLabel:  "__address__",
+						},
+						{
+							Action:       "replace",
+							Regex:        "(\\d+);((([0-9]+?)(\\.|$)){4})",
+							Replacement:  "$2:$1",
+							SourceLabels: []v1.LabelName{"__meta_kubernetes_pod_annotation_prometheus_io_port", "__meta_kubernetes_pod_ip"},
+							TargetLabel:  "__address__",
+						},
+						{
+							Action: "labeldrop",
+							Regex:  "__meta_kubernetes_pod_label_(.+)",
+						},
+						{
+							Action:       "replace",
+							SourceLabels: []v1.LabelName{"__meta_kubernetes_namespace"},
+							TargetLabel:  "namespace",
+						},
+						{
+							Action:       "replace",
+							SourceLabels: []v1.LabelName{"__meta_kubernetes_pod_name"},
+							TargetLabel:  "pod_name",
+						},
+						{
+							Action:      "replace",
+							Replacement: fmt.Sprintf("%s-istio-system", dsciName),
+							TargetLabel: "mesh_id",
+						},
+					},
 				},
 			},
 		},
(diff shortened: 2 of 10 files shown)


config/base/params.env
```diff
@@ -1 +1 @@
-odh-model-controller=quay.io/opendatahub/odh-model-controller:fast
+odh-model-controller=quay.io/spolti/odh-model-controller:RHOAIENG-6264
```
VedantMahabaleshwarkar: probably leftover from testing

controllers/reconcilers/kserve_istio_peerauthentication_reconciler.go
```diff
@@ -52,24 +52,7 @@ func NewKServeIstioPeerAuthenticationReconciler(client client.Client) *KserveIst
 }
 
 func (r *KserveIstioPeerAuthenticationReconciler) Reconcile(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
-	log.V(1).Info("Reconciling PeerAuthentication for target namespace")
```
VedantMahabaleshwarkar, line 55: If we remove the entire reconciliation process, that won't delete the existing resources for clusters doing an upgrade unless I'm mistaken. 
We'd want to modify this such that the existing resources get deleted as well. 
I think modifying the `createDesiredResource` func to return nil should make it such that when the delta is processed, the existing objects should be deleted. Since RHOAI is doing "stable" releases now, this "transition" logic should stay until the next release IMO. After that, we can remove these reconcilers altogether. 
WDYT?

controllers/reconcilers/kserve_metrics_service_reconciler.go
```diff
@@ -52,24 +52,7 @@ func NewKServeMetricsServiceReconciler(client client.Client) *KserveMetricsServi
 }
 
 func (r *KserveMetricsServiceReconciler) Reconcile(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
-	log.V(1).Info("Reconciling Metrics Service for InferenceService")
```
VedantMahabaleshwarkar, line 55: Same comment as the peerauth reconciler

controllers/reconcilers/kserve_metrics_servicemonitor_reconciler.go
```diff
@@ -47,24 +47,7 @@ func NewKServeMetricsServiceMonitorReconciler(client client.Client) *KserveMetri
 }
 
 func (r *KserveMetricsServiceMonitorReconciler) Reconcile(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
-	log.V(1).Info("Reconciling Metrics ServiceMonitor for InferenceService")
```
VedantMahabaleshwarkar, line 50: Same comment as the peerauth reconciler

VedantMahabaleshwarkar (changes requested):
I think we need to add logic such that existing resources related to the old solution are deleted.

VedantMahabaleshwarkar:
I'd also recommend changing the commit/PR title to reflect the generic solution rather than being vLLM specific. 
Maybe something like `refactor kserve metrics solution to work with prometheus annotations`

controllers/reconcilers/kserve_istio_peerauthentication_reconciler.go
spolti: that's a good idea, I didn't take into account upgrades, thanks for the heads up.

spolti: maybe making it returning nil, just add a empty object and remove the function.

controllers/reconcilers/kserve_istio_podmonitor_reconciler.go
```diff
@@ -95,6 +95,50 @@ func (r *KserveIstioPodMonitorReconciler) createDesiredResource(isvc *kservev1be
...
+							TargetLabel:  "namespace",
+						},
+						{
+							Action:       "replace",
+							SourceLabels: []v1.LabelName{"__meta_kubernetes_pod_name"},
+							TargetLabel:  "pod_name",
+						},
+						{
+							Action:      "replace",
+							Replacement: "data-science-smcp-istio-system",
```
VedantMahabaleshwarkar, line 138: We should take the control plane name and namespace from the DSCI. See [this ](https://docs.google.com/document/d/19_1YhQu73v3NEEDiSaJeJxQQ4k8RvJJIWUsV-O9YZNk/edit?disco=AAABMbqoKIw)comment from @zdtsw about how it's set in the operator

controllers/reconcilers/kserve_istio_servicemonitor_reconciler.go
```diff
@@ -93,6 +93,13 @@ func (r *KserveIstioServiceMonitorReconciler) createDesiredResource(isvc *kserve
 				{
 					Port:     "http-monitoring",
 					Interval: "30s",
+					RelabelConfigs: []*v1.RelabelConfig{
+						{
+							TargetLabel: "mesh_id",
+							Replacement: "data-science-smcp-istio-system",
```
VedantMahabaleshwarkar, line 99: Same comment as the podmonitor, re: name of the SMCP

controllers/reconcilers/kserve_istio_podmonitor_reconciler.go
zdtsw: ref: https://github.com/opendatahub-io/opendatahub-operator/blob/incubation/apis/infrastructure/v1/servicemesh_types.go#L20 for detail.

controllers/reconcilers/kserve_istio_peerauthentication_reconciler.go
spolti: ok, got it deleted by now.

controllers/reconcilers/kserve_metrics_servicemonitor_reconciler.go
spolti: fixed.

config/base/params.env
spolti: yes, was intentional to make it easier to test.

controllers/reconcilers/kserve_istio_podmonitor_reconciler.go
spolti: Update to read this value from ControlPlane spec.

spolti:
> refactor kserve metrics solution to work with prometheus annotations

Sure, updated.

controllers/utils/utils.go
```diff
@@ -162,7 +161,25 @@ func VerifyIfCapabilityIsEnabled(ctx context.Context, cli client.Client, capabil
 	}
 
 	return false, nil
+}
 
+// GetDSCIName returns the name of the DSCI in the cluster. If it fails to read the DSCI,
+// it returns the default DSCI name.
+func GetDSCIName(ctx context.Context, cli client.Client) (string, error) {
```
VedantMahabaleshwarkar, line 168: `DSCIName` would imply the name of the DSCI resource right? Whereas what we are looking for is the name of the control plane. WDYT about `GetIstioControlPlaneName` instead?

VedantMahabaleshwarkar:
Code changes look ok to me (see minor comment about function name `GetDSCIName`. Testing on a cluster now.

controllers/utils/utils.go
spolti: Yeah, that sounds good.

controllers/reconcilers/kserve_istio_podmonitor_reconciler.go
```diff
@@ -95,6 +103,50 @@ func (r *KserveIstioPodMonitorReconciler) createDesiredResource(isvc *kservev1be
...
+							Replacement:  "[$2]:$1",
+							SourceLabels: []v1.LabelName{"__meta_kubernetes_pod_annotation_prometheus_io_port", "__meta_kubernetes_pod_ip"},
+							TargetLabel:  "__address__",
+						},
+						{
+							Action:       "replace",
+							Regex:        "(\\d+);((([0-9]+?)(\\.|$)){4})",
+							Replacement:  "$2:$1",
+							SourceLabels: []v1.LabelName{"__meta_kubernetes_pod_annotation_prometheus_io_port", "__meta_kubernetes_pod_ip"},
+							TargetLabel:  "__address__",
```
israel-hdez, line 128: Is it correct that two replacements are targeting `__address__` ?

controllers/reconcilers/kserve_istio_podmonitor_reconciler.go
```diff
@@ -95,6 +103,50 @@ func (r *KserveIstioPodMonitorReconciler) createDesiredResource(isvc *kservev1be
...
+						},
+						{
+							Action:       "replace",
+							SourceLabels: []v1.LabelName{"__meta_kubernetes_pod_name"},
+							TargetLabel:  "pod_name",
+						},
+						{
+							Action:      "replace",
+							Replacement: fmt.Sprintf("%s-istio-system", istioControlPlaneName),
+							TargetLabel: "mesh_id",
```
israel-hdez, line 147: I'm not sure if this would be the same as Istio's `meshId` configuration. See: https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/ and search for "meshId".

Most likely not. If not, I'd suggest using a different name.

controllers/reconcilers/kserve_istio_servicemonitor_reconciler.go
```diff
@@ -93,6 +101,13 @@ func (r *KserveIstioServiceMonitorReconciler) createDesiredResource(isvc *kserve
 				{
 					Port:     "http-monitoring",
 					Interval: "30s",
+					RelabelConfigs: []*v1.RelabelConfig{
+						{
+							TargetLabel: "mesh_id",
```
israel-hdez, line 106: Same comment as in pod-monitor case...

controllers/reconcilers/kserve_istio_servicemonitor_reconciler.go
```diff
@@ -76,7 +79,12 @@ func (r *KserveIstioServiceMonitorReconciler) Cleanup(ctx context.Context, log l
 	return r.serviceMonitorHandler.DeleteServiceMonitor(ctx, types.NamespacedName{Name: istioServiceMonitorName, Namespace: isvcNs})
 }
 
-func (r *KserveIstioServiceMonitorReconciler) createDesiredResource(isvc *kservev1beta1.InferenceService) (*v1.ServiceMonitor, error) {
+func (r *KserveIstioServiceMonitorReconciler) createDesiredResource(ctx context.Context, isvc *kservev1beta1.InferenceService) (*v1.ServiceMonitor, error) {
```
israel-hdez, line 82: Isn't it enough with the pod monitors?

Also, the match labels is `istio=pilot`. I'd think this will match no pods, unless the ISVC is created on the istio namespace. For the latter case, I *think* (because I haven't checked) it would not fetch metrics from the ISVC, but from istio control plane pods... So, I'm confused about this.

controllers/reconcilers/kserve_istio_podmonitor_reconciler.go
VedantMahabaleshwarkar: One relabeling is for IPv4 addresses, and the other one is for IPv6 addresses.

controllers/reconcilers/kserve_istio_podmonitor_reconciler.go
israel-hdez: TIL that this is how it is documented: https://docs.openshift.com/container-platform/4.13/service_mesh/v2x/ossm-observability.html#ossm-integrating-with-user-workload-monitoring_observability.

Please, ignore my comment.

controllers/utils/utils.go
```diff
@@ -162,7 +161,25 @@ func VerifyIfCapabilityIsEnabled(ctx context.Context, cli client.Client, capabil
 	}
 
 	return false, nil
+}
 
+// GetIstioControlPlaneName returns the name of the DSCI in the cluster. If it fails to read the DSCI,
+// it returns the default DSCI name.
+func GetIstioControlPlaneName(ctx context.Context, cli client.Client) (string, error) {
+	objectList, err := getDSCIObject(ctx, cli)
```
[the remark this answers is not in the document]
spolti: Guess it is fine to query the DSCI.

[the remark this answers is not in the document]
spolti: Okay, after a good discussion on Slack we decided to use the CM, for unmamaged cases, we can instruct users to have that CM with the namesace and control plane names, will update the PR to reflect that.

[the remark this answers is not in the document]
spolti: @israel-hdez @VedantMahabaleshwarkar updated, please take a look.

controllers/reconcilers/kserve_istio_servicemonitor_reconciler.go
spolti: it is fetching the metrics correctly.
Doing a quick research: Pods with labels matching the selector criteria are considered Istio-enabled.
It seems it is correct.

israel-hdez: I understand that it works, yet I still wonder if a `ServiceMonitor` is needed?

This `ServiceMonitor` selects `Services`, not Pods. And this is filtering services having the `istio=pilot` label. On a test cluster, and after deploying an InferenceService, I don't see any `Service` that has such label on the namespace where InferenceService lives.

So, my question still stands: can we stop creating these `ServiceMonitors`?

controllers/utils/utils.go
```diff
@@ -27,6 +28,8 @@ var (
 )
 
 const (
+	defaultIstioControlPlaneName             = "data-science-smcp"
+	defaultMeshNamespace                     = "istio-system"
```
israel-hdez, line 32: Use the existing constant, if possible: https://github.com/opendatahub-io/odh-model-controller/blob/b1e69dd691bff8144ed07e703399b37af48ea0e6/controllers/constants/constants.go#L21

controllers/utils/utils.go
```diff
@@ -167,7 +168,48 @@ func VerifyIfCapabilityIsEnabled(ctx context.Context, cli client.Client, capabil
...
+// GetIstioControlPlaneName return the name of the Istio Control Plane and the mesh namespace.
+// It will first try to read the environment variables, if not found will then try to read the DSCI
+// If the required value is not available in the DSCI, it will return the default values
+func GetIstioControlPlaneName(ctx context.Context, cli client.Client) (istioControlPlane string, meshNamespace string) {
+	// first try toi retrieve it from the envs, it should be available through the service-mesh-refs ConfigMap
+	istioControlPlane = os.Getenv("CONTROL_PLANE_NAME")
+	meshNamespace = os.Getenv("MESH_NAMESPACE")
 
+	if len(istioControlPlane) == 0 || len(meshNamespace) == 0 {
+		objectList, _ := getDSCIObject(ctx, cli)
```
israel-hdez, line 182: So, the error is not handled. This is not wrong, _per se_.

However, if there is an error, then `objectList.Items` will be empty. In this case, the code inside the `for` loop will never run. So, `istioControlPlane`  and/or `meshNamespace` variables will remain unset, and empty values will be returned.

Looks like a check is needed to guarantee that, at least, default values returned.

israel-hdez (changes requested):

controllers/reconcilers/kserve_istio_servicemonitor_reconciler.go
spolti: I'll try just with PodMonitor.

In the meantime, @VedantMahabaleshwarkar, what do you think about it?
The ServiceMonitor was there already, I guess it was a suggestion from the istio team, yes?

controllers/utils/utils.go
spolti: changed, thanks.

controllers/utils/utils.go
spolti: indeed, nice catch.

controllers/reconcilers/kserve_istio_podmonitor_reconciler.go
```diff
@@ -95,6 +99,50 @@ func (r *KserveIstioPodMonitorReconciler) createDesiredResource(isvc *kservev1be
 				{
 					Path:     "/stats/prometheus",
 					Interval: "30s",
+					RelabelConfigs: []*v1.RelabelConfig{
```
VedantMahabaleshwarkar, line 102: Thinking more about this `PodMonitor` now, I am wondering if we need to add a labelselector to it. 
Consider scenario: Namespace has single model serving deployed along with other workloads (notebooks for example)
Here, I think the Podmonitor would end up creating an aggregate endpoint target even for the notebook pods which would end up being `Down`. I don't know if this even has the possibility to break anything for the other workloads (seems unlikely). But something to consider. 

Adding the labelselector however, does not have a downside even if no other workloads are present. We are just ensuring we don't create unnecessary Targets in prometheus (which would be broken anyway). 

cc @spolti @israel-hdez wdyt?

israel-hdez: I guess you are talking about improving line 90, rather than 102?
If so, your comment makes sense to me.

VedantMahabaleshwarkar: Yeah the location of the comment wasn't specific to what I was talking about here 😅 
Just want the PodMonitor to have a labelselector on it for `component: predictor` probably

israel-hdez: Yes, that makes sense. Otherwise, this can be selecting too many pods.

controllers/reconcilers/kserve_istio_servicemonitor_reconciler.go
spolti: okay, the istiod-system service monitor is not needed, going to remove it.

controllers/reconcilers/kserve_istio_podmonitor_reconciler.go
spolti: In this case, don't we also want to have explainers and transformers?
using only predictors might lead to problems if we also need to fetch metrics from the explainers in the future.

For this, I am using the label selector as In, and the values: predictor, explainer or transformer:

```yaml
  selector:
    matchExpressions:
      - key: istio-prometheus-ignore
        operator: In
        values:
        - predictor
        - explainer
        - transformer

```

spolti: with this setup it is also working and limited to objects that have the `component: predictor| explainer|transformer` label.

VedantMahabaleshwarkar (approved):
lgtm

israel-hdez (approved):
The code looks good.
I haven't been able to try the changes. I'll use the upcoming ODH release to do any needed testings there. It is not idea, but it is for what I have time.

/lgtm
Great work @spolti & @VedantMahabaleshwarkar .

PR #203, opened 2024-05-09
[RHOAIENG-6877] - odh-model-controller breaks Knative if KServe-Serve…

spolti:
…rless is not enabled

chore: Make the validating.odh-model-controller.opendatahub.io intercept only ksvc that
	   contains the `serving.kserve.io/inferenceservice` label. This label is added by KServe
	   to any ksvc handled by it, this way we avoid other OpenShift Serverless user not be
	   affected by this webhook in case it is not started by odh-model-controller.

## Description

**Steps to test:**

- Deploy the RHOAI Operator with KServe and ServiceMesh disable, make sure the `odh-model-controller`  does not starts the webhook.
- Try to deploy the ksvc again, you will notice that it will fail indicating the validating webhook is not available.

**Applying the patch:**

- Update the DSC with this content:
```yaml
spec:
  components:
    codeflare:
      managementState: Removed
    kserve:
      managementState: Removed
      serving:
        ingressGateway:
          certificate:
            type: SelfSigned
        managementState: Removed
        name: knative-serving
    trustyai:
      managementState: Removed
    ray:
      managementState: Removed
    kueue:
      managementState: Removed
    workbenches:
      managementState: Removed
    dashboard:
      managementState: Removed
    modelmeshserving:
      devFlags:
        manifests:
          - contextDir: config
            sourcePath: 'base'
            uri: 'https://github.com/spolti/odh-model-controller/tarball/RHOAIENG-6877-test'
      managementState: Managed
    datasciencepipelines:
      managementState: Removed
```

- After the reconcile loops have ended, make sure that `odh-model-operator` didn't start the webhook, and can be checked in the logs.
- try to deploy the ksvc again. It should succeed as the `objectSelector` is only filtering, to confirm that, add the kserve label `serving.kserve.io/inferenceservice` with any value and try to deploy the ksvc again, it will fail. 

Note that, if the odh-model-controller image does not starts and it is not overridden, the test can be done anyways as the webhook will not be functional.
The validation webhook is correctly updated, you can check by querying it:
```bash
oc get ValidatingWebhookConfiguration validating.odh-model-controller.opendatahub.io -n redhat-ods-applications -o yaml
```

And check for these line:
```yaml
  objectSelector:
    matchExpressions:
    - key: serving.kserve.io/inferenceservice
      operator: Exists

```

## Merge criteria:

### DIFF
--- a/config/webhook/webhook_patch.yaml
+++ b/config/webhook/webhook_patch.yaml
@@ -1,0 +1,16 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: validating.odh-model-controller.opendatahub.io
+  annotations:
+    service.beta.openshift.io/inject-cabundle: true
+webhooks:
+  - name: validating.ksvc.odh-model-controller.opendatahub.io
+    clientConfig:
+      service:
+        name: odh-model-controller-webhook-service
+    objectSelector:
+      matchExpressions:
+        - key: serving.kserve.io/inferenceservice
+          operator: Exists
+
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,8 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=odh-model-controller-role,headerFile="hack/manifests_boilerplate.yaml.txt" crd paths="./..." output:crd:artifacts:config=config/crd/bases
+	# Any customization needed, apply to the webhook_patch.yaml file
+	$(CONTROLLER_GEN) rbac:roleName=odh-model-controller-role,headerFile="hack/manifests_boilerplate.yaml.txt" crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
 external-manifests: 
 	go get github.com/kserve/modelmesh-serving
--- a/config/webhook/field_patch.yaml
+++ b/config/webhook/field_patch.yaml
@@ -1,0 +1,3 @@
+- op: replace
+  path: /metadata/name
+  value: validating.odh-model-controller.opendatahub.io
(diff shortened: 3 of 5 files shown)


config/base/params.env
```diff
@@ -1 +1 @@
-odh-model-controller=quay.io/opendatahub/odh-model-controller:fast
+odh-model-controller=quay.io/spolti/odh-model-controller:RHOAIENG-6264
```
terrytangyuan, line 1: remove testing changes

spolti: it was added for testing.

config/webhook/webhook_patch.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: validating.odh-model-controller.opendatahub.io
+  annotations:
+    service.beta.openshift.io/inject-cabundle: true
+webhooks:
+  - name: validating.ksvc.odh-model-controller.opendatahub.io
+    clientConfig:
+      service:
+        name: odh-model-controller-webhook-service
```
israel-hdez, line 11: May the patch include the namespace?

Running `kustomize build config/base` on the `main` branch, then on this PR, and doing a diff of both results shows that the `clientConfig` keeps `system` namespace, which would be a bug:

image

spolti: Hmm, yesterday, we talked about it and you suggested not to patch the namespace as it would be replaced anyway.
I can revert back to the patch I have added.

spolti: From the cluster:
```yaml
    service:
      name: odh-model-controller-webhook-service
      namespace: redhat-ods-applications
      path: /validate-serving-knative-dev-v1-service
      port: 443
  failurePolicy: Fail
  matchPolicy: Equivalent
  name: validating.ksvc.odh-model-controller.opendatahub.io
  namespaceSelector: {}
  objectSelector:
    matchExpressions:
    - key: serving.kserve.io/inferenceservice
      operator: Exists

```

israel-hdez: OK, I just tested with odh-operator and the result is the expected.
It looks like the odh-operator is doing some additional processing than kustomize.

israel-hdez (approved):
/lgtm

PR #204, opened 2024-05-14
Kservedashboard

ashley-o0o:
## Merge criteria:


spolti:
Hi @ashley-o0o it seems to be unrelated changes, yes?
You might need to rebase.

Plus, sign off your commit.

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -0,0 +1,238 @@
...
+				Namespace: isvc.Namespace,
+			}, isvcServingRuntimeList.Items[i])
+
+			// if there is error with getting the runtimes, throw an error
+			if err != nil {
+				if apierrs.IsNotFound(err) {
+					return nil, err
+				}
+			}
+	}
```
VedantMahabaleshwarkar, line 116: each `InferenceService` will trigger the reconcile loop separately, so we don't need to fetch the list. So within the reconcile loop, include logic that you want to be executed for that 1 isvc. 

So here we can get the `ServingRuntime` for the `InferenceService` _that triggered this loop_ as follows : 
```suggestion
        isvcRuntime := isvc.Spec.Predictor.Model.Runtime
	runtime := &kservev1alpha1.ServingRuntime{}
	if err := r.client.Get(ctx, types.NamespacedName{Name: *isvcRuntime, Namespace: isvc.Namespace}, runtime); err != nil {
		log.Error(err, "Could not determine servingruntime for isvc")
	}
```

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -0,0 +1,238 @@
...
+						log.Error(err, "Unable to load tgis-metrics file")
+					}
+
+		case vllm: 
+			if vllmData == nil : 
+					vllmData, err := os.ReadFile("vllm-metrics.json") 
+					if err != nil {
+						log.Error(err, "Unable to load vllm-metrics file")
+					}
+	}
```
VedantMahabaleshwarkar, line 145: is this block still WIP?

VedantMahabaleshwarkar (changes requested):

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
ashley-o0o: Yes it's still WIP

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -0,0 +1,223 @@
...
+		if ovmsData == nil {
+			data, err := os.ReadFile("ovms-metrics.json")
+			if err != nil {
+				log.Error(err, "Unable to load metrics dashboard template file:", err)
+			}
+			err = json.Unmarshal(data, &ovmsData)
+			if err != nil {
+				log.Error(err, "Error unmarshalling JSON:", err)
+			}
+		}
```
VedantMahabaleshwarkar, line 122: I tried to play around with this myself. I think this also works without having to unmarshal anything at all. 
I think it's enough to just read the file and directly process the `[] byte` returned. Also we will have to declare `data` outside the switch block and give it the correct value in the switch loop, such that it contains the correct data once we exit the switch. 
```suggestion
        var data []byte
	switch servingRuntime {
	case "openvino_model_server":
		if ovmsData == nil {
			ovmsData, err := os.ReadFile("ovms-metrics.json")
			if err != nil {
				log.Error(err, "Unable to load metrics dashboard template file:", err)
			}
			data = ovmsData
		}
```

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -0,0 +1,223 @@
...
+	case "text-generation-inference":
+		if tgisData == nil {
+			data, err := os.ReadFile("tgis-metrics.json")
+			if err != nil {
+				log.Error(err, "Unable to load metrics dashboard template file:", err)
+			}
+			err = json.Unmarshal(data, &tgisData)
```
VedantMahabaleshwarkar, line 129: same comment as for the ovms block

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -0,0 +1,223 @@
...
+			}
+		}
+	}
+
+	var configMapData MetricsDashboardConfigMapData
+	//TODO: move read file logic to switch and only read if global variable is nil
+	data, err := os.ReadFile("ovms-metrics.json")
+	if err != nil {
+		log.Error(err, "Unable to load metrics dashboard template file")
+	}
```
VedantMahabaleshwarkar, line 141: don't need this anymore right? 
```suggestion
```

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -0,0 +1,223 @@
...
+	// Create ConfigMap object
+	configMap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      isvc.Name + "-metrics-dashboard",
+			Namespace: isvc.Namespace,
+		},
+		Data: map[string]string{
+			"Data": string(jsonData), //TODO: multiple runtimes support
+		},
+	}
```
VedantMahabaleshwarkar, line 165: might be good to separate the string replacement logic into a function for readability. Also since this is the final processing step of the data before we create the configmap, we can pass the function the bytearray we have in `data` and have it return the string that we will use in the configmap. 
```suggestion
	finaldata := substituteVariablesInQueries(data, isvc.Namespace, isvc.Name)
	// Create ConfigMap object
	configMap := &corev1.ConfigMap{
		ObjectMeta: metav1.ObjectMeta{
			Name:      isvc.Name + "-metrics-dashboard",
			Namespace: isvc.Namespace,
		},
		Data: map[string]string{
			"metrics": finaldata,
		},
	}
```

VedantMahabaleshwarkar (changes requested):

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -0,0 +1,198 @@
...
+		if tgisData == nil {
+			tgisData, err := os.ReadFile("tgis-metrics.json")
+			if err != nil {
+				log.Error(err, "Unable to load metrics dashboard template file:", err)
+			}
+			data = tgisData
+		}
+	}
+
+	finaldata := substituteVariablesInQueries(data, isvc.Namespace, isvc.Name)
```
VedantMahabaleshwarkar, line 129: You'd need to define and write the logic for this function `substituteVariablesInQueries`

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -0,0 +1,202 @@
...
+			Name:      isvc.Name + "-metrics-dashboard",
+			Namespace: isvc.Namespace,
+		},
+		Data: map[string]string{
+			"metrics": finaldata,
+		},
+	}
+	// Add labels to the configMap
+	configMap.Labels = map[string]string{
+		"app.opendatahub.io/kserve": "true",
```
spolti, line 143: it might be useful to add the servingRuntime name here as well, wdyt?

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -0,0 +1,202 @@
...
+	findImageName := re.FindStringSubmatch(servingRuntimeImage)
+	servingRuntime := findImageName[1]
+
+	var data []byte
+	switch servingRuntime {
+	case "openvino_model_server":
+		if ovmsData == nil {
+			ovmsData, err := os.ReadFile("ovms-metrics.json")
+			if err != nil {
+				log.Error(err, "Unable to load metrics dashboard template file:", err)
```
spolti, line 116: if there are any failures to load the data, is it safe to continue the execution?
because the data will be empty and the configMap will be created with no data on it.

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -0,0 +1,202 @@
...
+	// resolve SR
+	isvcRuntime := isvc.Spec.Predictor.Model.Runtime
+	runtime := &kservev1alpha1.ServingRuntime{}
+	if err := r.client.Get(ctx, types.NamespacedName{Name: *isvcRuntime, Namespace: isvc.Namespace}, runtime); err != nil {
+		log.Error(err, "Could not determine servingruntime for isvc")
+	}
+
+	servingRuntimeImage := runtime.Spec.Containers[0].Image
+	re := regexp.MustCompile(`/([^/@]+)[@:]`)
+	findImageName := re.FindStringSubmatch(servingRuntimeImage)
```
spolti, line 107: not sure if it is safe to match the image name, custom serving runtimes for openvino could have different image name than `openvino_model_server`

Wondering if there is a better way to do this.

Maybe it is safer to match the supported formats?

@Xaenalt @terrytangyuan wdyt?

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -0,0 +1,202 @@
...
+	if err := ctrl.SetControllerReference(isvc, configMap, r.client.Scheme()); err != nil {
+		log.Error(err, "Unable to add OwnerReference to the Metrics Dashboard Configmap")
+		return nil, err
+	}
+
+	return configMap, nil
+}
+
+func substituteVariablesInQueries(data string, namespace string, name string) string {
+	return strings.ReplaceAll(strings.ReplaceAll(data, "${namespace}", namespace), "${model_name}", name)
```
spolti, line 154: in the files proposed here https://github.com/opendatahub-io/odh-model-controller/pull/209/files these two envs are uppercase.

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	if err := ctrl.SetControllerReference(isvc, configMap, r.client.Scheme()); err != nil {
+		log.Error(err, "Unable to add OwnerReference to the Metrics Dashboard Configmap")
+		return nil, err
+	}
+
+	return configMap, nil
+}
+
+func substituteVariablesInQueries(data string, namespace string, name string) string {
```
spolti, line 153: maybe move it to the utils package?

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -0,0 +1,202 @@
...
+	servingRuntimeImage := runtime.Spec.Containers[0].Image
+	re := regexp.MustCompile(`/([^/@]+)[@:]`)
+	findImageName := re.FindStringSubmatch(servingRuntimeImage)
+	servingRuntime := findImageName[1]
+
+	var data []byte
+	switch servingRuntime {
+	case "openvino_model_server":
+		if ovmsData == nil {
+			ovmsData, err := os.ReadFile("ovms-metrics.json")
```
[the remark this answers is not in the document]
VedantMahabaleshwarkar: nope, I don't believe the logic is added yet but we'll watch the configmap for changes and revert to the original state if the user changes anything manually.

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
VedantMahabaleshwarkar: This was my idea. I landed on using this as the best approach to distinguish "supported" because the support matrix for runtimes and modelformats is really weird. For e.g: OVMS supports pytorch, but so do a lot of other runtimes that we don't include. So it's difficult to make an assumption just based on the format. This approach was selected based on. "We'll support OOTB runtimes, and any clones of the OOTB runtimes you make(for e.g to customize any parameter)." 
I'm curious to see if there's a better approach, but I don't think supportedformats is the way to go.

controllers/constants/constants.go
```diff
@@ -52,4 +52,5 @@ const (
 
 const (
 	DefaultStorageConfig = "storage-config"
+	KServeImageName      = "openvino_model_server"
```
spolti, line 55: what is the purpose of this constant?

ashley-o0o:
PR work moved to new PR - [220](https://github.com/opendatahub-io/odh-model-controller/pull/220)

PR #210, opened 2024-05-30
Upgrade kserve version to v0.12.1 in odh-model-controller

vaibhavjainwiz:
## Description
modelmesh-serving has already been upgraded to Kserve v0.12 but odh-model-controller is left behind. With this PR I mean to upgrade the kserve version `odh-model-controller` to v0.12.

## Merge criteria:


Containerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM registry.access.redhat.com/ubi8/go-toolset:1.19 as builder
+FROM registry.access.redhat.com/ubi8/go-toolset:1.21 as builder
```
Jooho, line 2: could you please use ubi9?

vaibhavjainwiz: done

controllers/webhook/ksvc_validator.go
```diff
@@ -26,11 +27,11 @@ func NewKsvcValidator(client client.Client) *ksvcValidator {
 	return &ksvcValidator{client: client}
 }
 
-func (v *ksvcValidator) ValidateCreate(ctx context.Context, obj runtime.Object) error {
+func (v *ksvcValidator) ValidateCreate(ctx context.Context, obj runtime.Object) (warnings admission.Warnings, err error) {
```
davidesalerno, line 30: Why do you need to add an additional return object (warnings) that seems to be always null?

vaibhavjainwiz: its a new method signature of `ValidateCreate` in controller-runtime v0.16.3. 
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.4/pkg/webhook/admission/validator.go#L42

PR #213, opened 2024-05-30
Auth race condition mitigation

israel-hdez:
A race condition was found that leads to odh-model-controller starting in a state not suitable for KServe, despite KServe is installed with authorization enabled.

This mitigates such condition by:
* Always adding AuthConfigs to the schema. Adding types to the schema doesn't seem to have any bad effects, even if such types does not exist in the cluster. This prevents a "no kind is registered" error when trying to reconcile an InferenceService if Authorino setup finished after odh-model-controller booted.
* Invoking `Owns` on inferenceservice_controller.go setup based on the existence of the AuthConfig CRD in the cluster, rather than based on the existence of a specific AuthorizationPolicy.

These changes, together with opendatahub-io/opendatahub-operator#1019 should mitigate/fix the race condition and odh-model-controller should properly start in a good state suitable for KServe.

Related to https://issues.redhat.com/browse/RHOAIENG-7312

## How Has This Been Tested?
Three different test were done:
* Install only ModelMesh (no service mesh, nor knative, nor authorino operators installed), and check that odh-model-controller would start properly.
* Install KServe but skip installation of Authorino operator. Check that odh-model-controller would start properly. Also, check that a model can be deployed and inference requests work.
* Install KServe with all prerequisites. Check that odh-model-controller would start properly. Also, check that a model can be deployed and inference requests work.

## Merge criteria:

### DIFF
--- a/controllers/inferenceservice_controller.go
+++ b/controllers/inferenceservice_controller.go
@@ -155,16 +155,17 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
 		r.log.V(1).Error(kserveWithMeshEnabledErr, "could not determine if kserve have service mesh enabled")
 	}
 
-	authorinoEnabled, capabilityErr := utils.VerifyIfMeshAuthorizationIsEnabled(context.Background(), r.client)
-	if capabilityErr != nil {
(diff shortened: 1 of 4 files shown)


VedantMahabaleshwarkar (approved):
Didn't test the changes on a cluster, but didn't see any issues on the code. 
/approve

PR #217, opened 2024-06-04
Fix performance issues in odh model controller

vaibhavjainwiz:
## Description
The ODH model controller is currently using far too much memory and is getting OOMKilled.

### Replication
I am able to replicate this issue with 2000 namespaces where each namespace has 20 secrets of 1mb each.

### Analysis
As analyzed, odh-model-controller consumes a lot of memory because it loads all the Secret of all namespaces into in-memory controller pod.
It happened because we have defined Secret in a list of watch objects inside OpenshiftInferenceServiceReconciler.

### Details analysis
When we define any Object kind in the Watch list(as shown below), the operator would maintain the cache for all instances of watched object type. In short, the operator loads all the object of the watch type from all namespaces into the cache.

```
builder := ctrl.NewControllerManagedBy(mgr).
    For(&kservev1beta1.InferenceService{}).
    Owns(&kservev1alpha1.ServingRuntime{}).
    Owns(&corev1.Namespace{}).
    Owns(&routev1.Route{}).
    Owns(&corev1.ServiceAccount{}).
    Owns(&corev1.Service{}).
    Owns(&corev1.Secret{}).
    Owns(&authv1.ClusterRoleBinding{}).
    Owns(&networkingv1.NetworkPolicy{}).
    Owns(&monitoringv1.ServiceMonitor{}).
    Owns(&monitoringv1.PodMonitor{}).
    Watches(&kservev1alpha1.ServingRuntime{}, 
```

For Example : In above code block Secret is defined in the watch object list so by default Operator would load all Secret from all namespaces into the controller pod.

### Solution
**Solution 1 :** We should carefully define the watcher. If there is any object we should not watch then we should remove it from the watcher list. For example in OpenshiftInferenceServiceReconciler we had defined Secret in the watch list but we didn't create or maintain any Secret in that reconciler so we should remove it.

**Solution 2 :** Cache for watch list is maintained at operator level. Let's suppose if we have 4 controllers and only one of them requires to watch Secret. Even In that case all secrets from all namespace would be loaded into operator memory. If this type of scenario happened then Solution 1 would not be much helpful.
In this case the best possible solution is to restrict the resources which the Operator could monitor. We could configure the controller by defining the cache options while creating a manager instance.

```
mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
.
.
.
Cache: cache.Options{},
} 
```

Cache restriction could mainly be define at two levels :
**Namespace Level** : Cache all resources from specific namespace.
**Object Level** : Cache Object of kind using selector.

**Namespace Level cache**
In this approach, we could provide a list of namespaces which we want to monitor and the controller would only load the objects from these namespaces only. 

Sample implementation :
```
mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
.
.
.
Cache: cache.Options{
  DefaultNamespaces: map[string]cache.Config{
        "opendatahub": {
          LabelSelector: labels.SelectorFromSet(labels.Set{
                      "opendatahub.io/managed": "true",
          }),
        },
      },
},
}
```

**Challenge:**
We can't provide label_selector to select the namespaces. We can't dynamically select a namespace to cache. If we want to restrict by namespaces then all cache namespace should be provided at time of starting operator.
Providing the name of all inference namespaces at time of starting operator is very unlikely because RHOAI provides functionality to create new projects any time which is equivalent to a namespace.

**Object Level cache**
We could define the selector on object type in below format :
```
mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
    .
    .
    .
    Cache: cache.Options{
      ByObject: map[client.Object]cache.ByObject{
        &v1.Secret{}: {
          Label: labels.SelectorFromSet(labels.Set{
              "opendatahub.io/managed": "true",
              "opendatahub.io/dashboard": "true",
          }),
        },
        },
      },
    },
  })
```

We could define the selector in terms of a different combination of Label and Field selector.
With the above cache configuration, the operator has created a separate cache to keep just only secrets which have labels opendatahub.io/managed=true && opendatahub.io/dashboard=true and separate cache to keep all objects of all namespace apart from secret.

**Cache and Client working together**
I do some quick tests on client and cache behavior together. Client also work on what we define in cache.
It means if in the cache option we put a selector to select a secret whose name is storage-config then the client would only be able to query a secret whose name is storage-config. If the client tries to fetch other secrets whose name is [test-1](https://issues.redhat.com//browse/test-1) then it would return an not found error.

In our context it means if we decide to put a selector on Secret then we need to make sure the selector covers all the secrets that we are going to use in odh-model-controller.

**Selector challenge** 
Multiple values in the label/field selector work with AND operation. 
For example in below code
```
Cache: cache.Options{
      ByObject: map[client.Object]cache.ByObject{
        &v1.Secret{}: {
          Label: labels.SelectorFromSet(labels.Set{
              "opendatahub.io/managed": "true",
              "opendatahub.io/dashboard": "true",
          }),
        },
        },
      },
```

Operator would only cache those secrets which have both labels(opendatahub.io/managed=true , opendatahub.io/dashboard=true).
This type of selectors works fine with Secret because all Secret in odh-model-controller have common labels but It causes issues with ConfigMap selection.
As per my analysis, we are querying below 3 ConfigMaps and none of them have a common label/field.

- inferenceservice-config
- odh-kserve-custom-ca-bundle
- odh-trusted-ca-bundle

## How Has This Been Tested?
To test it follow the following steps :
- Have a cluster with 300 namespace where each namespace has 20 secrets of 1mb each.
- Deploy the ServingRuntime
- Deploy the ISVC
- Use the Following DSC:
```
spec:
  components:
    codeflare:
      managementState: Removed
    kserve:
      devFlags:
        manifests:
          - contextDir: config
            sourcePath: ''
            uri: 'https://github.com/vaibhavjainwiz/odh-model-controller/tarball/RHOAIENG-1006_test'
      managementState: Managed
      serving:
        ingressGateway:
          certificate:
            type: SelfSigned
        managementState: Managed
        name: knative-serving
    trustyai:
      managementState: Removed
    ray:
      managementState: Removed
    kueue:
      managementState: Removed
    workbenches:
      managementState: Removed
    dashboard:
      managementState: Removed
    modelmeshserving:
      managementState: Removed
    datasciencepipelines:
      managementState: Removed
```
### Test Result
Memory consuption of odh-model-controller should be around ~150mb.

## Merge criteria:

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,9 +20,12 @@ import (
 	"context"
 	"flag"
 	"github.com/opendatahub-io/odh-model-controller/controllers/webhook"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	knservingv1 "knative.dev/serving/pkg/apis/serving/v1"
 	"os"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"strconv"
 
@@ -124,6 +127,15 @@ func main() {
 		Client: client.Options{
 			Cache: &client.CacheOptions{
 				DisableFor: []client.Object{&v1beta1.AuthorizationPolicy{}},
+			},
+		},
+		Cache: cache.Options{
+			ByObject: map[client.Object]cache.ByObject{
+				&v1.Secret{}: {
+					Label: labels.SelectorFromSet(labels.Set{
+						"opendatahub.io/managed": "true",
+					}),
+				},
 			},
 		},
 	})


main.go
```diff
@@ -126,6 +129,15 @@ func main() {
 				DisableFor: []client.Object{&v1beta1.AuthorizationPolicy{}},
 			},
 		},
+		Cache: cache.Options{
+			ByObject: map[client.Object]cache.ByObject{
+				&v1.Secret{}: {
+					Label: labels.SelectorFromSet(labels.Set{
+						"opendatahub.io/managed": "true",
```
danielezonca, line 136: We should document the usage of this annotation, with this PR you are making it part of the contract: if the secrets has no annotation it will not be monitored.
I just want to make sure we keep track of the usage of this because it can easily break given that it is an impl detail.

@zdtsw @VaishnaviHire 
Do you have any advice here? Is the usage of this annotation documented/part of the "API" of the operator?

zdtsw: i think we only used annotation opendatahub.io/managed in Operator, not label. and it was introduced when we added support for RawDeployment.
I dont think this annotation has been documented anywhere

jwforres: Should we be using `app.kubernetes.io/managed-by` for this instead?
https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-managed-by

vaibhavjainwiz: @danielezonca @zdtsw  With Selector approach we could only apply filters overs Labels and Fields. It doesn't support selector over annotations. For this approach to work we need to make sure all secrets which we want to watch should have `opendatahub.io/managed: true` label.

IF we don't follow the standard approach to add `opendatahub.io/managed: true` label then we need to think about using namespace selector.

vaibhavjainwiz: > Should we be using `app.kubernetes.io/managed-by` for this instead? https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-managed-by

`app.kubernetes.io/managed-by` is not maintined in our application.

zdtsw: @danielezonca  I think we are a bit derived from original PR.  if the plan is to use a label on the resources (specific in this case for CM and secret) then it is not the same thing in Operator, even they are set to the same key-value (opendatahub.io/managed: true). 

@vaibhavjainwiz i am not sure by introducing `opendatahub.io/managed` here.  is it only intended for modelmesh(serving)'s resources or is planned to extend to all ODH components later(like an unified way across platform). if it is the former one, isnt `app.kubernetes.io/part-of:model-mesh` or `app.opendatahub.io/model-mesh: 'true' `labels enough? 

@jwforres we have a draft https://github.com/opendatahub-io/architecture-decision-records/pull/26 on the labeling but did not get work continued.  :cry:

vaibhavjainwiz: >@vaibhavjainwiz i am not sure by introducing opendatahub.io/managed here. is it only intended for modelmesh(serving)'s resources or is planned to extend to all ODH components later(like an unified way across platform). if it is the former one, isnt app.kubernetes.io/part-of:model-mesh or app.opendatahub.io/model-mesh: 'true' labels enough?

OOM in operator is common issue in almost all operator that we are facing with a client which have 2000 namespace and 49809 secrets. I had started a [doc](https://docs.google.com/document/d/1vr5Q6jn_PNYaNBW69ju5TaMHOhzCl337-GM1fBR8biU/edit#heading=h.mf8y6r2wogks) to describe this problem and its solutions. We have stated to fix this issue with **odh-model-controller** first and their share our finding with other team.

odh-model-controller query the Secret which are not labeled with `app.kubernetes.io/part-of:model-mesh`. Like Data Connection secret which is created by dashboard.

israel-hdez: > We should document the usage of this annotation, with this PR you are making it part of the contract:

FWIW, it seems that this "contract" has been there for a while, because the watchers are already doing filtering by this label at client side. See: https://github.com/opendatahub-io/odh-model-controller/blob/ead3013973795f8a6965d69938e0ef3ae2a1c1e0/controllers/storageconfig_controller.go#L177-L197

That seems to introduced by the end of 2022: https://github.com/opendatahub-io/odh-model-controller/commit/de166dc8a63171661abe1ee139d0c75b266ca79c. So, I think limiting the cache using that same label won't make any difference to usability. Indeed, it looks like there is a KB article in Red Hat pointing the requirement of such labels: https://access.redhat.com/solutions/7067310.

We may want to later review if we should be doing such filtering, but given current state I think this PR is on the right track.

What I think this PR is not taking into account is that odh-model-controller is creating subordinate secrets based on the watched ones. Such secrets do not have the `opendatahub.io/managed` labels, thus healing won't work. So, @vaibhavjainwiz I think you should review that... IDK if healing works currently. If it doesn't work, well, we may be good with this PR. But if it works, I'd see it as a regression.

vaibhavjainwiz: @israel-hdez **odh-model-controller** creates the `storage-config` secret which include `opendatahub.io/managed` label by watching dataConnectionSecretsList. 

https://github.com/opendatahub-io/odh-model-controller/blob/1ab78dfe482ce1b74b58fbe19c7e344ca69298a1/controllers/storageconfig_controller.go#L120

israel-hdez: > @israel-hdez **odh-model-controller** creates the `storage-config` secret which include `opendatahub.io/managed` label by watching dataConnectionSecretsList.

...good! Then I don't see anything wrong with your PR.

Jooho: > return labels["opendatahub.io/managed"] == "true" 

For platform resources, it is ok but why don't we have another label for modelserving team?  something like

```
return labels["opendatahub.io/component"] == "modelserving" 
```

one more label can give us some flexibility to work with.

vaibhavjainwiz: > > return labels["opendatahub.io/managed"] == "true"
> 
> For platform resources, it is ok but why don't we have another label for modelserving team? something like
> 
> ```
> return labels["opendatahub.io/component"] == "modelserving" 
> ```
> 
> one more label can give us some flexibility to work with.

@Jooho nice thought. Could you please comment it on https://github.com/opendatahub-io/architecture-decision-records/pull/26 ADR. Its for defining unified way of labeling for RHOAI.

israel-hdez (approved):

spolti (approved):

lugi0:
Hi @vaibhavjainwiz, I've tested this on top of RHOAI 2.10 RC2.
I can confirm that before applying the manifests the memory consumption of odh-model-controller would keep climbing
Screenshot 2024-06-14 at 16 56 21
before-manifest
And after applying the manifests the memory consumption dropped drastically. 
Screenshot 2024-06-14 at 16 55 45

I unfortunately don't have a screenshot of this, but the only weird thing I have noticed is that after applying the manifests but before deploying a model (so before creating the SR/ISVC) the memory consumption was around ~210 MiB. After deploying the model it dropped down to ~150 as expected and shown in the screenshot. 
I also tried removing the model and its namespace to see if it would climb back up again, but it remained at ~150.
Do you have any idea what might have caused this?
Regardless I'm approving the PR since overall it works as we expect it to.

lugi0 (approved):

vaibhavjainwiz:
> Hi @vaibhavjainwiz, I've tested this on top of RHOAI 2.10 RC2. I can confirm that before applying the manifests the memory consumption of odh-model-controller would keep climbing Screenshot 2024-06-14 at 16 56 21 before-manifest And after applying the manifests the memory consumption dropped drastically. Screenshot 2024-06-14 at 16 55 45
> 
> I unfortunately don't have a screenshot of this, but the only weird thing I have noticed is that after applying the manifests but before deploying a model (so before creating the SR/ISVC) the memory consumption was around ~210 MiB. After deploying the model it dropped down to ~150 as expected and shown in the screenshot. I also tried removing the model and its namespace to see if it would climb back up again, but it remained at ~150. Do you have any idea what might have caused this? Regardless I'm approving the PR since overall it works as we expect it to.

@lugi0 Services usually take more space on startup because it loads lots of objects in memory. Some memory released later because garbage collection process. It nothing related to loading model. I had also observed similiar behaviour with any Inference as well.

PR #221, opened 2024-06-12
Feature: Create cert Secret and update KServe local gateway

Jooho:
Description:

This is one of tasks for Private Endpoint Tasks (https://github.com/opendatahub-io/kserve/issues/371)

This PR support the following:
- add Serving Cert annotation to runtime Service
- create the same secret that was created by Serving cert in the istio-system
- update kserve local gateway for the isvc

Jira: https://issues.redhat.com/browse/RHOAIENG-7921

**(NOTE) Loopy was tested on fedora only.**

### Env setup
* Using source
  ~~~
  mkdir /tmp/test
  cd /tmp/test
  git clone git@github.com:Jooho/loopy.git
  cd loopy
  make init  
  ~~~

*  Using docker image
  ~~~
  docker run --name loopy -it  quay.io/jooholee/loopy:latest /bin/bash
  ~~~

## Create a cluster information script
  ~~~
  cd loopy
  
  cat cluster_info.sh
  CLUSTER_CONSOLE_URL=https://console-openshift-console.XXXX
  CLUSTER_API_URL=https://api.XXX:443
  CLUSTER_ADMIN_ID=admin
  CLUSTER_ADMIN_PW=password
  CLUSTER_TOKEN=sha256~XXX
  CLUSTER_TYPE=ROSA
  ~~~

### Install Kserve with a new manifests
~~~
./loopy playbooks run  odh-stable-install-kserve-serverless-on-existing-cluster -p CUSTOM_KSERVE_MANIFESTS=https://github.com/jooho/kserve/tarball/test-manifest  -p CUSTOM_ODH_MODEL_CONTROLLER_MANIFESTS=https://github.com/jooho/odh-model-controller/tarball/test-manifests -vvv -i ./cluster.sh
~~~

### Deploy sample sklearn model
~~~
./loopy units run test-kserve-sklearn-v2-iris-role -vvv -i ./cluster_info.sh 
~~~

### Checks
1. Verify if runtime Service add serving cert annotation 
~~~
oc get service sklearn-example-isvc-iris-v2-rest -ojsonpath='{.metadata.annotations}'
~~~

2. Verify if Serving Cert Secret is created
~~~
oc get secret sklearn-example-isvc-iris-v2-rest
~~~

3. Verify if Kserve gateway have the Server
~~~
kubectl get gateway kserve-local-gateway -n istio-system -o jsonpath='{.spec.servers[?(@.port.name=="sklearn-example-isvc-iris-v2-rest")]}'
~~~


config/crd/external/gateways.networking.istio.io.yaml
```diff
@@ -0,0 +1,258 @@
+apiVersion: apiextensions.k8s.io/v1
```
spolti, line 1: why is this needed?
Doesn't istio installation provides it already?

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+
+				time.Sleep(secretCheckInterval)
+			} else {
+				log.Error(getSecretErr, "Failed to retrieve the certificate secret for the InferenceService (ISVC)")
+				return getSecretErr
+			}
+		}
+		if err := r.copyServingCertSecretFromIsvcNamespace(ctx, secret); err != nil {
+			log.V(1).Error(err, "Failed to update KServe local gateway in the istio-system namespace")
+
```
spolti, line 100:
```suggestion
```

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+	if !destinationSecretExist {
+		secret := &corev1.Secret{}
+		for attempt := 1; attempt <= maxSecretCheckAttempts; attempt++ {
+			getSecretErr := r.client.Get(ctx, types.NamespacedName{Name: isvc.Name, Namespace: isvc.Namespace}, secret)
+			if getSecretErr == nil {
+				break
+			}
+			if errors.IsNotFound(getSecretErr) {
+				log.V(2).Info("The certificate secret is not created yet. Retrying...", "attempt_number", attempt)
+
```
spolti, line 91:
```suggestion
```

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+		if errors.IsNotFound(err) {
+			destinationSecretExist = false
+		} else {
+			return err
+		}
+	}
+
+	if !destinationSecretExist {
+		secret := &corev1.Secret{}
+		for attempt := 1; attempt <= maxSecretCheckAttempts; attempt++ {
```
spolti, line 84: isn't it safer to retriger the reconcile loop?

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	// Process Delta
+	if err = r.processDelta(ctx, log, desiredResource, existingResource); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *KserveGatewayReconciler) createDesiredResource(isvc *kservev1beta1.InferenceService) (*istioclientv1beta1.Gateway, error) {
```
spolti, line 127: the verb create led us to think that is it being created/deployed in the cluster, maybe `getDesiredResource`?

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+	hostname := fmt.Sprintf("%s.%s.%s", isvc.Name, isvc.Namespace, internalSerivceHostnameDomain)
+
+	desiredGateway := &istioclientv1beta1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kserve-local-gateway",
```
spolti, line 132: constant?

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+			Namespace: isvc.Namespace,
+		},
+		Spec: istiov1beta1.Gateway{
+			Servers: []*istiov1beta1.Server{
+				{
+					Hosts: []string{hostname},
+					Port: &istiov1beta1.Port{
+						Name:     isvc.Name,
+						Number:   8445,
```
spolti, line 141: are these info expected to never change?

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+
+	return desiredGateway, nil
+}
+
+func (r *KserveGatewayReconciler) getExistingResource(ctx context.Context) (*istioclientv1beta1.Gateway, error) {
+	return r.gatewayHandler.Get(ctx, types.NamespacedName{Name: constants.KServeGatewayName, Namespace: constants.IstioNamespace})
+}
+
+func (r *KserveGatewayReconciler) Delete(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+	log.V(1).Info(fmt.Sprintf("Deleting serving cert secret(%s) in the namespace(%s)", isvc.Name, isvc.Namespace))
```
spolti, line 161: this can be debug

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+	if err := r.deleteServingCertSecretInIstioNamespace(ctx, isvc.Name); err != nil {
+		log.V(1).Error(err, "Failed to delete the copied serving cert secret in the namespace")
+		return err
+	}
+
+	log.V(1).Info(fmt.Sprintf("Deleting the Server(%s) from KServe local gateway in the istio-system namespace", isvc.Name))
```
spolti, line 167: here as well

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+	if err := r.removeServerFromGateway(ctx, log, isvc.Name); err != nil {
+		log.V(1).Error(err, "Failed to remove the server from KServe local gateway in the istio-system namespace")
+		return err
+	}
+
+	return nil
+}
+
+func (r *KserveGatewayReconciler) Cleanup(ctx context.Context, log logr.Logger, isvcName string) error {
+	// NOOP - Resources should not be deleted until the kserve component is uninstalled.
```
spolti, line 177: I didn't see the resource owner being set, how it is going to be deleted?
And who owns these resources?

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+	return nil
+}
+
+func (r *KserveGatewayReconciler) processDelta(ctx context.Context, log logr.Logger, desiredGateway *istioclientv1beta1.Gateway, existingGateway *istioclientv1beta1.Gateway) (err error) {
+	comparator := comparators.GetGatewayComparator()
+	delta := r.deltaProcessor.ComputeDelta(comparator, desiredGateway, existingGateway)
+
+	if !delta.HasChanges() {
+		// Note: This code will not be executed.
```
spolti, line 186: expected?
If user deletes or change it by accident, what happens?

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+	if err := r.gatewayHandler.Update(ctx, gateway); err != nil {
+		log.V(1).Error(err, "Failed to update KServe local gateway in istio-system namespace")
+		return err
+	}
+
+	return nil
+}
+
+func (r *KserveGatewayReconciler) copyServingCertSecretFromIsvcNamespace(ctx context.Context, sourceSecret *corev1.Secret) error {
+	destinationSecret := &corev1.Secret{
```
spolti, line 238: what does this function do?

spolti: nvm

controllers/utils/init.go
```diff
@@ -34,6 +35,7 @@ func RegisterSchemes(s *runtime.Scheme) {
 	utilruntime.Must(maistrav1.SchemeBuilder.AddToScheme(s))
 	utilruntime.Must(knservingv1.AddToScheme(s))
 	utilruntime.Must(authorinov1beta2.SchemeBuilder.AddToScheme(s))
+	utilruntime.Must(istioclientv1beta1.AddToScheme(s))
```
spolti, line 38: wouldn't this break setups that does not have istio?

controllers/resources/service.go
```diff
@@ -39,14 +43,38 @@ func NewServiceHandler(client client.Client) ServiceHandler {
...
 		return nil, nil
 	} else if err != nil {
 		return nil, err
 	}
 	log.V(1).Info("Successfully fetch deployed Service")
-	return route, nil
+	return svc, nil
+}
+
+func (r *serviceHandler) FetchWithRetryAndDelay(ctx context.Context, log logr.Logger, key types.NamespacedName, retryDelay time.Duration, retry int) (*v1.Service, error) {
```
spolti, line 58: FetchServiceWith...

controllers/resources/service.go
```diff
@@ -39,14 +43,38 @@ func NewServiceHandler(client client.Client) ServiceHandler {
...
+				continue
+			} else {
+				return nil, err
+			}
+		}
+		break
+	}
+
+	if svc == nil {
+		log.Info(fmt.Sprintf("Failed to fetch the Service(%s) after retries(%d)", key.Name, retry))
```
spolti, line 77: at this point, the service was not found instead of a failure.

controllers/reconcilers/kserve_isvc_service_cert_reconciler.go
```diff
@@ -0,0 +1,132 @@
...
+
+func NewKserveIsvcServiceReconciler(client client.Client) *KserveIsvcServiceReconciler {
+	return &KserveIsvcServiceReconciler{
+		client:         client,
+		serviceHandler: resources.NewServiceHandler(client),
+	}
+}
+
+func (r *KserveIsvcServiceReconciler) Delete(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+	// NOOP - Resources are deleted along with the deletion of InferenceServices
```
spolti, line 47: same here, who is setting the ownership?

config/crd/external/gateways.networking.istio.io.yaml
Jooho: For the test, this is needed. The new reconciler updated KServe Local Gateway in the istio-system.

controllers/comparators/gateway_comparator.go
```diff
@@ -0,0 +1,44 @@
...
+				exists = true
+				break
+			}
+		}
+
+		return exists
+	}
+}
+
+// Function to compare two Server objects
```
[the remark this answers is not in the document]
Jooho: The port name is set to the ISVC name by the new reconciliation, so checking the port name means whether the server object has been updated for ISVC or not. So I wanted to deliver the real meaning of this comparison.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: It is kind of standard format that @vaibhavjainwiz defined for odh-model-controller. So I followed the way and I think the name would be ok.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: good catch. updated

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: yes, it is managed by odh-model-controller and if it is changed by something else, local communication will fail. So I don't think this information won't be changed.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: I referenced the other reconcile and they were using info so I followed it.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: same here

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: this is for updating Gateway contents, not Gateway itself. The Gateway will be created by opendatahub operator so I believe the operator will set the ownResource information.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: Right. If user deletes the KServe local gateway, it doesn't do anything. The Gateway is managed by opendatahub.

When opendatahub creates the KServe local gateway again, it also does not do anything with the IsAdded().

But if the Gateway is created again, it will update the gateway using IsUpdated()

controllers/resources/service.go
Jooho: updated.

controllers/resources/service.go
Jooho: Right I changed it to Error and return err

controllers/utils/init.go
Jooho: No I don't think so because another istio scheme already added there. This is the second istio scheme

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: That's a good question. There are some time gaps between creating a Secret for isvc and calling this gateway Reconciler. So I think it would be better to retry 5 times (but I expect only 1~2 times will happen) to wait for the secret creation.  @israel-hdez what do you think?

controllers/reconcilers/kserve_isvc_service_cert_reconciler.go
Jooho: This is managed by kserve.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+				log.V(2).Info("The certificate secret is not created yet. Retrying...", "attempt_number", attempt)
+
+				time.Sleep(secretCheckInterval)
+			} else {
+				log.Error(getSecretErr, "Failed to retrieve the certificate secret for the InferenceService (ISVC)")
+				return getSecretErr
+			}
+		}
+		if err := r.copyServingCertSecretFromIsvcNamespace(ctx, secret); err != nil {
+			log.V(1).Error(err, "Failed to update KServe local gateway in the istio-system namespace")
```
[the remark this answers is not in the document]
Jooho: updated

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
spolti: can we confirm it?

controllers/utils/init.go
spolti: can we do a test without service mesh installed?
cc @lugi0

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
spolti: okay, but, on normal operation the fewer info messages we have the better.
IMHO, this is one example of informational messages that could be set to debug only.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
spolti: just my 2 cents, we could change that :D
it is actually not creating anything :)

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
spolti: when calling the reconciler loop, IIRC we can set the wait time to run again, we could have a counter, if that reaches, we give up.

controllers/comparators/gateway_comparator.go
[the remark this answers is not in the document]
spolti: my suggestions is more like to use go lang pattern for method comments, it should start with the method name.

config/crd/external/gateways.networking.istio.io.yaml
spolti: so, it could be moved elsewhere?

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -95,6 +103,89 @@ var _ = Describe("The Openshift Kserve model controller", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+
+			url, err := apis.ParseURL("https://example-onnx-mnist-default.test.com")
+			Expect(err).NotTo(HaveOccurred())
+			deployedInferenceService.Status.URL = url
+
+			err = cli.Status().Update(ctx, deployedInferenceService)
+			Expect(err).NotTo(HaveOccurred())
+
+			isvcService, err := waitForService(cli, testNs, inferenceService.Name, 5, 2*time.Second)
```
israel-hdez, line 133: Better to use `Eventually` and let the test framework to re-try for yourself.

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -95,6 +103,89 @@ var _ = Describe("The Openshift Kserve model controller", func() {
...
 
+		It("With a new Kserve InferenceService, serving cert annotation should be added to the runtime Service object.", func() {
+			// Create a new InferenceService
+			inferenceService := &kservev1beta1.InferenceService{}
+			err := convertToStructuredResource(KserveInferenceServicePath1, inferenceService)
+			Expect(err).NotTo(HaveOccurred())
+			inferenceService.SetNamespace(testNs)
+			Expect(cli.Create(ctx, inferenceService)).Should(Succeed())
+
+			// Stub: Create a Runtime Service, which must be created by the OpenDataHub operator.
```
israel-hdez, line 114: Comment seems confusing... This is not a Runtime Service, but a KServe service.

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -95,6 +103,89 @@ var _ = Describe("The Openshift Kserve model controller", func() {
...
+			// Stub: Create a certificate Secret, which must be created by the openshift service-ca operator.
+			secret := &corev1.Secret{}
+			err = convertToStructuredResource(testIsvcSvcSecretPath, secret)
+			Expect(err).NotTo(HaveOccurred())
+			secret.SetNamespace(inferenceService.Namespace)
+			Expect(cli.Create(ctx, secret)).Should(Succeed())
+
+			// Verify that the certificate secret is found in the same namespace where the InferenceService is created.
+			_, err = waitForSecret(cli, inferenceService.Namespace, inferenceService.Name, 5, 1*time.Second)
+			Expect(err).NotTo(HaveOccurred())
```
israel-hdez, line 175: Do you need this?
Looks like the test itself creates this... Not sure why the check is required.

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -95,6 +103,89 @@ var _ = Describe("The Openshift Kserve model controller", func() {
...
+			// Verify that the certificate secret is created in the istio-system namespace.
+			_, err = waitForSecret(cli, constants.IstioNamespace, inferenceService.Name, 5, 5*time.Second)
+			Expect(err).NotTo(HaveOccurred())
+
+			gateway, err := waitForUpdatedGatewayCompletion(cli, "add", constants.IstioNamespace, constants.KServeGatewayName, inferenceService.Name, 5, 2*time.Second)
+			Expect(err).NotTo(HaveOccurred())
+
+			// Ensure that the server is successfully added to the KServe local gateway within the istio-system namespace.
+			targetServerExist := isServerExistFromGateway(gateway, inferenceService.Name)
+			Expect(targetServerExist).Should(BeTrue())
```
israel-hdez, line 186: Again... it should be better to use `Eventually` for these checks.

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -95,6 +103,89 @@ var _ = Describe("The Openshift Kserve model controller", func() {
...
+		})
+
```
israel-hdez, line 188: Certificate rotation is not covered in the units.

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -121,8 +212,99 @@ var _ = Describe("The Openshift Kserve model controller", func() {
...
+			gateway, err := waitForUpdatedGatewayCompletion(cli, "delete", constants.IstioNamespace, constants.KServeGatewayName, isvcName, 5, 2*time.Second)
+			Expect(err).NotTo(HaveOccurred())
+
+			// Ensure that the server is successfully removed from the KServe local gateway within the istio-system namespace.
+			targetServerExist := isServerExistFromGateway(gateway, isvcName)
+			Expect(targetServerExist).Should(BeFalse())
+
+			// Ensure that the created Secret is successfully deleted within the istio-system namespace.
+			secret, err := waitForSecret(cli, constants.IstioNamespace, isvcName, 5, 2*time.Second)
```
israel-hdez, line 302: Use `Eventually`

controllers/constants/constants.go
```diff
@@ -49,8 +49,14 @@ const (
 	KServeCACertConfigMapName  = "odh-kserve-custom-ca-bundle"
 	ODHGlobalCertConfigMapName = "odh-trusted-ca-bundle"
 	ODHCustomCACertFileName    = "odh-ca-bundle.crt"
+	KServeGatewayName          = "kserve-local-gateway"
```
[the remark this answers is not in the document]
Jooho: "I don't think we should use 'envvar' because it is specifically for the KServe local gateway. If the name changes, then we can update this, but I don't think there is a possibility of it changing.

main.go
```diff
@@ -59,6 +59,8 @@ func init() { //nolint:gochecknoinits //reason this way we ensure schemes are al
 // +kubebuilder:rbac:groups=serving.kserve.io,resources=servingruntimes/finalizers,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=networking.istio.io,resources=virtualservices,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=networking.istio.io,resources=virtualservices/finalizers,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=networking.istio.io,resources=gateways,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=networking.istio.io,resources=gateways/finalizers,verbs=get;list;watch;create;update;patch;delete
```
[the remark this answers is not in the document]
Jooho: good catch!. yes I think we don't need this because it should not delete the Gateway.
I will update it to enable more fine-grained control.

config/rbac/role.yaml
```diff
@@ -122,6 +122,30 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - networking.istio.io
+  resources:
+  - gateways
+  verbs:
+  - create
+  - delete
```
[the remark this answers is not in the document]
Jooho: As mentioned earlier, I will update role parts

controllers/reconcilers/kserve_isvc_service_cert_reconciler.go
```diff
@@ -0,0 +1,132 @@
...
+			Annotations: map[string]string{
+				constants.ServingCertAnnotationKey: isvc.Name,
+			},
+		},
+	}
+	return service, nil
+}
+
+func (r *KserveIsvcServiceReconciler) getExistingResource(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) (*v1.Service, error) {
+	return r.serviceHandler.FetchServiceWithRetryAndDelay(ctx, log, types.NamespacedName{Name: isvc.Name, Namespace: isvc.Namespace}, 1*time.Second, 10)
```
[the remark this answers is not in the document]
Jooho: Thanks for the recommendation. I will update the code.

controllers/reconcilers/kserve_isvc_service_cert_reconciler.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func (r *KserveIsvcServiceReconciler) processDelta(ctx context.Context, log logr.Logger, desiredService *v1.Service, existingService *v1.Service) (err error) {
+	if isUpdated(desiredService, existingService, log) {
+		log.V(1).Info("Delta found", "update", existingService.GetName())
+		service := existingService.DeepCopy()
+		if service.Annotations == nil {
+			service.Annotations = make(map[string]string)
+		}
+		service.Annotations = desiredService.Annotations
```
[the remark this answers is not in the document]
Jooho: Yes. that's a good idea! I will use webhook.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+	log.V(1).Info("Reconciling KServe local gateway for Kserve InferenceService")
+
+	// return if URL is not set
+	if isvc.Status.URL == nil {
+		log.V(1).Info("Waiting for the URL as the Inference Service is not ready yet")
+		return nil
+	}
+
+	destinationSecretExist := true
+	if _, err := r.secretHandler.Get(ctx, types.NamespacedName{Name: isvc.Name, Namespace: constants.IstioNamespace}); err != nil {
```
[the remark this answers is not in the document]
Jooho: ok I updated.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+
+	// Process Delta
+	if err = r.processDelta(ctx, log, desiredResource, existingResource); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *KserveGatewayReconciler) createDesiredResource(isvc *kservev1beta1.InferenceService) (*istioclientv1beta1.Gateway, error) {
+	hostname := fmt.Sprintf("%s.%s.%s", isvc.Name, isvc.Namespace, internalSerivceHostnameDomain)
```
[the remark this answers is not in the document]
Jooho: updated

controllers/kserve_inferenceservice_controller_test.go
Jooho: I intended to represent a Service created for each ISVC as "Runtime Service." It seems clearer to rename it to "Kserve Service," so I will proceed with that change.

controllers/kserve_inferenceservice_controller_test.go
Jooho: Good catch. This is not needed. I will remove it

controllers/comparators/gateway_comparator.go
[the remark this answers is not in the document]
Jooho: ah got it. I updated.

controllers/kserve_inferenceservice_controller_test.go
Jooho: is it using async way? this test needs to proceed sync way. like if there is not the Service, it would not move forward to next line. `Eventually` will support this?

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	if !destinationSecretExist {
+		secret := &corev1.Secret{}
+		for attempt := 1; attempt <= maxSecretCheckAttempts; attempt++ {
+			getSecretErr := r.client.Get(ctx, types.NamespacedName{Name: isvc.Name, Namespace: isvc.Namespace}, secret)
+			if getSecretErr == nil {
+				break
+			}
+			if errors.IsNotFound(getSecretErr) {
```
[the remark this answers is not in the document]
Jooho: The 218 PR was merged after this PR was created. So it might be different now but the issue happened to me.
I am going to check this again

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: yes. I agreed.
updated.

controllers/constants/constants.go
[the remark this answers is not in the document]
israel-hdez: Well... if the users do the manual install, there is the chance they can use any name.
But, this is just a recommendation... definitely not a blocking comment.

controllers/kserve_inferenceservice_controller_test.go
israel-hdez: Yes, `Eventually` blocks the program flow until it times out, or its assertion succeeds.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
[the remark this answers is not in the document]
israel-hdez: Well... after your last push, the units are indeed failing with _not found_ errors...

config/crd/external/gateways.networking.istio.io.yaml
spolti: okay, is this packaged with the bundle image?

Jooho: nope

controllers/constants/constants.go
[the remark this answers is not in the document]
Jooho: let's leave it for now.

controllers/kserve_inferenceservice_controller_test.go
Jooho: added a new case for cert rotation.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: yes I confirmed it has ownReference.
~~~
    ownerReferences:
    - apiVersion: features.opendatahub.io/v1
      kind: FeatureTracker
      name: opendatahub-serverless-serving-gateways
      uid: 2009687a-17a5-4b4b-a7ac-78a1708c876d
~~~

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: updated

controllers/utils/init.go
Jooho: for unit test, yes we can test it without service mesh installed. The CRD is for unit test

controllers/reconcilers/kserve_isvc_service_cert_reconciler.go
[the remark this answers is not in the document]
Jooho: updated

controllers/reconcilers/kserve_isvc_service_cert_reconciler.go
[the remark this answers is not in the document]
Jooho: added mutating webhook

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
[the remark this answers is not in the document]
Jooho: yes it should have some errors at that time but now I pushed all changes so it should not have any errors now.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	return nil
+}
+
+func (r *KserveGatewayReconciler) copyServingCertSecretFromIsvcNamespace(ctx context.Context, sourceSecret *corev1.Secret, isvc *kservev1beta1.InferenceService) error {
+
+	destinationSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      sourceSecret.Name,
```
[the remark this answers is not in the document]
Jooho: updated.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      constants.KServeGatewayName,
+			Namespace: isvc.Namespace,
+		},
+		Spec: istiov1beta1.Gateway{
+			Servers: []*istiov1beta1.Server{
+				{
+					Hosts: []string{hostname},
+					Port: &istiov1beta1.Port{
+						Name:     isvc.Name,
```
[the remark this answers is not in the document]
Jooho: updated

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+
+func (r *KserveGatewayReconciler) Delete(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+	log.V(1).Info(fmt.Sprintf("Deleting serving cert secret(%s) in the namespace(%s)", isvc.Name, isvc.Namespace))
+	if err := r.deleteServingCertSecretInIstioNamespace(ctx, isvc.Name); err != nil {
+		log.V(1).Error(err, "Failed to delete the copied serving cert secret in the namespace")
+		return err
+	}
+
+	log.V(1).Info(fmt.Sprintf("Deleting the Server(%s) from KServe local gateway in the istio-system namespace", isvc.Name))
+	if err := r.removeServerFromGateway(ctx, log, isvc.Name); err != nil {
```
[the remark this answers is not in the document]
Jooho: updated

controllers/kserve_inferenceservice_controller_test.go
Jooho: updated

controllers/kserve_inferenceservice_controller_test.go
Jooho: removed

controllers/kserve_inferenceservice_controller_test.go
Jooho: updated

controllers/kserve_inferenceservice_controller_test.go
Jooho: updated

VedantMahabaleshwarkar (changes requested):
I think the MutatingWebhook needs patches similar to what we have for the ValidatingWebhook. 
[field patch](https://github.com/opendatahub-io/odh-model-controller/blob/main/config/webhook/field_patch.yaml) and [webhook patch](https://github.com/opendatahub-io/odh-model-controller/blob/main/config/webhook/webhook_patch.yaml)

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,261 @@
...
+	portName = fmt.Sprintf("%s-%s", "https", isvc.Name)
+
+	// return if Address.URL is not set
+	if isvc.Status.Address != nil && isvc.Status.Address.URL == nil {
+		log.V(1).Info("Waiting for the URL as the InferenceService is not ready yet")
+		return nil
+	}
+
+	// return if serving cert secret in the source namespace is not created
+	srcCertSecret, err := r.secretHandler.Get(ctx, types.NamespacedName{Name: isvc.Name, Namespace: isvc.Namespace})
```
israel-hdez, line 75: I'm still unsure that this is really going to work. Units may be passing, because `suite_test.go` is not configuring for the client cache, like in `main.go`. But in the actual deployment, it should be always returning _not found_.

config/webhook/manifests.yaml
```diff
@@ -1,5 +1,30 @@
 ---
 apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: mutating-webhook-configuration
+webhooks:
```
israel-hdez, line 6: Additionally to [Vedant's comment](https://github.com/opendatahub-io/odh-model-controller/pull/221#pullrequestreview-2132997987), if possible, configure some selector/filter on the webhook, to prevent it being called on every Svc on the cluster.

spolti: +1

controllers/inferenceservice_controller.go
```diff
@@ -147,6 +148,23 @@ func (r *OpenshiftInferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager)
 					})
 				}
 				return reconcileRequests
+			})).
+		Watches(&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, o client.Object) []reconcile.Request {
```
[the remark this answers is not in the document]
spolti: agreed, it could lead to the same issue as @vaibhavjainwiz is dealing with on odh-model-controller

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/kserve/kserve v0.12.1
 	github.com/kuadrant/authorino v0.15.0
+	github.com/moby/moby v27.0.0+incompatible
```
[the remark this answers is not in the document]
Jooho: interesting. why this libs added? ^^;; I cleaned up necessary lib

controllers/inferenceservice_controller.go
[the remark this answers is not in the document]
Jooho: oh right! I add filter to work with a secret that has `ownerReference.Kind == "InferenceService"` because the secret does not have any label.

controllers/webhook/kserve_service_mutator.go
```diff
@@ -0,0 +1,61 @@
...
+	log := logf.FromContext(ctx).WithName("KserviceServiceMutateWebhook")
+
+	svc, ok := obj.(*corev1.Service)
+	if !ok {
+		return fmt.Errorf("unexpected object type")
+	}
+
+	// Retrieve the InferenceService directly by name
+	var isvc kservev1beta1.InferenceService
+	err := m.client.Get(ctx, client.ObjectKey{Name: svc.Name, Namespace: svc.Namespace}, &isvc)
```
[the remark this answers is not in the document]
Jooho: sure. updated

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,261 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	istiov1beta1 "istio.io/api/networking/v1beta1"
+	istioclientv1beta1 "istio.io/client-go/pkg/apis/networking/v1beta1"
+)
+
+var _ SubResourceReconciler = (*KserveGatewayReconciler)(nil)
+var meshNamespace string
+var destSecretName string
+var portName string
```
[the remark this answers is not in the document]
Jooho: make sense. I changed `destSecretName`,`portName` but I believe meshNamespace should be ok because it is always the same.

config/webhook/manifests.yaml
Jooho: at the moment, the service does not have any labels so it is not possible to filter. In order to do this, I need to send another pr to upstream kserve. I will follow up this with another ticket: [RHOAIENG-8978](https://issues.redhat.com/browse/RHOAIENG-8978)

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+			Servers: []*istiov1beta1.Server{
+				{
+					Hosts: []string{hostname},
+					Port: &istiov1beta1.Port{
+						Name:     isvc.Name,
+						Number:   8445,
+						Protocol: "HTTPS",
+					},
+					Tls: &istiov1beta1.ServerTLSSettings{
+						CredentialName: isvc.Name,
```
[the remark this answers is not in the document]
Jooho: updated

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: The secretHandler just get secret with name and namespace. This is not working with cache stuff.
So I don't think it has something to do with cache.

Jooho:
@israel-hdez FYI, I removed mutating webhook and re-add  isvc_service_cert_reconciler because it turned out kserve controller rollback the annotation.

config/webhook/manifests.yaml
Jooho: Remove mutatingwebhook so it is not needed

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,279 @@
...
+		Data: sourceSecret.Data,
+		Type: sourceSecret.Type,
+	}
+
+	// Remove old secret if src secret is updated
+	if preDestSecret != nil {
+		if err := r.client.Delete(ctx, preDestSecret); err != nil {
+			return err
+		}
+	}
```
israel-hdez, line 252: @bartoszmajsak @rcernich This will delete a no longer useful TLS secret and create a new update one with the same name (e.g. at rotation).

Do you know if the ingress gateway will reload it without any changes to the `Gateway`? Or does it need to have a new name, and the `Gateway` must be updated to effectively reload it?

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func (r *KserveGatewayReconciler) copyServingCertSecretFromIsvcNamespace(ctx context.Context, sourceSecret *corev1.Secret, preDestSecret *corev1.Secret) error {
+	destinationSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-%s", sourceSecret.Name, sourceSecret.Namespace),
+			Namespace: meshNamespace,
+			Labels: map[string]string{
+				"opendatahub.io/managed": "true",
+			},
```
israel-hdez, line 241: There is an ADR covering recommended labels, although not merged yet. Additionally to this one, not sure if you would like to follow the ADR and add the other suggested labels: https://github.com/opendatahub-io/architecture-decision-records/pull/26/files#diff-87d90e8a91b0f96b79615d8b6d9887902a66d2a4ab38415757fc7945b94068a0R94.

Example here: https://github.com/opendatahub-io/odh-model-controller/blob/86cc36042347e1817f7263aacc1ceb818468e85b/controllers/reconcilers/kserve_istio_smm_reconciler.go#L73-L77.

This is just a suggestion. Not a blocker...

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -0,0 +1,279 @@
...
+				return err
+			}
+		}
+		return err
+	}
+
+	// Recreate copied secrt when src secret is updated
+	if !reflect.DeepEqual(srcCertSecret.Data, copiedCertSecret.Data) {
+		log.V(1).Info(fmt.Sprintf("Recreating for serving certificate Secret(%s) in %s namespace", copiedCertSecret.Name, meshNamespace))
+		if err := r.copyServingCertSecretFromIsvcNamespace(ctx, srcCertSecret, copiedCertSecret); err != nil {
```
israel-hdez, line 102: Without the `Watch` on the controller, I'm failing to understand how this will catch the certificate rotation, because looks like none of the watched & owned resources will be updated and a reconcile won't be triggered.

I think you should add [these lines](https://github.com/opendatahub-io/odh-model-controller/blob/86cc36042347e1817f7263aacc1ceb818468e85b/main.go#L127-L145), [over here](https://github.com/opendatahub-io/odh-model-controller/blob/86cc36042347e1817f7263aacc1ceb818468e85b/controllers/suite_test.go#L146) so that units can catch this (in case there is an issue).

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -94,6 +103,119 @@ var _ = Describe("The Openshift Kserve model controller", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+
+			// Verify that the certificate secret is created in the istio-system namespace.
+			Eventually(func() error {
+				secret := &corev1.Secret{}
+				err := cli.Get(ctx, client.ObjectKey{Namespace: constants.IstioNamespace, Name: fmt.Sprintf("%s-%s", inferenceService.Name, inferenceService.Namespace)}, secret)
+				if err != nil {
+					return err
+				}
+				return nil
```
israel-hdez, line 202:
```suggestion
				return cli.Get(ctx, client.ObjectKey{Namespace: constants.IstioNamespace, Name: fmt.Sprintf("%s-%s", inferenceService.Name, inferenceService.Namespace)}, secret)
```

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -94,6 +103,119 @@ var _ = Describe("The Openshift Kserve model controller", func() {
...
+			}, timeout, interval).Should(Succeed())
+
+			// Verify that the gateway is updated in the istio-system namespace.
+			var gateway *istioclientv1beta1.Gateway
+			Eventually(func() error {
+				gateway, err = waitForUpdatedGatewayCompletion(cli, "add", constants.IstioNamespace, constants.KServeGatewayName, inferenceService.Name)
+				if err != nil {
+					return err
+				}
+				return nil
```
israel-hdez, line 212:
```suggestion
				return err
```

There are several ones like this. I won't point all them :-)

israel-hdez (commented):
Adding some additional comment. It may be easy to just reply so that I understand better.

If you, by chance, would push some code changes, I saw some code commented in `_test.go` files. It would be good if you can clean-up them.

controllers/reconcilers/kserve_isvc_service_cert_reconciler.go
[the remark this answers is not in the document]
Jooho: updated

controllers/webhook/kserve_service_mutator.go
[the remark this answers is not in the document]
Jooho: removed mutating webhook so I closed this conversation.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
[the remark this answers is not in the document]
Jooho: updated.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: I used APIReader instead of clientset. It is easier to use in this case because manager has this object.

controllers/reconcilers/kserve_isvc_service_cert_reconciler.go
```diff
@@ -0,0 +1,132 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+func isUpdated(desiredService *v1.Service, existingService *v1.Service, log logr.Logger) bool {
+	if existingService == nil {
+		log.Info("The service for the InferenceService has not been created yet")
+		return false
```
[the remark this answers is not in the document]
Jooho: The Secret that is created by serving cert does not have the lave `opendatahub.io` so it is not cached. For this reason, it is useless to add watch here.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: good point!. i added the labels.

controllers/reconcilers/kserve_isvc_gateway_reconciler.go
Jooho: As I commented from [here](https://github.com/opendatahub-io/odh-model-controller/pull/221#discussion_r1656422292), watch can not catch the Secret because it does not have the specific level. I need to think more to solve this rotation part but at the moment, Watch is not the right approach.

controllers/kserve_inferenceservice_controller_test.go
Jooho: updated

controllers/kserve_inferenceservice_controller_test.go
Jooho: I updated all similar cases.

Jooho:
This PR was somehow merged accidentally so I reverted it.

controllers/inferenceservice_controller.go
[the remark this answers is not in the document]
Jooho: The secret that we need to watch does not have the specific label so it is not cached. For this reason, I removed this Watch part.

PR #226, opened 2024-06-24
Fix: update params.env with caikit image

christinaexyou:
Updates `config/base/params.env` by adding the Caikit Standalone image.


heyselbi:
@christinaexyou it should be image in quay.io/opendatahub

heyselbi:
Please update it to
quay.io/opendatahub/caikit-nlp:stable-994ec60 
@christinaexyou

PR #220, opened 2024-06-11
Kserve Dashboard Reconciler

ashley-o0o:
## Description
Add logic for configmap creation for Kserve metrics in ODH Model Controller
Closes: [RHOAIENG-3956](https://issues.redhat.com/browse/RHOAIENG-3956)

## How Has This Been Tested?
- install odh
- install the following DSC 
- ```spec:
  components:
    codeflare:
      managementState: Removed
    kserve:
      devFlags:
        manifests:
          - contextDir: config
            sourcePath: overlays/odh
            uri: 'https://github.com/opendatahub-io/kserve/tarball/master'
          - contextDir: config
            sourcePath: ''
            uri: 'https://github.com/ashley-o0o/odh-model-controller/tarball/kserveReconciler'
      managementState: Managed
      serving:
        ingressGateway:
          certificate:
            type: SelfSigned
        managementState: Managed
        name: knative-serving
    modelregistry:
      managementState: Removed
    trustyai:
      managementState: Removed
    ray:
      managementState: Removed
    kueue:
      managementState: Removed
    workbenches:
      managementState: Removed
    dashboard:
      managementState: Managed
    modelmeshserving:
      managementState: Removed
    datasciencepipelines:
      managementState: Removed
    trainingoperator:
      managementState: Removed
- create a data science project
- deploy a sample model for each supported ServingRuntime 
- ensure <modelname>-metrics-dashboard configmap exists, and contents are correct 
- ensure configmap is deleted when model is deleted. 
- ensure that the configmap cannot be changed by the user (the reconciler should revert changes) 

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #240, opened 2024-07-17
Remove configmap filter and watch

VedantMahabaleshwarkar:
## Description

Fixes https://issues.redhat.com/browse/RHOAIENG-9744  and https://issues.redhat.com/browse/RHOAIENG-8766 by removing the filter and watch on configmaps. 
Additionally the following changes have been made : 
- Removed the need to have json files that need to be copied to the image by replacing them with constants that can be used in the same way 
- Fixed request count queries to return an integer by adding a `round()`
- Fixed cpu and memory % queries and have made them the same as existing queries for modelmesh for the same graphs that can be found [here](https://github.com/opendatahub-io/odh-dashboard/blob/1c5b7917402a88f5b187cb2488abbb3c612d19d9/frontend/src/pages/modelServing/screens/metrics/utils.tsx#L30-L56) 

## How Has This Been Tested?

DSC : 
```
spec:
  components:
    codeflare:
      managementState: Managed
    kserve:
      devFlags:
        manifests:
          - contextDir: config
            sourcePath: overlays/odh
            uri: 'https://github.com/opendatahub-io/kserve/tarball/master'
          - contextDir: config
            sourcePath: ''
            uri: 'https://github.com/VedantMahabaleshwarkar/odh-model-controller/tarball/devflags'
      managementState: Managed
      serving:
        ingressGateway:
          certificate:
            type: SelfSigned
        managementState: Managed
        name: knative-serving
    modelregistry:
      managementState: Removed
    trustyai:
      managementState: Managed
    ray:
      managementState: Managed
    kueue:
      managementState: Managed
    workbenches:
      managementState: Managed
    dashboard:
      managementState: Managed
    modelmeshserving:
      managementState: Managed
    datasciencepipelines:
      managementState: Managed
    trainingoperator:
      managementState: Removed
```

Rest of the testing instructions are the same as #220 

## Merge criteria:


VedantMahabaleshwarkar:
@alexcreasy PTAL regarding the query fixes. The PR description has a DSC that can be used to deploy the changes for testing.

controllers/kserve_customcacert_controller_test.go
```diff
@@ -78,21 +77,6 @@ var _ = Describe("KServe Custom CA Cert ConfigMap controller", func() {
 	})
 })
 
-func waitForConfigMap(cli client.Client, namespace, configmapName string, timeout time.Duration) (*corev1.ConfigMap, error) {
```
terrytangyuan, line 81: What was the reason to move this to another file?

VedantMahabaleshwarkar: earlier it was only used in this file but now it is a common function used in multiple places so I moved it to the test suite file

terrytangyuan (commented):
LGTM

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -184,13 +143,9 @@ func (r *KserveMetricsDashboardReconciler) createDesiredResource(ctx context.Con
 		},
```
Jooho, line 143: (optional) looks like this variable used several places 
```
isvc.Name + "-metrics-dashboard",
```
So I think it would be a good idea to use a variable like the following to make the code more readable and easier to maintain.to maintain.
```
metricsCMSuffix := "-metrics-dashboard"
metricsCMName := isvc.Name + metricsCMSuffix

configMap := &corev1.ConfigMap{
			Name:      metricsCMName,
			Namespace: isvc.Namespace,
		},
```

controllers/suite_test.go
```diff
@@ -291,3 +293,25 @@ func createTestNamespaceName() string {
...
+		if err == nil {
+			return configMap, nil
+		}
+		if !apierrs.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to get configmap %s/%s: %v", namespace, configMapName, err)
+		}
+
+		if try < maxTries {
+			time.Sleep(1 * time.Second)
+			return nil, err
```
Jooho, line 313: This is something wrong. With this, it does not retry.

Makefile
```diff
@@ -37,7 +37,7 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	# Any customization needed, apply to the webhook_patch.yaml file
-	$(CONTROLLER_GEN) rbac:roleName=odh-model-controller-role,headerFile="hack/manifests_boilerplate.yaml.txt" crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) rbac:roleName=odh-model-controller-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```
Jooho, line 40: I know the txt file does not have anything but it does not harm even if it is specified here. Is there any specific reason you remove this?

Containerfile
```diff
@@ -27,10 +23,6 @@ RUN CGO_ENABLED=0 GOOS=linux go build -a -o manager main.go
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
 WORKDIR /
 COPY --from=builder /workspace/manager .
-COPY --from=builder /workspace/metrics_dashboards/ovms-metrics.json .
```
Jooho, line 30: this is not for this PR but just for your information, please remove this from midstream as well

controllers/reconcilers/kserve_metrics_dashboard_reconciler.go
```diff
@@ -112,47 +106,16 @@ func (r *KserveMetricsDashboardReconciler) createDesiredResource(ctx context.Con
 	findImageName := re.FindStringSubmatch(servingRuntimeImage)
 	servingRuntime := findImageName[1]
 
-	var data []byte
+	var data string
 	switch servingRuntime {
```
Jooho, line 110: (optional)  This is not a blocker. just for recommendation.

You changed the json to constant so the runtime name and also metricsData are String. If so, I think using map would be easier to handle this flow. something like here:
~~~
	runtimeMetricsData := map[string]string{
		constants.OvmsImageName:   constants.OvmsData,
		constants.TgisImageName:   constants.TgisData,
		constants.VllmImageName:   constants.VllmData,
		constants.CaikitImageName: constants.CaikitData,
	}

	data, supported := runtimeData[servingRuntime]
~~~
if no supported runtime comes, it will return the following:
```
data = ""  //(empty string)
supported=false
```

with this, you can use a unified creating ConfigMap process. 
this is just example(not tested)
~~~
	configMap := &corev1.ConfigMap{
			Name:      metricsCMName,
			Namespace: isvc.Namespace,
		},
		Data: map[string]string{
			"supported": supported,
		},
	}
	if supported == "true" {
		configMap.Data["metrics"] = metrics
	}

	if err := ctrl.SetControllerReference(isvc, configMap, r.client.Scheme()); err != nil {
		log.Error(err, "Unable to add OwnerReference to the Metrics Dashboard Configmap")
		return nil, err
	}
~~~

Jooho (changes requested):
most of the comments are recommendations not a blocker but some of them should be updated. please review the comments.

controllers/kserve_inferenceservice_controller_metrics_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	})
+
+	When("deploying a Kserve model", func() {
+		It("if the runtime is supported for metrics, it should create a configmap with prometheus queries", func() {
+			_ = createServingRuntime(testNs, KserveServingRuntimePath1)
+			_ = createInferenceService(testNs, KserveOvmsInferenceServiceName, KserveInferenceServicePath1)
+
+			metricsConfigMap, err := waitForConfigMap(cli, testNs, KserveOvmsInferenceServiceName+ConfigMapNameSuffix, 30, 1*time.Second)
+			fmt.Println(err)
```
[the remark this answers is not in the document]
VedantMahabaleshwarkar: forgot to remove this

controllers/kserve_inferenceservice_controller_metrics_test.go
```diff
@@ -0,0 +1,143 @@
...
+				},
+				Data: map[string]string{
+					"supported": "true",
+					"metrics":   finaldata,
+				},
+			}
+			Expect(compareConfigMap(metricsConfigMap, expectedmetricsConfigMap)).Should(BeTrue())
+		})
+
+		It("if the runtime is not supported for metrics, it should create a configmap with the unsupported config", func() {
```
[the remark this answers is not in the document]
VedantMahabaleshwarkar: The logic to determine "supported" vs "unsupported" is based on the image name in the serving runtime. I changed the image for the sklearn runtime in testdata to `kserve/unsupportedimage:v0.12.1`. 
If/When we add sklearn as an officially supported runtime we will add the correct `SklearnImageName` constant with the correct image name

controllers/suite_test.go
VedantMahabaleshwarkar: ah stupid mistake

Makefile
VedantMahabaleshwarkar: for some reason if I have this in the makefile it doesn't work for me. But anyway, I added it back.

terrytangyuan (approved):
/lgtm

PR #241, opened 2024-07-18
odh-model-controller manifests kustomize v5 upgrade

KillianGolds:
## **Background:**
As part of the task to upgrade all odh-model-controller manifests to be compatible with Kustomize v5, an issue arose with the ClusterRoleBinding definition in the role_binding.yaml file. This issue highlighted the need to simplify the ClusterRoleBinding metadata naming convention by removing the redundant $(mesh-namespace) suffix. This change is necessary because there will only be one odh-model-controller-rolebinding, making the namespace suffix redundant. Additionally, since this change introduces a "new" ClusterRoleBinding from Kubernetes's perspective, the old one will need to be removed.

### **1. Removing the Namespace Suffix from ClusterRoleBinding Name:**
We will update the ClusterRoleBinding definition in the role_binding.yaml file to remove the $(mesh-namespace) variable from the metadata.name field. The new naming convention will simplify the identification and management of the ClusterRoleBinding.

- Old Manifest (role_binding.yaml):
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: odh-model-controller-rolebinding-$(mesh-namespace)
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
subjects:
  - kind: ServiceAccount
    name: odh-model-controller
```

- New Manifest (role_binding.yaml):
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: odh-model-controller-rolebinding
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
subjects:
  - kind: ServiceAccount
    name: odh-model-controller
```

### **2. Deleting the Old ClusterRoleBinding:**
Since the new ClusterRoleBinding is considered a separate resource from the old one by Kubernetes, we need to explicitly remove the old ClusterRoleBinding. This will be handled in the operator repository as part of **THIS** [pull request](https://github.com/opendatahub-io/opendatahub-operator/pull/1140) to ensure the old resource is cleaned up during the upgrade process.

### **3. Core Issue and Resolution:**
The core issue is that the $(mesh-namespace) variable is not being correctly resolved and substituted in the ClusterRoleBinding metadata.name field. Instead of resolving to odh-model-controller-rolebinding-opendatahub, it's resulting in opendatahub. By removing the need for this substitution entirely, we avoid this resolution issue.

### **4. Outcome:**
The updated ClusterRoleBinding manifest will generate as follows:
```
 apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: odh-model-controller-rolebinding
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
subjects:
  - kind: ServiceAccount
    name: odh-model-controller
    namespace: opendatahub
```

## **Summary:**
- The $(mesh-namespace) suffix will be removed from the ClusterRoleBinding name to simplify the manifest.
- The operator repository will handle the deletion of the old ClusterRoleBinding.
- There is no need to update the ConfigMap replacement for the CRB, as the substitution issue is resolved by removing the need for it.

## How Has This Been Tested?

To ensure the correctness of these changes, the following steps were taken:

1. Ran the kustomize build command on the original manifests to produce the current state of the resources.
2. Updated all other manifests to kustomize v5 labels.
3. Made the necessary changes to the role_binding.yaml file to remove the $(mesh-namespace) suffix from the ClusterRoleBinding name.
4. Ran the kustomize build command again on the updated manifests.
5. Compared the output of the kustomize build command from both the original and updated manifests.
6. Verified that the only difference in the output was the metadata.name field of the ClusterRoleBinding, confirming that the rest of the resources remain unchanged.

Gist of all kustomize build pre changes and post changes outputs: https://gist.github.com/KillianGolds/307312b54660b8a0b71c0971600afaa6

## Merge criteria:


VedantMahabaleshwarkar:
looking at the documentation for kustomize, 
I think this might help replicate the existing behavior of appending the namespace to the CRB https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/#delimiter 
probably something like 
```
options:
  delimiter: '-'
  index: 4
``` 
@KillianGolds  

I personally still do not see the value of having the namespace appended to the CRB, but if we do want to replicate existing behavior this can be tried

VedantMahabaleshwarkar:
Once this is ready for review please squash and sign the commits @KillianGolds

PR #245, opened 2024-07-24
add sanity check on regex match results while determining runtime image

VedantMahabaleshwarkar:
## Description

Fixes : https://issues.redhat.com/browse/RHOAIENG-10286 
- Added a sanity check to the regex match array so that the controller does not panic. 
  - If the regex match is not what is expected odh-model-controller will end up creating a configmap for unsupported metrics 
- Modified unit tests to cover this case 
## How Has This Been Tested?

- Unit test coverage
- Try to deploy the SR mentioned in the JIRA and verify odh-model-controller doesn't crash 

## Merge criteria:

### DIFF
--- a/controllers/constants/constants.go
+++ b/controllers/constants/constants.go
@@ -60,6 +60,7 @@ const (
 	TgisImageName                    = "text-generation-inference"
 	VllmImageName                    = "vllm"
 	CaikitImageName                  = "caikit-nlp"
+	ServingRuntimeFallBackImageName  = "unsupported"
 )
 
 // openshift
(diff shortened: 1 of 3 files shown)


Jooho (approved):
At the moment, it is ok for current situation so I will approve this but  the regexp should be improved to cover all possible image URL formats like the following:
```
		"docker.io/library/ubuntu:latest",
		"myregistry.com/mynamespace/myimage@sha256:abcdef123456",
		"myregistry.com/myimage:latest",
		"myimage@sha256:abcdef123456",
		"myimage:latest",
```
and need to add unit tests for this.

/lgtm

PR #250, opened 2024-08-05
add kserve raw reconciler

VedantMahabaleshwarkar:
## Description

https://issues.redhat.com/browse/RHOAIENG-10293 
## How Has This Been Tested?

```
spec:
  components:
    codeflare:
      managementState: Removed
    kserve:
      defaultDeploymentMode: RawDeployment
      devFlags:
        manifests:
          - contextDir: config
            sourcePath: overlays/odh
            uri: 'https://github.com/opendatahub-io/kserve/tarball/master'
          - contextDir: config
            sourcePath: ''
            uri: 'https://github.com/VedantMahabaleshwarkar/odh-model-controller/tarball/devflags'
      managementState: Managed
      serving:
        ingressGateway:
          certificate:
            type: OpenshiftDefaultIngress
        managementState: Removed
        name: knative-serving
    modelregistry:
      managementState: Removed
    trustyai:
      managementState: Removed
    ray:
      managementState: Removed
    kueue:
      managementState: Removed
    workbenches:
      managementState: Removed
    dashboard:
      managementState: Managed
    modelmeshserving:
      managementState: Removed
    datasciencepipelines:
      managementState: Removed
    trainingoperator:
      managementState: Removed
```

## Merge criteria:


controllers/reconcilers/kserve_serverless_inferenceservice_reconciler.go
```diff
@@ -38,8 +38,6 @@ func NewKServeServerlessInferenceServiceReconciler(client client.Client, clientR
 	subResourceReconciler := []SubResourceReconciler{
 		NewKserveServiceMeshMemberReconciler(client),
 		NewKserveRouteReconciler(client),
-		NewKServeMetricsServiceReconciler(client),
-		NewKServeMetricsServiceMonitorReconciler(client),
```
terrytangyuan, line 42: Would this break existing metrics functionality for serverless?

controllers/reconcilers/kserve_raw_route_reconciler.go
```diff
@@ -0,0 +1,195 @@
...
+	if enableAuth, err = strconv.ParseBool(isvc.Labels[constants.LabelEnableAuth]); err != nil {
+		enableAuth = false
+	}
+	createRoute := false
+	if createRoute, err = strconv.ParseBool(isvc.Labels[constants.LabelEnableRoute]); err != nil {
+		createRoute = false
+	}
+
+	if !createRoute {
+		log.Info("InferenceService does not have '" + constants.LabelEnableRoute + "' annotation set to 'True'. Skipping route creation")
```
terrytangyuan, line 102: Either set to false or does not exist

controllers/reconcilers/kserve_raw_route_reconciler.go
```diff
@@ -0,0 +1,195 @@
...
+	return nil
+}
+
+func (r *KserveRawRouteReconciler) Delete(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+	log.V(1).Info("Deleting Kserve inference service generic route")
+	return r.routeHandler.DeleteRoute(ctx, types.NamespacedName{Name: isvc.Name, Namespace: isvc.Namespace})
+}
+
+func (r *KserveRawRouteReconciler) Cleanup(_ context.Context, _ logr.Logger, _ string) error {
+	// NOOP - resources are deleted together with ISVCs
```
terrytangyuan, line 86: Do we need to check existing route that are not created by the controller, e.g. no owner reference?

controllers/reconcilers/kserve_raw_metrics_servicemonitor_reconciler.go
```diff
@@ -22,34 +22,36 @@ import (
...
+type KserveRawMetricsServiceMonitorReconciler struct {
 	NoResourceRemoval
 	client                client.Client
 	serviceMonitorHandler resources.ServiceMonitorHandler
 	deltaProcessor        processors.DeltaProcessor
 }
 
-func NewKServeMetricsServiceMonitorReconciler(client client.Client) *KserveMetricsServiceMonitorReconciler {
-	return &KserveMetricsServiceMonitorReconciler{
+func NewRawKServeMetricsServiceMonitorReconciler(client client.Client) *KserveRawMetricsServiceMonitorReconciler {
```
terrytangyuan, line 42: Maybe keep the same but skips reconciliation for serverless mode?

controllers/reconcilers/kserve_serverless_inferenceservice_reconciler.go
terrytangyuan: I think I got my answer: https://github.com/opendatahub-io/odh-model-controller/pull/250#discussion_r1705967262

controllers/reconcilers/kserve_raw_route_reconciler.go
```diff
@@ -0,0 +1,195 @@
...
+
+func NewKserveRawRouteReconciler(client client.Client) *KserveRawRouteReconciler {
+	return &KserveRawRouteReconciler{
+		client:         client,
+		routeHandler:   resources.NewRouteHandler(client),
+		deltaProcessor: processors.NewDeltaProcessor(),
+	}
+}
+
+func (r *KserveRawRouteReconciler) Reconcile(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
```
israel-hdez, line 58: For exposing, KServe supports creating `Ingress` resources, but we have it [turned off](https://github.com/opendatahub-io/kserve/blob/ae30b971320f45bc98dd362b6b8873d7c1a48873/config/overlays/odh/inferenceservice-config-patch.yaml#L28). [OpenShift can follow-up](https://docs.openshift.com/container-platform/4.16/networking/routes/route-configuration.html#nw-ingress-creating-a-route-via-an-ingress_route-configuration) and create Routes.

Given the push for better supporting the API, rather than doing our own thing, we should try to align better with upstream.

By creating Routes on our side, KServe is no longer aware. So, it won't expose what's the endpoint in `status.url` of the InferenceService. This will just increase complexity on dashboard code. Also, it will be more complex for users because the `status.url` field will be useless for Raw, but will be useful for Serverless.

VedantMahabaleshwarkar: Ah nice, I realized the thing about the `status.url` myself but didn't realize an option existed to do this in a way such that Kserve is aware. I'll incorporate this.

controllers/reconcilers/kserve_raw_metrics_service_reconciler.go
```diff
@@ -71,15 +74,58 @@ func (r *KserveMetricsServiceReconciler) Reconcile(ctx context.Context, log logr
...
+		log.V(1).Info("No Prometheus annotations on ServingRuntime, skipping creation of metrics resources")
+		return nil, nil
+	}
+
+	prometheusPortAnnotationValue := isvcRuntime.Spec.Annotations[constants.PrometheusPortAnnotationKey]
+	prometheusPort, err := strconv.ParseInt(prometheusPortAnnotationValue, 10, 64)
+	if err != nil {
+		return nil, err
+	}
+	metricsService := &v1.Service{
```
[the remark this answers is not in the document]
VedantMahabaleshwarkar: Hmm this puts the onus of enabling metrics for a inferenceservice on the creator of the inferenceservice, rather than the platform. Even if we add logic to inject this in the UI flow, if a user is using yamls then they would have to specifically remember to enable metrics. In ODH/RHOAI, I'm guessing we always want metrics and the user should not have to worry about remembering to enable them. 
But I can be convinced otherwise :) 
cc @danielezonca @terrytangyuan WDYT

VedantMahabaleshwarkar:
@israel-hdez 
>What I remember is that these reconcilers were left for doing clean-up of some resources that we no longer need (see the old createDesiredResource that just returns nil as the "wanted" resource).
By removing these reconcilers, the clean-up is what will no longer happen for version upgrades. I'm not sure if not doing the clean-up will break something, but I'm quite sure we still want the clean-up.

@terrytangyuan 
>Maybe keep the same but skips reconciliation for serverless mode?

The old metrics reconcilers for kserve serverless were present pre 2.8 (EUS release) 
Kserve Raw support will happen after 2.13 at minimum (2.12 is a EUS if i'm not wrong) 
I can keep an "empty reconciliation" loop around, but the old resources should long by the time kserve raw is GA. This was my reasoning for assuming that it's safe to delete the old reconcilers for serverless

controllers/reconcilers/kserve_raw_route_reconciler.go
VedantMahabaleshwarkar: I didn't get this. Could you elaborate?

controllers/reconcilers/kserve_serverless_inferenceservice_reconciler.go
VedantMahabaleshwarkar: The old metrics reconcilers for kserve serverless were present pre 2.8 (EUS release)
Kserve Raw support will happen after 2.13 at minimum (2.12 is a EUS if i'm not wrong)
I can keep an "empty reconciliation" loop around, but the old resources should long by the time kserve raw is GA. This was my reasoning for assuming that it's safe to delete the old reconcilers for serverless

controllers/reconcilers/kserve_raw_inferenceservice_reconciler.go
```diff
@@ -26,16 +27,30 @@ import (
...
 	}
 }
 
-func (r *KserveRawInferenceServiceReconciler) Reconcile(_ context.Context, log logr.Logger, _ *kservev1beta1.InferenceService) error {
-	log.V(1).Info("No Reconciliation to be done for inferenceservice as it is using RawDeployment mode")
-	return nil
+func (r *KserveRawInferenceServiceReconciler) Reconcile(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+	var reconcileErrors *multierror.Error
+	for _, reconciler := range r.subResourceReconcilers {
+		reconcileErrors = multierror.Append(reconcileErrors, reconciler.Reconcile(ctx, log, isvc))
```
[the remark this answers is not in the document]
VedantMahabaleshwarkar: Yep those changes are made already :D

controllers/reconcilers/kserve_serverless_inferenceservice_reconciler.go
VedantMahabaleshwarkar: the upgrade path for any install that still has those reconcilers should look like 
`2.7 or lower` -> `2.8` -> `2.12` (at minimum) -> kserve raw GA. 
If i'm mistaken in this assumption I'll re-add the cleanup.

controllers/reconcilers/kserve_raw_route_reconciler.go
VedantMahabaleshwarkar: I didn't get this, could you elaborate?

controllers/reconcilers/kserve_raw_metrics_service_reconciler.go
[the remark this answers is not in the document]
israel-hdez: Well, we could enable it by default in the ConfigMap: https://github.com/opendatahub-io/kserve/blob/ae30b971320f45bc98dd362b6b8873d7c1a48873/config/configmap/inferenceservice.yaml#L447. But doing that would also affect Serverless deployments, so not sure if it will impact something.

What I don't know is if this is specific to Serverless, because docs only refer to the Serverless use-case.

In case it works for Raw, I said _to inject_ meaning we could use a mutating webhook in odh-model-controller to inject the annotation on the ISVC. I wasn't meaning that the UI should do it.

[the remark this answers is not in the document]
VedantMahabaleshwarkar: @israel-hdez revisiting this comment, we actively moved away from the kserve native annotations for metrics i.e `serving.kserve.io/enable-prometheus-scraping: "true"` for Serverless, and we have instead opted for `prometheus.io/port` and `prometheus.io/path` annotations on the ServingRuntime instead. 
So going back to `serving.kserve.io/enable-prometheus-scraping: "true"` for RawDeployment would IMO be more confusing. Creating a SeviceMonitor and Service for metrics will allow us to be consistent w.r.t metrics within RHOAI.

VedantMahabaleshwarkar:
not sure why the images job is failing, I think it's likely a infra issue. Local builds are succeeding

mwaykole:
hi @VedantMahabaleshwarkar do we need to add unit test for this pr ?

VedantMahabaleshwarkar:
@mwaykole about unit tests: 
- For Kserve RawDeployment Route : According to the conversation [here ](https://github.com/opendatahub-io/odh-model-controller/pull/250#discussion_r1712153605), the route creation will be done by enabling the ingress creation in the `inferenceservice-config` ConfigMap. So here odh-model-controller is no longer involved. No code, no unit tests 
- For Kserve Raw Route Auth : this PR is not currently targetting that, 
- For Kserve RawDeployment Metrics : the existing test for metrics (controllers/kserve_inferenceservice_controller_metrics_test.go) was modified to add test cases for RawDeployment, the changes are in this PR. 

Hope that helps

controllers/reconcilers/kserve_raw_metrics_service_reconciler.go
```diff
@@ -19,41 +19,44 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strconv"
 )
 
 const (
 	inferenceServiceLabelName = "serving.kserve.io/inferenceservice"
 )
 
-var _ SubResourceReconciler = (*KserveMetricsServiceReconciler)(nil)
+var _ SubResourceReconciler = (*KserveRawMetricsServiceReconciler)(nil)
```
mwaykole, line 40: so we will support KserveRawMetricsServiceReconciler as well ,  and it should work same as server less ?

VedantMahabaleshwarkar:
replaced by https://github.com/opendatahub-io/odh-model-controller/pull/274

PR #254, opened 2024-08-19
RHOAIENG-1044: Update hardcoded constants as environment variables 

mholder6:
## Description

Refactored the controllers/utils function GetIstioControlPlaneName to no longer reference IstioNamespace and IstioControlPlaneName constants, and to return errors instead of automatically setting the env vars. Removed the constants from the constants/constants.go file, and replaced all references of the constants in the code with a call to the utils function mentioned before. 

## Merge criteria:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" POD_NAMESPACE=default go test -v ./controllers/... -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" POD_NAMESPACE=default MESH_NAMESPACE=istio-system CONTROL_PLANE_NAME=istio-system go test -v ./controllers/... -coverprofile cover.out
 
 .PHONY: e2e-test
 e2e-test: manifests generate fmt vet ## Run e2e-tests.
--- a/controllers/constants/constants.go
+++ b/controllers/constants/constants.go
@@ -18,8 +18,6 @@ package constants
 const (
 	InferenceServiceKind = "InferenceService"
 
-	IstioNamespace                   = "istio-system"
-	IstioControlPlaneName            = "data-science-smcp"
 	ServiceMeshMemberRollName        = "default"
 	ServiceMeshMemberName            = "default"
 	IstioIngressService              = "istio-ingressgateway"
--- a/controllers/kserve_inferenceservice_controller_test.go
+++ b/controllers/kserve_inferenceservice_controller_test.go
@@ -28,6 +28,7 @@ import (
 	. "github.com/onsi/gomega/gstruct"
 	gomegatypes "github.com/onsi/gomega/types"
 	"github.com/opendatahub-io/odh-model-controller/controllers/constants"
+	"github.com/opendatahub-io/odh-model-controller/controllers/utils"
 	routev1 "github.com/openshift/api/route/v1"
 	istioclientv1beta1 "istio.io/client-go/pkg/apis/networking/v1beta1"
 	corev1 "k8s.io/api/core/v1"
@@ -69,6 +70,7 @@ var _ = Describe("The Openshift Kserve model controller", func() {
 		})
 
 		It("With Kserve InferenceService a Route be created", func() {
+			_, meshNamespace := utils.GetIstioControlPlaneName(ctx, cli)
 			inferenceService := &kservev1beta1.InferenceService{}
 			err := convertToStructuredResource(KserveInferenceServicePath1, inferenceService)
 			Expect(err).NotTo(HaveOccurred())
@@ -79,7 +81,7 @@ var _ = Describe("The Openshift Kserve model controller", func() {
 			By("By checking that the controller has not created the Route")
 			Consistently(func() error {
 				route := &routev1.Route{}
-				key := types.NamespacedName{Name: getKServeRouteName(inferenceService), Namespace: constants.IstioNamespace}
+				key := types.NamespacedName{Name: getKServeRouteName(inferenceService), Namespace: meshNamespace}
 				err = cli.Get(ctx, key, route)
 				return err
 			}, timeout, interval).Should(HaveOccurred())
@@ -98,7 +100,7 @@ var _ = Describe("The Openshift Kserve model controller", func() {
 			By("By checking that the controller has created the Route")
 			Eventually(func() error {
 				route := &routev1.Route{}
-				key := types.NamespacedName{Name: getKServeRouteName(inferenceService), Namespace: constants.IstioNamespace}
+				key := types.NamespacedName{Name: getKServeRouteName(inferenceService), Namespace: meshNamespace}
 				err = cli.Get(ctx, key, route)
 				return err
 			}, timeout, interval).Should(Succeed())
@@ -192,16 +194,18 @@ var _ = Describe("The Openshift Kserve model controller", func() {
 			err = cli.Status().Update(ctx, deployedInferenceService)
 			Expect(err).NotTo(HaveOccurred())
 
+			_, meshNamespace := utils.GetIstioControlPlaneName(ctx, cli)
+
 			// Verify that the certificate secret is created in the istio-system namespace.
 			Eventually(func() error {
 				secret := &corev1.Secret{}
-				return cli.Get(ctx, client.ObjectKey{Namespace: constants.IstioNamespace, Name: fmt.Sprintf("%s-%s", inferenceService.Name, inferenceService.Namespace)}, secret)
+				return cli.Get(ctx, client.ObjectKey{Namespace: meshNamespace, Name: fmt.Sprintf("%s-%s", inferenceService.Name, inferenceService.Namespace)}, secret)
 			}, timeout, interval).Should(Succeed())
 
 			// Verify that the gateway is updated in the istio-system namespace.
 			var gateway *istioclientv1beta1.Gateway
 			Eventually(func() error {
-				gateway, err = waitForUpdatedGatewayCompletion(cli, "add", constants.IstioNamespace, constants.KServeGatewayName, inferenceService.Name)
+				gateway, err = waitForUpdatedGatewayCompletion(cli, "add", meshNamespace, constants.KServeGatewayName, inferenceService.Name)
 				return err
 			}, timeout, interval).Should(Succeed())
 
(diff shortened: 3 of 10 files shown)


VedantMahabaleshwarkar:
@mholder6 the changelist on this PR looks wrong, needs a rebase probably

Containerfile
```diff
@@ -27,10 +23,6 @@ RUN CGO_ENABLED=0 GOOS=linux go build -a -o manager main.go
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
 WORKDIR /
 COPY --from=builder /workspace/manager .
-COPY --from=builder /workspace/metrics_dashboards/ovms-metrics.json .
```
spolti, line 30: Guess you need to rebase your PR.

spolti: a lot of unrelated changes.

controllers/utils/utils.go
```diff
@@ -186,8 +186,8 @@ func GetIstioControlPlaneName(ctx context.Context, cli client.Client) (istioCont
 		log.V(1).Info("Trying to read Istio Control Plane name and namespace from DSCI")
 		objectList, err := getDSCIObject(ctx, cli)
 		if err != nil {
-			log.V(0).Error(err, "Failed to fetch the DSCI object, using default values")
-			return constants.IstioControlPlaneName, constants.IstioNamespace
```
israel-hdez, line 190: For the time being, it may be better to invoke `panic`, rather than returning the error string in the namespace return value.

mholder6: got it, thanks for the input!!

controllers/utils/utils.go
```diff
@@ -19,8 +19,6 @@ import (
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	"github.com/opendatahub-io/odh-model-controller/controllers/constants"
```
hdefazio, line 23: I think you need to add this back

mholder6: The import was no longer being used in the file -- still should be added back?

hdefazio: I think after rebasing on the latest master it will need the import again

controllers/constants/constants.go
```diff
@@ -18,8 +18,6 @@ package constants
 const (
 	InferenceServiceKind = "InferenceService"
 
-	IstioNamespace                   = "istio-system"
```
spolti, line 21: Note that it is still used in case the func `utils.GetIstioControlPlaneName`

can't find anything: https://github.com/mholder6/odh-model-controller/blob/7a14b2092913e57cc2367ce14c0fa7787b2b6d1e/controllers/utils/utils.go#L217C6-L217C46

same for `IstioControlPlaneName`

It can't be deleted, you need to override where these values are hardcoded to the values returned from this function.

controllers/utils/utils.go
```diff
@@ -213,8 +212,8 @@ func GetIstioControlPlaneName(ctx context.Context, cli client.Client) (istioCont
 					meshNamespace = namespace
 				} else {
 					log.V(1).Info("Mesh Namespace is not set in DSCI")
-					// at this point, it is not set anywhere, lets just use the default
-					meshNamespace = constants.IstioNamespace
+					// at this point, it is not set anywhere, lets return an error
```
spolti, line 215: it is better to return the default instead panicking, no @israel-hdez ?

controllers/constants/constants.go
mholder6: Yes I actually am invoking panic instead of defaulting to those values -- you can see this here:
https://github.com/opendatahub-io/odh-model-controller/pull/254/files#diff-d93cfa921e60384e543609474ebc28fdbad7ae20f135f76398ef8d821333b3f2R194

controllers/utils/utils.go
mholder6: thanks for this! I've rebased and pulled all the recent changes now. :)

controllers/utils/utils.go
spolti: actually, it is logging an error and returning the default values, keeping backward capabilities.

controllers/utils/utils.go
israel-hdez: My rationale here is that we shouldn't be using any defaults in the code. We should be forcing this to be specified.
Hoping I'm not wrong, it is always going to be set on the Deployment. If it is not set, the manifests would be faulty.

config/webhook/manifests.yaml
```diff
@@ -9,7 +9,7 @@ webhooks:
   clientConfig:
     service:
       name: webhook-service
-      namespace: $(mesh-namespace)
+      namespace: system
```
israel-hdez, line 12: Can you revert this? It doesn't look related.

controllers/suite_test.go
```diff
@@ -52,7 +52,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
-	"github.com/opendatahub-io/odh-model-controller/controllers/constants"
+	//"github.com/opendatahub-io/odh-model-controller/controllers/constants"
```
israel-hdez, line 55: Remove comment.

config/webhook/manifests.yaml
spolti: Yeah. it is updated every time the manifests are generated.
@Jooho fyi.

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -192,16 +194,18 @@ var _ = Describe("The Openshift Kserve model controller", func() {
 			err = cli.Status().Update(ctx, deployedInferenceService)
 			Expect(err).NotTo(HaveOccurred())
 
+			_, meshName := utils.GetIstioControlPlaneName(ctx, cli)
```
spolti, line 197: please rename to meshNamespace to avoid confunsions.

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -246,6 +250,7 @@ var _ = Describe("The Openshift Kserve model controller", func() {
 			ctx := context.Background()
 			testNamespace := Namespaces.Create(cli)
 			testNs = testNamespace.Name
+			_, meshName := utils.GetIstioControlPlaneName(ctx, cli)
```
spolti, line 253: same

controllers/kserve_inferenceservice_controller_test.go
```diff
@@ -366,10 +372,12 @@ var _ = Describe("The Openshift Kserve model controller", func() {
 				Expect(err).NotTo(HaveOccurred())
 				Expect(cli.Delete(ctx, deployedInferenceService)).Should(Succeed())
 
+				_, meshName := utils.GetIstioControlPlaneName(ctx, cli)
```
spolti, line 375: here.

controllers/reconcilers/kserve_route_reconciler.go
```diff
@@ -129,10 +131,12 @@ func (r *KserveRouteReconciler) createDesiredResource(isvc *kservev1beta1.Infere
 			}
 		}
 
+		_, meshName := utils2.GetIstioControlPlaneName(context.Background(), r.client)
+
 		route := &v1.Route{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:        getKServeRouteName(isvc),
-				Namespace:   constants2.IstioNamespace,
+				Namespace:   meshName,
```
spolti, line 139: here as well.

controllers/suite_test.go
```diff
@@ -140,10 +140,11 @@ var _ = BeforeSuite(func() {
 	Expect(cli).NotTo(BeNil())
 
 	// Create istio-system namespace
+	_, meshName := utils.GetIstioControlPlaneName(ctx, cli)
 	istioNamespace := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      constants.IstioNamespace,
-			Namespace: constants.IstioNamespace,
+			Name:      meshName,
```
spolti, line 146: here as well.

controllers/utils/utils.go
spolti: That make sense, but is this the kind of error we want to block the operator startup?

Makefile
```diff
@@ -69,7 +69,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" POD_NAMESPACE=default go test -v ./controllers/... -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" POD_NAMESPACE=default MESH_NAMESPACE=istio-system CONTROL_PLANE_NAME=istio-system go test -v ./controllers/... -coverprofile cover.out
```
spolti, line 72: Can you please elaborate on why is this needed?

mholder6: I believe the test was not building correctly before making these changes, @israel-hdez do you remember specifically why this was necessary?

israel-hdez: Since there is no default value anymore, running tests would fail if we don't set these envvars, because the code would see empty values.

controllers/utils/utils.go
israel-hdez: As it is, this should not block operator startup. Although, it will crash during reconciliation of some ISVC.
I expect no crash happening as long as all configuration is correct.

But yes, we can block operator startup to fail early. The panics would work as a back-up. Probably better to do that in another PR to not complicate this one.

israel-hdez (approved):
It is looking good to me now.

I'll let @spolti to give the final approval, since he also provided some feedback.

controllers/utils/utils.go
```diff
@@ -245,7 +244,8 @@ func VerifyIfMeshAuthorizationIsEnabled(ctx context.Context, cli client.Client)
 	// is available, taking into account both managed an unmanaged configurations.
 
 	authPolicies := &v1beta1.AuthorizationPolicyList{}
-	if err := cli.List(ctx, authPolicies, client.InNamespace(constants.IstioNamespace)); err != nil && !meta.IsNoMatchError(err) {
+	_, meshName := GetIstioControlPlaneName(ctx, cli)
+	if err := cli.List(ctx, authPolicies, client.InNamespace(meshName)); err != nil && !meta.IsNoMatchError(err) {
```
spolti, line 248: you missed some places like this :)

spolti: sry about asking you to rename, but it it makes things easier to read :)

mholder6: No problem, everything should be updated now :)

config/webhook/manifests.yaml
spolti: @mholder6 can you please revert?

controllers/utils/utils.go
spolti: Right, @mholder6 can you please open a follow-up Jira?

spolti (commented):
Just one more thing and I think it is good to go.

spolti (approved):

PR #256, opened 2024-08-20
[RHOAIENG-11527] - [backend] Incorrect values for number of requests …

spolti:
…in Kserve Performance metrics
A summarized decision:
```
Update the chart title to match the approved by UXD. Use "Requests rate per 5 min" the same way as the model mesh to avoid confusion and maintain consistency as it denotes rate instead concrete values.

```

## Merge criteria:

### DIFF
--- a/controllers/constants/runtime-metrics.go
+++ b/controllers/constants/runtime-metrics.go
@@ -20,16 +20,16 @@ const (
 	CaikitMetricsData = `{
         "config": [
 			{
-				"title": "Number of requests",
+				"title": "Requests per 5 minutes",
 				"type": "REQUEST_COUNT",
 				"queries": [
 					{
 						"title": "Number of successful incoming requests",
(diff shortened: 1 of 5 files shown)


controllers/constants/runtime-metrics.go
```diff
@@ -15,21 +15,23 @@ limitations under the License.
 
 package constants
 
+// ${DASHBOARD_RATE_INTERVAL} is a placeholder for the metrics that needs to receive values from the dashboard at the
```
terrytangyuan, line 18: This should be `${REQUEST_RATE_INTERVAL}`?

spolti: actually this comment needs to be removed, thanks for the heads up.

spolti: updated, thanks

israel-hdez (approved):
/lgtm

We are moving forward this PR. This part is working fine, but there seems to be an issue in the dashboard leading to a bad chart.

PR #257, opened 2024-08-21
runtimes: remove distributed executor backend option from vllm

dtrifiro:
it's the default


israel-hdez:
@dtrifiro May you either mention the ticket/jira, or have somebody else from runtimes to approve this?

PR #264, opened 2024-09-11
[RHOAIENG-12726] - odh-model-controller crashes with empty isvc.model…

spolti:
… field

chore:	fix nil pointer if the custom isvc does not have the model field.

## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #265, opened 2024-09-13
[RHOAIENG-12726] - odh-model-controller crashes with empty isvc.model field

spolti:
Cherry-picks https://github.com/opendatahub-io/odh-model-controller/pull/264


spolti:
sync already done.

PR #268, opened 2024-09-20
[RHOAIENG-7034] Updated manifests to have 1 replica of odh-model-controller pods 

mholder6:
Updated manifests to have 1 replica of odh-model-controller pods instead of 3. 

- The `config/manager/manager.yaml` is the main definition of odh-model-controller deployment -- change is needed here for replica value to be 1
- The `/overlay/dev` is to define some manifests that we don't want to deploy in ODH/RHOAI but are needed for local development or unit tests -- change is needed for the replica value to be 1
- The `overlays/odh/manager_patch` doesn't need to specify the replicas in this patch file anymore since the main file has it set to 1 -- replica completely removed here. 

Tested by running `kustomize build` after making changes and ensured that the replicas were set to 1 instead of 3. 

## Merge criteria:

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -9,7 +9,7 @@ spec:
   selector:
     matchLabels:
       control-plane: odh-model-controller
-  replicas: 3
+  replicas: 1
   template:
     metadata:
       annotations:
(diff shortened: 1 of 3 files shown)


israel-hdez (approved):
/lgtm

PR #274, opened 2024-10-09
Add reconciliation for Kserve Raw

VedantMahabaleshwarkar:
## Description

Fixes : https://issues.redhat.com/browse/RHOAIENG-10291, https://issues.redhat.com/browse/RHOAIENG-13444 
- Added logic to create `kserve-sa` ServiceAccount for kserve raw, and create a CRB for auth delegation. This is needed for oauth proxy for kserve raw.  
- Added logic to create route if label is added to isvc
Test with : https://github.com/opendatahub-io/kserve/pull/419

## Merge criteria:

### DIFF
--- a/controllers/reconcilers/kserve_raw_route_reconciler.go
+++ b/controllers/reconcilers/kserve_raw_route_reconciler.go
@@ -120,15 +120,16 @@ func (r *KserveRawRouteReconciler) createDesiredResource(ctx context.Context, lo
 			}
 		}
 	}
-	var desiredPort string
-	if !enableAuth {
-		desiredPort = "http"
-	} else {
-		desiredPort = "https"
-	}
+
 	for _, port := range targetService.Spec.Ports {
-		if port.Name == desiredPort {
-			servicePort = port.Port
+		if !enableAuth {
+			if port.Name == "http" || port.Name == targetService.Name {
+				servicePort = port.Port
+			}
+		} else {
+			if port.Name == "https" {
+				servicePort = port.Port
+			}
 		}
 	}
 
--- a/controllers/resources/clusterrolebinding.go
+++ b/controllers/resources/clusterrolebinding.go
@@ -109,7 +109,10 @@ func (r *clusterRoleBindingHandler) ProcessDelta(ctx context.Context, log logr.L
 	if delta.IsAdded() {
 		log.V(1).Info("Delta found", "create", desiredCRB.GetName())
 		if err = r.client.Create(ctx, desiredCRB); err != nil {
-			return
+			if errors.IsAlreadyExists(err) {
+				return nil
+			}
+			return err
 		}
 	}
 	if delta.IsUpdated() {


spolti:
Please sign the commit.

terrytangyuan (commented):
Is this ready for review? Looks like still in draft.

VedantMahabaleshwarkar:
> Is this ready for review? Looks like still in draft.

@terrytangyuan yes marked as ready for review

controllers/reconcilers/kserve_raw_inferenceservice_reconciler.go
```diff
@@ -26,16 +29,65 @@ import (
...
+	var deleteErrors *multierror.Error
+	for _, reconciler := range r.subResourceReconcilers {
+		deleteErrors = multierror.Append(deleteErrors, reconciler.Delete(ctx, log, isvc))
+	}
+
+	return deleteErrors.ErrorOrNil()
+}
+
+func (r *KserveRawInferenceServiceReconciler) CleanupNamespaceIfNoKserveIsvcExists(ctx context.Context, log logr.Logger, namespace string) error {
+	inferenceServiceList := &kservev1beta1.InferenceServiceList{}
```
spolti, line 68: shouldn't we have this in a common shared method to use with non-raw as well?

controllers/reconcilers/kserve_raw_inferenceservice_reconciler.go
```diff
@@ -26,16 +29,65 @@ import (
...
+	if err := r.client.List(ctx, inferenceServiceList, client.InNamespace(namespace)); err != nil {
+		return err
+	}
+
+	for i := len(inferenceServiceList.Items) - 1; i >= 0; i-- {
```
spolti, line 73: why not use range?

controllers/kserve_raw_inferenceservice_controller_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+		inferenceServiceConfig := &corev1.ConfigMap{}
+		Expect(convertToStructuredResource(InferenceServiceConfigPath1, inferenceServiceConfig)).To(Succeed())
+		if err := cli.Create(ctx, inferenceServiceConfig); err != nil && !apierrs.IsAlreadyExists(err) {
+			Fail(err.Error())
+		}
+
+	})
+
+	When("deploying a Kserve RawDeployment model", func() {
```
spolti, line 62: Maybe add a test to check if the service accounts and rolebindings are deleted, wdyt?

spolti (commented):
Just a few comments.
Nice work btw

controllers/reconcilers/kserve_raw_inferenceservice_reconciler.go
VedantMahabaleshwarkar: we have this method for serverless as well : https://github.com/opendatahub-io/odh-model-controller/blob/main/controllers/reconcilers/kserve_serverless_inferenceservice_reconciler.go#L79 

Each will perform cleanup on only their type of isvcs (raw reconciler will only cleanup raw isvcs, etc)

spolti: yep, that's the point, these are basically copy from each other, simplifying it would be a good thing :)

controllers/kserve_raw_inferenceservice_controller_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"time"
+
+	kservev1alpha1 "github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
+	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+)
+
+const (
+	KserveSA = "kserve-sa"
```
terrytangyuan, line 18: Reuse constant

terrytangyuan (approved):
Just a small comment. Otherwise LGTM.

controllers/kserve_raw_inferenceservice_controller_test.go
VedantMahabaleshwarkar: done

controllers/kserve_raw_inferenceservice_controller_test.go
VedantMahabaleshwarkar: added

controllers/constants/constants.go
```diff
@@ -31,6 +31,9 @@ const (
 	LabelEnableRoute   = "enable-route"
 
 	CapabilityServiceMeshAuthorization = "CapabilityServiceMeshAuthorization"
+
+	ModelMeshServiceAccountName = "modelmesh-serving-sa"
+	KserveServiceAccountName    = "kserve-sa"
```
israel-hdez, line 36: Since this is going to be for oauth-proxy uses, maybe?:
```suggestion
	KserveServiceAccountName    = "kserve-oauth-sa"
```

controllers/reconcilers/kserve_raw_inferenceservice_reconciler.go
```diff
@@ -17,6 +17,9 @@ package reconcilers
 
 import (
 	"context"
+	"github.com/hashicorp/go-multierror"
```
israel-hdez, line 20: Add an empty line after `context`

controllers/reconcilers/kserve_raw_inferenceservice_reconciler.go
```diff
@@ -17,6 +17,9 @@ package reconcilers
...
+	"github.com/opendatahub-io/odh-model-controller/controllers/constants"
+	"github.com/opendatahub-io/odh-model-controller/controllers/utils"
```
israel-hdez, line 22: Remove the empty line after this one.

controllers/reconcilers/kserve_raw_inferenceservice_reconciler.go
```diff
@@ -26,16 +29,65 @@ import (
...
+		isvcDeploymentMode, err := utils.GetDeploymentModeForIsvc(ctx, r.client, &inferenceService)
+		if err != nil {
+			return err
+		}
+		if isvcDeploymentMode != utils.RawDeployment {
+			inferenceServiceList.Items = append(inferenceServiceList.Items[:i], inferenceServiceList.Items[i+1:]...)
+		}
+	}
+
+	// If there are no Kserve InferenceServices in the namespace, delete namespace-scoped resources needed for Kserve Metrics
```
israel-hdez, line 84:
```suggestion
	// If there are no Raw Kserve InferenceServices in the namespace, delete namespace-scoped resources
```

israel-hdez (commented):
As much as I like code reusability, I think I prefer more changes to be self-contained.

The changes in opendatahub-io/kserve#419 won't work without the changes here.
I think I prefer porting some code from here to our KServe fork, so that kserve-controller is self-sufficient.

Just for your consideration. Otherwise, I left some small comments.

VedantMahabaleshwarkar:
> hey @VedantMahabaleshwarkar where are the unit test for this feature

https://github.com/opendatahub-io/odh-model-controller/pull/274/files#diff-879306cf12962e807b6e44b8ca651a940ed20a6ca8372ed0bdb21f752a385323R21 

@mwaykole

controllers/kserve_raw_inferenceservice_controller_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"time"
+
+	kservev1alpha1 "github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
+	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+)
```
terrytangyuan, line 19: Looks like imports are not sorted. Do we have a lint check in this repo?

controllers/reconcilers/kserve_raw_route_reconciler.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	if !createRoute {
+		log.Info("InferenceService does not have label '" + constants.KserveNetworkVisibility + "' annotation" +
+			" set to '" + constants.LabelEnableRoute + "'. Skipping route creation")
+		return nil, nil
+	}
+
+	// Fetch the service with the label "serving.kserve.io/inferenceservice=isvc.Name" in the isvc namespace
+	serviceList := &corev1.ServiceList{}
+	labelSelector := client.MatchingLabels{"serving.kserve.io/inferenceservice": isvc.Name}
```
terrytangyuan, line 108: Is there a constant to reuse?

controllers/reconcilers/kserve_raw_route_reconciler.go
```diff
@@ -0,0 +1,195 @@
...
+	err = r.client.List(ctx, serviceList, client.InNamespace(isvc.Namespace), labelSelector)
+	if err != nil || len(serviceList.Items) == 0 {
+		log.Error(err, "Failed to fetch service for InferenceService", "InferenceService", isvc.Name)
+		return nil, err
+	}
+	targetServiceName := serviceList.Items[0].Name
```
israel-hdez, line 114: You can have more than one if the `InferenceService` has transformers, predictors,etc... In that case, I don't think you can blindly use the first one.

controllers/reconcilers/kserve_raw_route_reconciler.go
```diff
@@ -0,0 +1,195 @@
...
+			},
+		},
+		Spec: v1.RouteSpec{
+			To: v1.RouteTargetReference{
+				Kind:   "Service",
+				Name:   targetServiceName,
+				Weight: pointer.Int32(100),
+			},
+			Port: &v1.RoutePort{
+				TargetPort: intstr.FromInt(isvcServicePort),
```
israel-hdez, line 131: I don't think we can hard-code to port `8888`.

Caikit runtimes are using `8080`, OVMS is using `8888`, tgis is using `8033`, and vLLM is using `8080`. If we add new runtimes, other ports may be coming. So, it is better to resolve the port.

...But the Route is pointing to the `Service`. So, I wonder if it shouldn't be the common http port?

controllers/constants/constants.go
```diff
@@ -24,13 +24,17 @@ const (
 	IstioIngressServiceHTTPPortName  = "http2"
 	IstioIngressServiceHTTPSPortName = "https"
 	IstioSidecarInjectAnnotationName = "sidecar.istio.io/inject"
+	KserveNetworkVisibility          = "networking.kserve.io/visibility"
 
 	LabelAuthGroup     = "security.opendatahub.io/authorization-group"
 	LabelEnableAuthODH = "security.opendatahub.io/enable-auth"
```
mwaykole, line 30: Here's a reworded version:
IMO
Instead of using enable-auth, it should be set to true. This will appear as "networking.kserve.io/visibility": "true". 
 how to change the status if they want to use disable-auth with current node  ?

controllers/reconcilers/kserve_raw_route_reconciler.go
VedantMahabaleshwarkar: https://github.com/opendatahub-io/kserve/pull/419/files#diff-401b6cf36f71bdda9ab1593723f98632b0e7ee8d3a98151d65439e13d8d2ea9fR264 
the oauth exposed port is 8443, the oauth proxy will in turn route to the correct runtime port

controllers/constants/constants.go
VedantMahabaleshwarkar: To disable route - `networking.kserve.io/visibility : cluster-local` 
To disable auth - remove label `security.opendatahub.io/enable-auth`

spolti:
There are still some comments not addressed from @israel-hdez plus the failing test.

/approve

VedantMahabaleshwarkar:
>There are still some comments not addressed from @israel-hdez plus the failing test.

@spolti 2 comments from edgar pending are about import sort, not sure why but that's how goland is sorting the imports for me even if I manually rearrange. 

About the failing unit test, 
This is what the logs are showing 
```
FAIL	github.com/opendatahub-io/odh-model-controller/controllers	0.040s
=== RUN   TestAPIs
Running Suite: Resources Suite
==============================
Random Seed: 1731430313
Will run 7 of 7 specs
•••••••
Ran 7 of 7 Specs in 0.010 seconds
SUCCESS! -- 7 Passed | 0 Failed | 0 Pending | 0 Skipped
``` 
Not sure what the failure is, it says 0 Failed as well

controllers/reconcilers/kserve_metrics_servicemonitor_reconciler.go
```diff
@@ -67,9 +77,36 @@ func (r *KserveMetricsServiceMonitorReconciler) Reconcile(ctx context.Context, l
...
-// TODO remove this reconcile loop in future versions
-func (r *KserveMetricsServiceMonitorReconciler) createDesiredResource(isvc *kservev1beta1.InferenceService) (*v1.ServiceMonitor, error) {
-	return nil, nil
+func (r *KserveMetricsServiceMonitorReconciler) createDesiredResource(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) (*v1.ServiceMonitor, error) {
+	isvcRuntime, err := utils.FindSupportingRuntimeForISvc(ctx, r.client, log, isvc)
+	if err != nil {
+		return nil, err
+	}
+
+	desiredServiceMonitor := &v1.ServiceMonitor{
```
israel-hdez, line 86: I'm very confused about including metrics in this PR. This wasn't here on my previous review. Also, the Jiras listed in the PR description don't include any Jiras about metrics.

Do we really want metrics in this PR?

controllers/reconcilers/kserve_raw_route_reconciler.go
israel-hdez: You are right: when auth is enabled there is some code below that will change this to `8443`. However, this line seems to be covering the case when auth is disabled. For that case, it looks like the Route will be pointing to an erroneous port.

spolti:
> > There are still some comments not addressed from @israel-hdez plus the failing test.
> 
> @spolti 2 comments from edgar pending are about import sort, not sure why but that's how goland is sorting the imports for me even if I manually rearrange.
> 
> About the failing unit test, This is what the logs are showing
> 
> ```
> FAIL	github.com/opendatahub-io/odh-model-controller/controllers	0.040s
> === RUN   TestAPIs
> Running Suite: Resources Suite
> ==============================
> Random Seed: 1731430313
> Will run 7 of 7 specs
> •••••••
> Ran 7 of 7 Specs in 0.010 seconds
> SUCCESS! -- 7 Passed | 0 Failed | 0 Pending | 0 Skipped
> ```
> 
> Not sure what the failure is, it says 0 Failed as well

There is a nil pointer happening around here:
/go/src/github.com/opendatahub-io/odh-model-controller/controllers/kserve_raw_inferenceservice_controller_test.go:25 +0xde


```go
--- FAIL: TestAPIs (0.01s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1af059e]
goroutine 46 [running]:
testing.tRunner.func1.2({0x1cf5060, 0x3705c00})
	/usr/local/go/src/testing/testing.go:1545 +0x238
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1548 +0x397
panic({0x1cf5060?, 0x3705c00?})
	/usr/local/go/src/runtime/panic.go:914 +0x21f
github.com/opendatahub-io/odh-model-controller/controllers.glob..func6.1({0x0, 0x0}, {0x2070183, 0x38})
	/go/src/github.com/opendatahub-io/odh-model-controller/controllers/kserve_raw_inferenceservice_controller_test.go:25 +0xde
github.com/opendatahub-io/odh-model-controller/controllers.glob..func6.4()
	/go/src/github.com/opendatahub-io/odh-model-controller/controllers/kserve_raw_inferenceservice_controller_test.go:56 +0x44
github.com/onsi/ginkgo/internal/suite.(*Suite).PushContainerNode(0xc000871d50, {0xc00011e390, 0x2b}, 0xc000c941a0, 0x0, {{0x2b6cc56, 0x71}, 0x37, {0xc000b1d0[80](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/opendatahub-io_odh-model-controller/274/pull-ci-opendatahub-io-odh-model-controller-main-unit/1856378018981744640#1:build-log.txt%3A80), 0x56a}})
14 skipped lines...
	/go/src/github.com/opendatahub-io/odh-model-controller/controllers/suite_test.go:102 +0x3c
testing.tRunner(0xc000471860, 0x212a[82](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/opendatahub-io_odh-model-controller/274/pull-ci-opendatahub-io-odh-model-controller-main-unit/1856378018981744640#1:build-log.txt%3A82)0)
	/usr/local/go/src/testing/testing.go:1595 +0xff
created by testing.(*T).Run in goroutine 1
	/usr/local/go/src/testing/testing.go:1648 +0x3ad
FAIL	github.com/opendatahub-io/odh-model-controller/controllers	0.040s
```

VedantMahabaleshwarkar:
the unit test failure is weird, and not an actual failure IMO, more of a timeout issue. 
the odh-model-controller logs from the test output show
```
2024-11-21T14:19:58Z	DEBUG	controllers.InferenceService-controller	Deleting ClusterRoleBinding object for target namespace	{"InferenceService": "example-onnx-mnist", "namespace": "test-ns-gmwss"}
```

Timed out after 20.001s.
              
    Unexpected error:
        <*errors.errorString | 0xc00066b780>: 
        crb deletion not detected
        {
            s: "crb deletion not detected",
        }
    occurred

but the `client.get` call doesn't return an error before the timeout. Testing on an actual cluster I verified the CRB gets deleted as a part of the isvc cleanup

spolti:
Even if it is not a real issue, it is related to the changes introduced by this pull request, we need to fix it, otherwise we will have this issue in all other PRs after this gets merged.

israel-hdez (changes requested):
Please, change the target branch to `incubating`.
The `main` branch is now a stable/downstream branch.

controllers/kserve_raw_inferenceservice_controller_test.go
```diff
@@ -0,0 +1,110 @@
...
+				key := types.NamespacedName{Name: inferenceService.Name, Namespace: inferenceService.Namespace}
+				return cli.Get(ctx, key, route)
+			}, timeout, interval).Should(HaveOccurred())
+		})
+	})
+	When("namespace no longer has any RawDeployment models", func() {
+		It("should delete the clusterrolebinding", func() {
+			_ = createServingRuntime(testNs, KserveServingRuntimePath1)
+			inferenceService := createInferenceService(testNs, KserveOvmsInferenceServiceName, KserveInferenceServicePath1)
+			if err := cli.Create(ctx, inferenceService); err != nil && !apierrs.IsAlreadyExists(err) {
```
spolti, line 82: for these ifs I would suggest removing it, and just check with Expect, it is valid for the other test cases.
If issues are happening with duplicated resources, use a different name, it might be the cause of the timeout happening in this test.

controllers/reconcilers/kserve_raw_inferenceservice_reconciler.go
```diff
@@ -26,16 +30,65 @@ import (
...
+		return err
+	}
+
+	for i := len(inferenceServiceList.Items) - 1; i >= 0; i-- {
+		inferenceService := inferenceServiceList.Items[i]
+		isvcDeploymentMode, err := utils.GetDeploymentModeForIsvc(ctx, r.client, &inferenceService)
+		if err != nil {
+			return err
+		}
+		if isvcDeploymentMode != utils.RawDeployment {
```
spolti, line 80: shouldn't this `utils.RawDeployment`be moved to constants?

controllers/reconcilers/kserve_raw_route_reconciler.go
```diff
@@ -0,0 +1,205 @@
...
+	return nil
+}
+
+func (r *KserveRawRouteReconciler) Delete(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+	log.V(1).Info("Deleting Kserve inference service generic route")
+	return r.routeHandler.DeleteRoute(ctx, types.NamespacedName{Name: isvc.Name, Namespace: isvc.Namespace})
+}
+
+func (r *KserveRawRouteReconciler) Cleanup(_ context.Context, _ logr.Logger, _ string) error {
+	// NOOP - resources are deleted together with ISVCs
```
spolti, line 82: can you add a test case for this?

controllers/reconcilers/kserve_raw_inferenceservice_reconciler.go
israel-hdez: I think this comment still stands.

controllers/reconcilers/kserve_raw_route_reconciler.go
```diff
@@ -0,0 +1,205 @@
...
+	// Process Delta
+	if err = r.processDelta(ctx, log, desiredResource, existingResource); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *KserveRawRouteReconciler) Delete(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+	log.V(1).Info("Deleting Kserve inference service generic route")
+	return r.routeHandler.DeleteRoute(ctx, types.NamespacedName{Name: isvc.Name, Namespace: isvc.Namespace})
```
israel-hdez, line 78: Is this needed at all?
Looks like you have properly set the owner reference. The cluster should do the clean-up for you.

controllers/reconcilers/kserve_raw_route_reconciler.go
```diff
@@ -0,0 +1,205 @@
...
+	labelSelector := client.MatchingLabels{constants.KserveGroupAnnotation: isvc.Name}
+	err = r.client.List(ctx, serviceList, client.InNamespace(isvc.Namespace), labelSelector)
+	if err != nil || len(serviceList.Items) == 0 {
+		log.Error(err, "Failed to fetch service for InferenceService", "InferenceService", isvc.Name)
+		return nil, err
+	}
+	var targetServiceName string
+	var servicePort int32
+	for _, service := range serviceList.Items {
+		if val, ok := service.Labels["component"]; ok && val == "predictor" {
```
israel-hdez, line 113: Similarly to [a comment in the other PR](https://github.com/opendatahub-io/kserve/pull/419#discussion_r1805380869), you cannot hard-code to the `predictor` service, because the Route would be bypassing the `transformer` in case the user has chosen to configure a transformer.

In serverless there is a _top-level_ service that would hide the detail of choosing between predictor and transformer. I'm not sure if Raw has it. If it exists, you should use that top-level service.

controllers/reconcilers/serviceaccount_reconciler.go
```diff
@@ -72,26 +71,26 @@ func (r *ModelMeshServiceAccountReconciler) Reconcile(ctx context.Context, log l
 	return nil
 }
 
-func (r *ModelMeshServiceAccountReconciler) Cleanup(ctx context.Context, log logr.Logger, isvcNs string) error {
+func (r *ServiceAccountReconciler) Cleanup(ctx context.Context, log logr.Logger, isvcNs string) error {
 	log.V(1).Info("Deleting ServiceAccount object for target namespace")
-	return r.serviceAccountHandler.DeleteServiceAccount(ctx, types.NamespacedName{Name: modelMeshServiceAccountName, Namespace: isvcNs})
+	return r.serviceAccountHandler.DeleteServiceAccount(ctx, types.NamespacedName{Name: r.serviceAccountName, Namespace: isvcNs})
```
israel-hdez, line 76: Looks like you've changed the KServe service account to  `default`... I don't think we should be deleting a ServiceAccount that we didn't create. Deleting the `default` SA can have an impact to the user.

controllers/reconcilers/kserve_raw_route_reconciler.go
VedantMahabaleshwarkar: raw doesn't have a top level service from what I saw, I added logic to check for the presence of a transformer service here, and if a transformer is detected that will be used as the target service for creating the route

controllers/reconcilers/serviceaccount_reconciler.go
VedantMahabaleshwarkar: I'm not calling the serviceaccount reconciler from the raw inferenceservice reconciler so the `default` SA is not deleted

israel-hdez (commented):
Quite fine. I'll do some testings on my Tuesday.

controllers/kserve_raw_inferenceservice_controller_test.go
```diff
@@ -0,0 +1,185 @@
...
+			service := &corev1.Service{}
+			Eventually(func() error {
+				key := types.NamespacedName{Name: isvcService.Name, Namespace: isvcService.Namespace}
+				return cli.Get(ctx, key, service)
+			}, timeout, interval).Should(Succeed())
+			route := &routev1.Route{}
+			Eventually(func() error {
+				key := types.NamespacedName{Name: inferenceService.Name, Namespace: inferenceService.Namespace}
+				return cli.Get(ctx, key, route)
+			}, timeout, interval).ShouldNot(HaveOccurred())
```
israel-hdez, line 145: About your comment in Slack, in CI I'm seeing this check failing.

In CI, I do see in the logs of the controller `Failed to fetch service for InferenceService`. However, that log message does not necessarily come because of an error. I see the code:

```go
if err != nil || len(serviceList.Items) == 0 {
		log.Error(err, "Failed to fetch service for InferenceService", "InferenceService", isvc.Name)
		return nil, err
}
```

Most likely, `err` is nil, but the query didn't return any items. Effectively, it returns `(nil, nil)` So, the reconcile continues normally as if no route is wanted and without dropping any error that would retry the reconcile. This would be leading to a failing test here.

I see this just as a race condition where the reconcile finishes before this unit is able to create the Service.

VedantMahabaleshwarkar: I moved the service creation before the isvc to avoid the race condition.

israel-hdez (approved):
/lgtm

Approving. Needs follow-up work.

PR #284, opened 2024-11-04
[RHOAIENG-15386] add a new servingruntime vllm-multinode-template

Jooho:
## Description

This is for a new servingRuntime `vllm-multinode-template` for vLLM multinode feature 
It was based on 2 templates : [vllm-template.yaml ](https://github.com/opendatahub-io/odh-model-controller/blob/main/config/runtimes/vllm-template.yaml), [ibm poc template](https://github.ibm.com/ai-foundation/fmaas-inference-server/issues/785)

## How Has This Been Tested?

It requires GPUs.
You can refer to this doc for testing. (https://docs.google.com/document/d/1GA-foSoC1NprgC_vEv_oY_23KFG6opOi99JNWg3YI1Y/edit?tab=t.0)

## Merge criteria:

### DIFF
--- a/config/runtimes/vllm-multinode-template.yaml
+++ b/config/runtimes/vllm-multinode-template.yaml
@@ -6,11 +6,11 @@ metadata:
     opendatahub.io/ootb: "true"
   annotations:
     description: vLLM is a high-throughput and memory-efficient inference and serving engine for LLMs
-    openshift.io/display-name: vLLM ServingRuntime for KServe
+    openshift.io/display-name: vLLM ServingRuntime Multi Node for KServe
     openshift.io/provider-display-name: Red Hat, Inc.
-    tags: rhods,rhoai,kserve,servingruntime
+    tags: rhods,rhoai,kserve,servingruntime,multi-node
     template.openshift.io/documentation-url: https://github.com/opendatahub-io/vllm
-    template.openshift.io/long-description: This template defines resources needed to deploy vLLM ServingRuntime with KServe in Red Hat OpenShift AI
+    template.openshift.io/long-description: This template defines resources needed to deploy vLLM ServingRuntime Multi-Node with KServe in Red Hat OpenShift AI
     opendatahub.io/modelServingSupport: '["single"]'
     opendatahub.io/apiProtocol: "REST"
   name: vllm-multinode-runtime-template
--- a/config/runtimes/vllm-multinode-template.yaml
+++ b/config/runtimes/vllm-multinode-template.yaml
@@ -1,0 +1,256 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  labels:
+    opendatahub.io/dashboard: "true"
+    opendatahub.io/ootb: "true"
+  annotations:
+    description: vLLM is a high-throughput and memory-efficient inference and serving engine for LLMs
+    openshift.io/display-name: vLLM ServingRuntime for KServe
+    openshift.io/provider-display-name: Red Hat, Inc.
+    tags: rhods,rhoai,kserve,servingruntime
+    template.openshift.io/documentation-url: https://github.com/opendatahub-io/vllm
+    template.openshift.io/long-description: This template defines resources needed to deploy vLLM ServingRuntime with KServe in Red Hat OpenShift AI
+    opendatahub.io/modelServingSupport: '["single"]'
+    opendatahub.io/apiProtocol: "REST"
+  name: vllm-multinode-runtime-template
+objects:
+  - apiVersion: serving.kserve.io/v1alpha1
+    kind: ServingRuntime
+    metadata:
+      name: vllm-multinode-runtime
+      annotations:
+        openshift.io/display-name: vLLM ServingRuntime for KServe
+        opendatahub.io/recommended-accelerators: '["nvidia.com/gpu"]'
+      labels:
+        opendatahub.io/dashboard: "true"
+    spec:
+      annotations:
+        prometheus.io/port: "8080"
+        prometheus.io/path: "/metrics"
+      multiModel: false
+      supportedModelFormats:
+        - autoSelect: true
+          name: vLLM
+          priority: 2
+      containers:
+        - name: kserve-container
+          image: $(vllm-image)
+          command: ["bash", "-c"]
+          args:
+            - |
+              ray start --head --disable-usage-stats --include-dashboard false 
+              # wait for other node to join
+              until [[ $(ray status | grep -c node_) -eq ${PIPELINE_PARALLEL_SIZE} ]]; do
+                echo "Waiting..."
+                sleep 1
+              done
+              ray status
+
+              export SERVED_MODEL_NAME=${MODEL_NAME}
+              export MODEL_NAME=${MODEL_DIR}
+
+              exec python3 -m vllm_tgis_adapter
+          env:
+            - name: RAY_PORT
+              value: "6379"
+            - name: RAY_ADDRESS
+              value: 127.0.0.1:6379
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: VLLM_CONFIG_ROOT
+              value: /tmp           
+            - name: DISTRIBUTED_EXECUTOR_BACKEND
+              value: "ray"
+            - name: DISABLE_CUSTOM_ALL_REDUCE
+              value: "true"
+            - name: MAX_SEQUENCE_LENGTH
+              value: "8192"
+            - name: MAX_NEW_TOKENS
+              value: "2048"
+            - name: MAX_BATCH_SIZE
+              value: "256"
+            - name: MAX_CONCURRENT_REQUESTS
+              value: "320"
+            - name: PORT
+              value: "8080"
+            - name: MAX_LOG_LEN
+              value: "100"
+            - name: HF_HUB_CACHE
+              value: /tmp
+          resources:
+            limits:
+              cpu: "16"
+              memory: 48Gi
+              nvidia.com/gpu: "1"
+            requests:
+              cpu: "8"
+              nvidia.com/gpu: "1"
+          volumeMounts:
+            - name: shm
+              mountPath: /dev/shm
+          livenessProbe:
+            failureThreshold: 3
+            periodSeconds: 30
+            successThreshold: 1
+            timeoutSeconds: 8
+            exec:
+              command:
+                - bash
+                - -c
+                - |
+                  # Check if the registered ray nodes count is the same as PIPELINE_PARALLEL_SIZE
+                  gpu_status=$(ray status | grep GPU)
+                  if [[ -z $gpu_status ]]; then
+                    echo "$1: GPU does not exist"
+                    exit 1
+                  fi
+
+                  used_gpu=$(echo "$gpu_status" | awk '{print $1}' | cut -d'/' -f1)
+                  reserved_gpu=$(echo "$gpu_status" | awk '{print $1}' | cut -d'/' -f2)
+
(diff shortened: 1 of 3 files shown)


config/runtimes/vllm-multinode-template.yaml
```diff
@@ -0,0 +1,256 @@
...
+                  reserved_gpu=$(echo "$gpu_status" | awk '{print $1}' | cut -d'/' -f2)
+
+                  # Determine health status based on GPU usage
+                  if [[ "$used_gpu" != "$reserved_gpu" ]]; then
+                    echo "$1: Unhealthy - Used: $used_gpu, Reserved: $reserved_gpu"
+                    exit 1
+                  fi
+
+                  # Check model health
+                  curl --silent --max-time 5 --fail-with-body http://localhost:8080/health
```
Xaenalt, line 153: Probes will likely encounter problems with large models, some models could take several minutes to respond

Jooho: right. it took around 12 minutes to load a 405B model. Therefore, I will be adding a health check (healthz) for the runtime during startup.

So I will add model healthz check in the StartupProbe.

*Startup Probe:*
- Purpose: To verify that ray is up and that the runtime returns the model as ready.
- Timeout: It will check for up to 20 minutes.

*Readiness Probe:*
- Purpose: After the Startup Probe succeeds, this probe checks if the environment is ready to serve requests.
- Checks Include:
  - Verifying the count of ray nodes and their readiness.
  - Checking the model's health status.
- Timeout: 30 secs

*Liveness Probe:*
- Purpose: To continuously monitor the running state of the ray nodes.

**Timeline Summary**
- Startup Probe: Runs for up to 20 minutes to ensure the model is loaded and ready.
- Readiness Probe: Activated after the Startup Probe passes; it checks the health of the nodes and the model.
- Liveness Probe: Monitors the health of the ray nodes throughout its lifecycle.

@Xaenalt wdyt?

config/runtimes/vllm-multinode-template.yaml
```diff
@@ -0,0 +1,256 @@
...
+              until [[ $(ray status | grep -c node_) -eq ${PIPELINE_PARALLEL_SIZE} ]]; do
+                echo "Waiting..."
+                sleep 1
+              done
+              ray status
+
+              export SERVED_MODEL_NAME=${MODEL_NAME}
+              export MODEL_NAME=${MODEL_DIR}
+
+              exec python3 -m vllm_tgis_adapter
```
Jooho, line 53: @vaibhavjainwiz @Xaenalt I used `vllm_tgis_adapter` based on IBM PoC. Runtime team is ok to use this?

config/runtimes/vllm-multinode-template.yaml
dtrifiro: Yeah, it's a good idea to use Startup and Readiness/Liveness probes. I think they're finally supported now:

https://issues.redhat.com/browse/SRVKS-1186

config/runtimes/vllm-multinode-template.yaml
dtrifiro: It'd be better to have `python -m vllm.entrypoints.openai.api_server` just like we have in the vLLM standalone template

config/runtimes/vllm-multinode-template.yaml
Jooho: multi node feature will be only supported with `rawDeployment`

config/runtimes/vllm-multinode-template.yaml
Jooho: ok I will use this

Jooho:
@Xaenalt @dtrifiro @vaibhavjainwiz could you please review this again? if there is no additional comments, please leave `/lgtlm`

config/runtimes/vllm-multinode-template.yaml
```diff
@@ -0,0 +1,259 @@
...
+            - name: MAX_BATCH_SIZE
+              value: "256"
+            - name: MAX_CONCURRENT_REQUESTS
+              value: "320"
+            - name: PORT
+              value: "8080"
+            - name: MAX_LOG_LEN
+              value: "100"
+            - name: HF_HUB_CACHE
+              value: /tmp
```
dtrifiro, line 82: We should remove all of these env vars, and only set `HF_HOME` instead (see existing vllm template for comparison).

`DISTRIBUTED_EXECUTOR_BACKEND` should be set as arg and port should be added aswell :

```
python3 -m vllm.entrypoints.openai.api_server <...> --port 8080 --distributed-executor-backend ray

```

config/runtimes/vllm-multinode-template.yaml
```diff
@@ -0,0 +1,259 @@
...
+                    fi
+                    echo "$SECONDS seconds elapsed: Still waiting for Global Control Service(GCS) to be ready."
+                    echo "For troubleshooting, refer to the FAQ at https://docs.ray.io/en/master/cluster/kubernetes/troubleshooting/troubleshooting.html#kuberay-troubleshootin-guides"
+                  fi
+                  
+                  sleep 5
+                done
+
+                echo "Attempting to connect to Ray cluster at $RAY_HEAD_ADDRESS ..."
+                RAY_HEAD_ADDRESS="${HEAD_SVC}.${POD_NAMESPACE}.svc.cluster.local:6379"
```
dtrifiro, line 208: Perhaps RAY_HEAD_ADDRESS should be set before the `echo`?

config/runtimes/vllm-multinode-template.yaml
```diff
@@ -0,0 +1,259 @@
...
+              name: http
+              protocol: TCP
+      volumes:
+        - name: shm
+          emptyDir:
+            medium: Memory
+            sizeLimit: 12Gi    
+      workerSpec:
+        pipelineParallelSize: 2
+        tensorParallelSize: 1
```
dtrifiro, line 179: Does this need to be hardcoded? Should this be a variable in the openshift template?

config/runtimes/vllm-multinode-template.yaml
```diff
@@ -0,0 +1,259 @@
...
+                  fieldRef:
+                    fieldPath: metadata.namespace
+            resources:
+              limits:
+                cpu: "16"
+                memory: 48Gi
+                nvidia.com/gpu: "1"
+              requests:
+                cpu: "8"
+                nvidia.com/gpu: "1"
```
dtrifiro, line 226: Same question as above here regarding the cpu/memory resources/requests

config/runtimes/vllm-multinode-template.yaml
Jooho: I hit an issue when I don't use `VLLM_CONFIG_ROOT` with old vllm image (quay.io/opendatahub/vllm@sha256:7f19dde68eb47abeea155f0d68d4e708f4d93cc91fc632b7a5a0de181d8d193b).  I don't have the error message now but I will leave for now. when kserve is ready to test, I am going to double check this parameter again. If error still happens, I will keep. The rest of parameters, I will remove it.

config/runtimes/vllm-multinode-template.yaml
Jooho: good catch!. updated.

config/runtimes/vllm-multinode-template.yaml
```diff
@@ -0,0 +1,259 @@
...
+        containers:
+          - name: worker-container
+            image: $(vllm-image)
+            command: ["bash", "-c"]
+            args:
+              - |
+                SECONDS=0
+
+                while true; do              
+                  if (( SECONDS <= 120 )); then
```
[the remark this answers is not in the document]
Jooho: The `SECONDS` variable in Bash is a special variable that automatically increments, representing the number of seconds that have elapsed since the script started running. Therefore, it’s not necessary to manually increment SECONDS in each iteration—it will automatically accumulate the elapsed time.

config/runtimes/vllm-multinode-template.yaml
dtrifiro: Ah, looking at the code, it looks like it might be enough to set `VLLM_NO_USAGE_STATS=1`

config/runtimes/vllm-multinode-template.yaml
Jooho: my answer would be the same too here. Upstream also have a lot of discussion on this as well but it turned out it really depends on user situation. For user who want to just test, this looks very high but for user who want to deploy a really big model, it is not enough. So it should be configurable by user with InferenceService.
However I think gpu part should be removed for now because it is not overwritten by inferenceservice. it will be added if the gpu type is different.

Jooho:
@dtrifiro thx for your review, I followed up your comments. please review again.

Jooho:
@dtrifiro thx for your review, I followed up your comments. please review again.

config/runtimes/vllm-multinode-template.yaml
Jooho: I removed the `nvidia.com/gpu`  so user must set this by isvc or they have to edit servingruntime. If they don't set , it will fail.

terrytangyuan: > gpu part should be removed for now because it is not overwritten by inferenceservice.  It will be added if the gpu type is different.

Is this a bug? 

If that's the case, then it makes sense to remove GPU specifications here.

terrytangyuan: The logic to merge/patch resource spec is here: https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/utils/utils.go#L403

config/runtimes/vllm-multinode-template.yaml
```diff
@@ -0,0 +1,241 @@
...
+kind: Template
+metadata:
+  labels:
+    opendatahub.io/dashboard: "true"
+    opendatahub.io/ootb: "true"
+  annotations:
+    description: vLLM is a high-throughput and memory-efficient inference and serving engine for LLMs
+    openshift.io/display-name: vLLM ServingRuntime for KServe
+    openshift.io/provider-display-name: Red Hat, Inc.
+    tags: rhods,rhoai,kserve,servingruntime
```
spolti, line 11: does it make sense to add the multi-node tag?

config/runtimes/vllm-multinode-template.yaml
```diff
@@ -0,0 +1,241 @@
...
+    template.openshift.io/documentation-url: https://github.com/opendatahub-io/vllm
+    template.openshift.io/long-description: This template defines resources needed to deploy vLLM ServingRuntime with KServe in Red Hat OpenShift AI
```
spolti, line 13: maybe update the descriptio to match the multi-node purpose?

config/runtimes/vllm-multinode-template.yaml
```diff
@@ -0,0 +1,241 @@
...
+    opendatahub.io/modelServingSupport: '["single"]'
+    opendatahub.io/apiProtocol: "REST"
+  name: vllm-multinode-runtime-template
+objects:
+  - apiVersion: serving.kserve.io/v1alpha1
+    kind: ServingRuntime
+    metadata:
+      name: vllm-multinode-runtime
+      annotations:
+        openshift.io/display-name: vLLM ServingRuntime for KServe
```
spolti, line 23: multi-node?

config/runtimes/vllm-multinode-template.yaml
Jooho: udpated

config/runtimes/vllm-multinode-template.yaml
Jooho: updated

spolti (approved):

PR #288, opened 2024-11-05
Create webhook for NIM enablement

xieshenzh:
Create the validation webhook for NIM Account.

## Description

1. Created the webhook
2. Updated the manifests to include the webhook
3. Check the DSC flag to decide whether to enable the webhook

## How Has This Been Tested?

1. Deploy the controller in Openshift
4. Create an Account CR.
2.1 Check if the Account is created when the api key secret is not created
2.2 Check if the Account is created when there is already an Account CR in the namespace
2.3 Check if the Account is created when the api key secret is created and there is no other Account in the namespace
5. Update an Account CR
3.1 Check if the Account is updated when updating the api key secret and the new api key secret is not created
3.2 Check if the Account is updated when updating the api key secret and the new api key secret is created

## Merge criteria:

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -116,7 +116,8 @@ var _ = BeforeSuite(func() {
 	By("Bootstrapping test environment")
 	envTest = &envtest.Environment{
 		CRDInstallOptions: envtest.CRDInstallOptions{
-			Paths:              []string{filepath.Join("..", "config", "crd", "external")},
+			Paths: []string{filepath.Join("..", "config", "crd", "external"),
+				filepath.Join("..", "config", "crd", "bases")},
 			ErrorIfPathMissing: true,
 			CleanUpAfterUse:    false,
 		},
--- a/controllers/webhook/nim_account_webhook.go
+++ b/controllers/webhook/nim_account_webhook.go
@@ -1,0 +1,64 @@
+package webhook
+
+import (
+	"context"
+	"fmt"
+
+	nimv1 "github.com/opendatahub-io/odh-model-controller/api/nim/v1"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// +kubebuilder:webhook:admissionReviewVersions=v1,path=/validate-nim-opendatahub-io-v1-account,mutating=false,failurePolicy=fail,groups=nim.opendatahub.io,resources=accounts,verbs=create;update,versions=v1,name=validating.nim.account.odh-model-controller.opendatahub.io,sideEffects=None
+
+type nimAccountValidator struct {
+	client client.Client
+}
+
+func NewNimAccountValidator(client client.Client) admission.CustomValidator {
+	return &nimAccountValidator{client: client}
+}
+
+func (v *nimAccountValidator) ValidateCreate(ctx context.Context, obj runtime.Object) (warnings admission.Warnings, err error) {
+	account := obj.(*nimv1.Account)
+
+	log := logf.FromContext(ctx).WithName("NIMAccountValidatingWebhook").
+		WithValues("namespace", account.Namespace, "account", account.Name)
+	log.Info("Validating NIM Account creation")
+
+	err = v.verifySingletonInNamespace(ctx, account)
+	if err != nil {
+		log.Error(err, "Rejecting NIM Account creation because checking singleton didn't pass")
+		return nil, err
+	}
+
+	return nil, nil
+}
+
+func (v *nimAccountValidator) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) (warnings admission.Warnings, err error) {
+	// For update, nothing needs to be validated
+	return nil, nil
+}
+
+func (v *nimAccountValidator) ValidateDelete(ctx context.Context, obj runtime.Object) (warnings admission.Warnings, err error) {
+	// For deletion, nothing needs to be validated
+	return nil, nil
+}
+
+func (v *nimAccountValidator) verifySingletonInNamespace(ctx context.Context, account *nimv1.Account) error {
+	accountList := nimv1.AccountList{}
+	err := v.client.List(ctx,
+		&accountList,
+		client.InNamespace(account.Namespace))
+	if err != nil {
+		return fmt.Errorf("failed to verify if there are existing Accounts with err: %s", err.Error())
+	}
+
+	if len(accountList.Items) > 0 {
+		return fmt.Errorf("rejecting creation of Account %s in namespace %s because there is already an Account created in the namespace", account.Name, account.Namespace)
+	}
+	return nil
+}
--- a/main.go
+++ b/main.go
@@ -44,6 +44,8 @@ import (
 
 	"github.com/opendatahub-io/odh-model-controller/controllers"
 	"github.com/opendatahub-io/odh-model-controller/controllers/utils"
+
+	nimv1 "github.com/opendatahub-io/odh-model-controller/api/nim/v1"
 	//+kubebuilder:scaffold:imports
 )
 
@@ -216,6 +218,14 @@ func main() {
 
 	} else {
 		setupLog.Info("Skipping setup of Knative Service validating/mutating Webhook, because KServe Serverless setup seems to be disabled in the DataScienceCluster resource.")
+	}
+
+	if err := builder.WebhookManagedBy(mgr).
+		For(&nimv1.Account{}).
+		WithValidator(webhook.NewNimAccountValidator(mgr.GetClient())).
+		Complete(); err != nil {
+		setupLog.Error(err, "unable to setup NIM validating Webhook")
+		os.Exit(1)
 	}
 
 	//+kubebuilder:scaffold:builder
--- a/PROJECT
+++ b/PROJECT
@@ -13,3 +13,6 @@ resources:
   kind: Account
   path: github.com/opendatahub-io/odh-model-controller/api/nim/v1
   version: v1
+  webhooks:
+    validation: true
+    webhookVersion: v1
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -23,3 +23,23 @@ webhooks:
     resources:
     - services
   sideEffects: None
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /validate-nim-opendatahub-io-v1-account
+  failurePolicy: Fail
+  name: validating.nim.account.odh-model-controller.opendatahub.io
+  rules:
+  - apiGroups:
+    - nim.opendatahub.io
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - accounts
+  sideEffects: None
--- a/config/webhook/webhook_patch.yaml
+++ b/config/webhook/webhook_patch.yaml
@@ -13,3 +13,7 @@ webhooks:
       matchExpressions:
         - key: serving.kserve.io/inferenceservice
           operator: Exists
+  - name: validating.nim.account.odh-model-controller.opendatahub.io
+    clientConfig:
+      service:
+        name: odh-model-controller-webhook-service
--- a/controllers/webhook_nim_account_validator_test.go
+++ b/controllers/webhook_nim_account_validator_test.go
@@ -1,0 +1,97 @@
+package controllers
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	nimv1 "github.com/opendatahub-io/odh-model-controller/api/nim/v1"
+	"github.com/opendatahub-io/odh-model-controller/controllers/webhook"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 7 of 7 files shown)


main.go
```diff
@@ -218,6 +220,25 @@ func main() {
...
+	if nimEnabled {
+		nimValidatorWebhookSetupErr := builder.WebhookManagedBy(mgr).
+			For(&nimv1.Account{}).
+			WithValidator(webhook.NewNimAccountValidator(mgr.GetClient())).
+			Complete()
+
+		if nimValidatorWebhookSetupErr != nil {
+			setupLog.Error(err, "unable to setup NIM validating Webhook")
+			os.Exit(1)
+		}
```
terrytangyuan, line 237: If `nimEnabledErr != nil`, we can skip this additional check

Jooho:
Could you please write tests for this webhook?

xieshenzh:
> Could you please write tests for this webhook?

I did: https://github.com/opendatahub-io/odh-model-controller/pull/288/files#diff-a276b8cfa9be48d35d7eb0c0ed9d8a2d4dea589741656fb6bac43c03dd2652a2

controllers/webhook/nim_account_webhook.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (v *nimAccountValidator) verifyAPIKeySecret(ctx context.Context, oldAccount, newAccount *nimv1.Account) error {
+	if oldAccount == nil || !reflect.DeepEqual(oldAccount.Spec.APIKeySecret, newAccount.Spec.APIKeySecret) {
+		if newAccount.Spec.APIKeySecret.Name == "" {
+			msg := "name of the Secret for the API key is invalid"
+			return field.Invalid(field.NewPath("spec").Child("apiKeySecret").Child("name"), newAccount.Spec.APIKeySecret.Name, msg)
+		}
+		name := newAccount.Spec.APIKeySecret.Name
+		namespace := newAccount.Spec.APIKeySecret.Namespace
```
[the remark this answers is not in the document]
xieshenzh: Technically, the namespace of the Secret can be different from the namespace of the Account.
If there is a requirement to disallow this scenario, this webhook should reject Account creation or update in this case.
But I have not received such requirement so far.

controllers/webhook/nim_account_webhook.go
```diff
@@ -0,0 +1,118 @@
...
+		if namespace == "" {
```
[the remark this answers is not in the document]
xieshenzh: Yes. 
We use type `ObjectReference` for `apiKeySecret`.
In `ObjectReference`, `namespace` is an optional field.

So it is possible that users don't set `namespace`.
In this case, the value will be `""`.

If namesapce is set in the `apiKeySecret`, it will be used.
Otherwise, the code will assume the `Secret` is in the same namespace as the Account.

controllers/utils/utils.go
```diff
@@ -103,39 +104,46 @@ func VerifyIfComponentIsEnabled(ctx context.Context, cli client.Client, componen
...
+			componentName, obj, errComp)
+	}
+
+	kserve, _, errKserve := unstructured.NestedString(obj.Object, kserveFields...)
+	if errKserve != nil {
+		return false, fmt.Errorf("failed to retrieve the component [%s] status from %+v. %w",
+			componentName, obj, errKserve)
+	}
+
+	return (comp == "Managed" || comp == "Unmanaged") && kserve == "Managed", nil
```
zdtsw, line 144: by moving this check into a new func, nim should treat Managed and Unmanaged the same way?
I think in "to be defined" API PR , the only valid state is Managed and Remove on this field.
serving does have logic on Unmanaged as well.

controllers/webhook/nim_account_webhook.go
```diff
@@ -0,0 +1,118 @@
...
+func NewNimAccountValidator(client client.Client) admission.CustomValidator {
+	return &nimAccountValidator{client: client}
+}
+
+func (v *nimAccountValidator) ValidateCreate(ctx context.Context, obj runtime.Object) (warnings admission.Warnings, err error) {
+	log := logf.FromContext(ctx).WithName("NIMAccountValidatingWebhook")
+
+	account := obj.(*nimv1.Account)
+
+	log = log.WithValues("namespace", account.Namespace, "account", account.Name)
```
zdtsw, line 35: why not just 
```suggestion
	account := obj.(*nimv1.Account)
	log := logf.FromContext(ctx).WithName("NIMAccountValidatingWebhook").WithValues("namespace", account.Namespace, "account", account.Name)
```

controllers/suite_test.go
```diff
@@ -116,7 +116,8 @@ var _ = BeforeSuite(func() {
 	By("Bootstrapping test environment")
 	envTest = &envtest.Environment{
 		CRDInstallOptions: envtest.CRDInstallOptions{
-			Paths:              []string{filepath.Join("..", "config", "crd", "external")},
+			Paths: []string{filepath.Join("..", "config", "crd", "external"),
+				filepath.Join("..", "config", "crd", "bases")},
```
zdtsw, line 120: tbh, i would think this new CRD to be shipped out by kserve as an overlay not by odh-model-controller.
say, kserve is disabled, modelmesh is enabled, CRD would still be installed in the cluster, should it? 

or not include "crd/bases" from "base/kustomization.yaml". 
instead, in Operator to create it on kserve demand.

xieshenzh: @etirelli @israel-hdez @lburgazzoli @tomerFI @jooho @mpaulgreen 
Any thoughts on the suggestions?

main.go
```diff
@@ -218,6 +220,25 @@ func main() {
 		setupLog.Info("Skipping setup of Knative Service validating/mutating Webhook, because KServe Serverless setup seems to be disabled in the DataScienceCluster resource.")
 	}
 
+	nimEnabled, nimEnabledErr := utils.VerifyIfComponentIsEnabled(context.Background(), mgr.GetClient(), utils.KServeWithNIMComponent)
```
xieshenzh, line 223: @etirelli @israel-hdez @lburgazzoli @TomerFi @mpaulgreen
I saw the other webhook `ksvcValidatorWebhook` checks the DSC flag before installation.
So I check the nim flag before installing the nim webhook.
Is this expected?

controllers/suite_test.go
mpaulgreen: Here are some legacy comments that can drive the decision. Please feel free to override me. https://issues.redhat.com/projects/NVPE/issues/NVPE-4?filter=allissues and from a slack thread - https://redhat-internal.slack.com/archives/C07CZDA6DE1/p1730391034630149?thread_ts=1730389624.036899&cid=C07CZDA6DE1

main.go
mpaulgreen: same as I mentioned here - https://github.com/opendatahub-io/odh-model-controller/pull/288#discussion_r1832916829

controllers/webhook/nim_account_webhook.go
xieshenzh: done

main.go
xieshenzh: done

main.go
xieshenzh: Removed the checking for the DSC flag: https://redhat-internal.slack.com/archives/C07CZDA6DE1/p1731018233090329?thread_ts=1730969824.426219&cid=C07CZDA6DE1

controllers/utils/utils.go
xieshenzh: The checking for the DSC flag is removed as requested: https://redhat-internal.slack.com/archives/C07CZDA6DE1/p1731018233090329?thread_ts=1730969824.426219&cid=C07CZDA6DE1

controllers/suite_test.go
zdtsw: i do not have objection to place CRD in such a way

controllers/webhook/nim_account_webhook.go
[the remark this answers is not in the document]
Jooho: odh-model-controller can access to all namespaces and also so can get secret but it is for internal usage for `odh-model-controller`. But if it is available to access all secrets in the cluster by this cr, I think it is not a good idea for security reasons. @etirelli @terrytangyuan @israel-hdez wdyt?

controllers/suite_test.go
terrytangyuan: No strong preference on this. I think it's fine keep it as is.

controllers/webhook/nim_account_webhook.go
```diff
@@ -0,0 +1,116 @@
...
+				return fmt.Errorf("failed to verify the API key Secret with err: %s", err.Error())
+			}
+		}
+	}
+	return nil
+}
+
+func (v *nimAccountValidator) verifySingletonInNamespace(ctx context.Context, account *nimv1.Account) error {
+	accountList := nimv1.AccountList{}
+	err := v.client.List(ctx,
```
terrytangyuan, line 105: Can we do a get instead of a list call?

xieshenzh: I am not sure how to do a GET without knowing the name and namespace of the object.

Could you please be more specific, and maybe give me an example? Thanks!

terrytangyuan: Oh I see. Then you'll need LIST.

Optional: It might be more efficient if you have any label selectors associated so that you don't have to scan the entire namespace.

controllers/webhook/nim_account_webhook.go
```diff
@@ -0,0 +1,116 @@
...
+
+	return nil, nil
+}
+
+func (v *nimAccountValidator) ValidateDelete(ctx context.Context, obj runtime.Object) (warnings admission.Warnings, err error) {
+	// For deletion, nothing needs to be validated
+	return nil, nil
+}
+
+func (v *nimAccountValidator) verifyAPIKeySecret(ctx context.Context, oldAccount, newAccount *nimv1.Account) error {
```
TomerFi, line 73: @xieshenzh @etirelli @mpaulgreen 
Why are we checking the API key Secret with the admission webhook?

Verifying only one Account exists in a namespace was the only requirement of the webhook. See [ADR](https://docs.google.com/document/d/1gvWjshALD3Ec6ifwkhqNkla_9dZGlsgyaTkvocuaN8U/edit?tab=t.0):

image

This webhook duplicates code that should (and does) exist in the controller. See #289.

TomerFi (changes requested):

controllers/webhook/nim_account_webhook.go
TomerFi: Label selectors won't help; we're trying to prevent two Accounts from existing in the same namespace because, for each Account, we're creating a _ConfigMap_ a _Template_ and a _Secret_ with a static naming convention. We don't want the two Accounts to overwrite each other's resources.

If we depend on labels, users can potentially create an Account without a label to work around our webhook.

See https://github.com/opendatahub-io/odh-model-controller/pull/289#discussion_r1835717365.

main.go
```diff
@@ -84,6 +86,9 @@ func init() { //nolint:gochecknoinits //reason this way we ensure schemes are al
 // +kubebuilder:rbac:groups=authorino.kuadrant.io,resources=authconfigs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=datasciencecluster.opendatahub.io,resources=datascienceclusters,verbs=get;list;watch
 // +kubebuilder:rbac:groups=dscinitialization.opendatahub.io,resources=dscinitializations,verbs=get;list;watch
+// +kubebuilder:rbac:groups=nim.opendatahub.io,resources=accounts,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=nim.opendatahub.io,resources=accounts/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=nim.opendatahub.io,resources=accounts/finalizers,verbs=update
```
TomerFi, line 91: Does the webhook require anything more than _get;list;watch_ for account resources?
I don't think the webhook needs the status, and we don't have a finalizer.

controllers/webhook/nim_account_webhook.go
terrytangyuan: That makes sense. Thanks for clarifying. Then this change looks good.

main.go
xieshenzh: These permissions are what a controller typically has for a new API.
I was supposed to add these in the PR for the API.

TomerFi: In general, I think these should be added with the controller, not the API.
Regardless, these are more than what we need. For the accounts, we only need _get_, _list_, and _watch_, which already exist in the controller PR. I mistakenly left out the status permissions, but this was fixed. As for the finalizer permissions, we don't use finalizers.

controllers/webhook/nim_account_webhook.go
TomerFi: But the controller also checks if the Secret exists—that's the duplication. The controller will update a status condition on the Account, indicating that the API key secret doesn't exist. If users create it after seeing the status, everything will work.

The API mandates the `spec.apiKeySecret`; shouldn't this be enough for users to understand that they must create a Secret?

controllers/webhook/nim_account_webhook.go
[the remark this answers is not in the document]
xieshenzh: @Jooho The webhook is updated. It won't check the Secret anymore.
If you still have security concerns on the usage of the Secret. 
Please add your comments and requests in PR #289 instead. 
Thanks.

controllers/webhook/nim_account_webhook.go
xieshenzh: Checks for the Secret is deleted.

xieshenzh:
@Jooho I don't think there is any outstanding concerns on this PR. 
Could you please add /ok-to-test?

TomerFi (approved):

controllers/suite_test.go
israel-hdez: I'm OK with having the CRD in this repo.

israel-hdez (approved):
/lgtm

PR #289, opened 2024-11-07
Task/NVPE-10: Create NIM Account controller

TomerFi:
## Description

- Added a controller reconciling for NIM accounts.
- Added test cases for the controller including mocks for NVIDIA NIM API.
- Added a "hack" script for troubelshooting NVIDIA API access.

Jira: https://issues.redhat.com/browse/NVPE-10

## How Has This Been Tested?

Tested manully on my station usign the script and helper code.
EDIT (Nov. 12):
Also tested against a live cluster, tests described in the following [google doc](https://docs.google.com/document/d/1GlYxgkyYdIAs0mK2GeuHv3xCpUAQNl34__pxB67P1pw/edit?usp=sharing).

## Merge criteria:

### DIFF
--- a/controllers/constants/constants.go
+++ b/controllers/constants/constants.go
@@ -71,3 +71,10 @@ const (
 const (
 	NoSuitableRuntimeError = "not found error: no suitable runtime found."
 )
+
+// NIM
+const (
+	NimDataConfigMapName   = "nvidia-nim-data"
+	NimRuntimeTemplateName = "nvidia-nim-serving-template"
+	NimPullSecretName      = "nvidia-nim-image-pull"
+)
--- a/controllers/constants/constants.go
+++ b/controllers/constants/constants.go
@@ -74,7 +74,8 @@ const (
 
 // NIM
 const (
-	NimDataConfigMapName   = "nvidia-nim-data"
-	NimRuntimeTemplateName = "nvidia-nim-serving-template"
-	NimPullSecretName      = "nvidia-nim-image-pull"
+	NimApplyConfigFieldManager = "nim-account-controller"
+	NimDataConfigMapName       = "nvidia-nim-data"
+	NimRuntimeTemplateName     = "nvidia-nim-serving-template"
+	NimPullSecretName          = "nvidia-nim-image-pull"
 )
--- a/controllers/nim_account_controller.go
+++ b/controllers/nim_account_controller.go
@@ -90,7 +90,6 @@ func (r *NimAccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	if err := r.Client.Get(ctx, req.NamespacedName, account); err != nil {
 		if k8serrors.IsNotFound(err) {
 			logger.V(1).Info("account deleted")
-			r.cleanupResources(ctx, req.Namespace)
 		} else {
 			logger.V(1).Error(err, "failed to fetch object")
 		}
@@ -98,7 +97,7 @@ func (r *NimAccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	}
 
 	if !account.DeletionTimestamp.IsZero() {
-		logger.V(1).Info("account being deleted, cleanups done")
+		logger.V(1).Info("account being deleted")
 		return ctrl.Result{}, nil
 	}
 
@@ -135,8 +134,6 @@ func (r *NimAccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 			meta.SetStatusCondition(&targetStatus.Conditions, makeConfigMapFailureCondition(account.Generation, msg))
 			meta.SetStatusCondition(&targetStatus.Conditions, makeTemplateFailureCondition(account.Generation, msg))
 			meta.SetStatusCondition(&targetStatus.Conditions, makePullSecretFailureCondition(account.Generation, msg))
-
-			r.cleanupResources(ctx, account.Namespace)
 		} else {
 			logger.V(1).Error(err, "failed to fetch api key secret")
 		}
@@ -336,37 +333,6 @@ func (r *NimAccountReconciler) updateStatus(ctx context.Context, subject types.N
 		account.Status = *status.DeepCopy()
 		if err = r.Client.Status().Update(ctx, account); err != nil {
 			logger.Error(err, "failed to update account status")
-		}
-	}
-}
-
-// cleanupResources is used for deleting the integration related resources (configmap, template, pull secret)
-func (r *NimAccountReconciler) cleanupResources(ctx context.Context, namespace string) {
-	logger := log.FromContext(ctx)
-	logger.V(1).Info("cleaning up")
-	delObjs := []client.Object{
-		&corev1.Secret{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      constants.NimPullSecretName,
-				Namespace: namespace,
-			}},
-		&templatev1.Template{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      constants.NimRuntimeTemplateName,
-				Namespace: namespace,
-			}},
-		&corev1.ConfigMap{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      constants.NimDataConfigMapName,
-				Namespace: namespace,
-			}},
-	}
-
-	for _, obj := range delObjs {
-		if err := r.Client.Delete(ctx, obj); err != nil {
-			if !k8serrors.IsNotFound(err) {
-				logger.Error(err, fmt.Sprintf("failed to delete %s", obj.GetObjectKind()))
-			}
 		}
 	}
 }
--- a/controllers/nim_account_controller.go
+++ b/controllers/nim_account_controller.go
@@ -332,8 +332,8 @@ func (r *NimAccountReconciler) reconcileNimPullSecret(
 
 	secretCfg := ssacorev1.
 		Secret(constants.NimPullSecretName, namespace).
-		WithData(map[string][]byte{".dockercfg": credsJson}).
-		WithType(corev1.SecretTypeDockercfg).
+		WithData(map[string][]byte{corev1.DockerConfigJsonKey: credsJson}).
+		WithType(corev1.SecretTypeDockerConfigJson).
 		WithOwnerReferences(ownerCfg).
 		WithLabels(labels)
 
--- a/controllers/nim_account_controller.go
+++ b/controllers/nim_account_controller.go
@@ -41,6 +41,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"strings"
 )
 
 type (
@@ -53,6 +54,10 @@ type (
 
 const (
 	apiKeySpecPath = "spec.apiKeySecret.name"
+)
+
+var (
+	labels = map[string]string{"opendatahub.io/managed": "true"}
 )
 
 func (r *NimAccountReconciler) SetupWithManager(mgr ctrl.Manager, ctx context.Context) error {
@@ -134,7 +139,7 @@ func (r *NimAccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	if err := r.Client.Get(ctx, apiKeySecretSubject, apiKeySecret); err != nil {
 		var msg string
 		if k8serrors.IsNotFound(err) {
-			msg = "api key secret was deleted"
+			msg = "api key secret not found"
 			logger.Info(msg)
 		} else {
 			msg = "failed to fetch api key secret"
@@ -157,7 +162,7 @@ func (r *NimAccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		r.cleanupResources(ctx, account)
 		return ctrl.Result{}, err
 	}
-	apiKeyStr := string(apiKeyBytes)
+	apiKeyStr := strings.TrimSpace(string(apiKeyBytes))
 	gotApiKey := "got api key"
 	logger.V(1).Info(gotApiKey)
 	meta.SetStatusCondition(&targetStatus.Conditions, makeAccountFailureCondition(account.Generation, gotApiKey))
@@ -263,7 +268,8 @@ func (r *NimAccountReconciler) reconcileNimConfig(
 
 	cmCfg := ssacorev1.ConfigMap(constants.NimDataConfigMapName, namespace).
 		WithData(data).
-		WithOwnerReferences(ownerCfg)
+		WithOwnerReferences(ownerCfg).
+		WithLabels(labels)
 
 	cm, err := r.KClient.CoreV1().ConfigMaps(namespace).
 		Apply(ctx, cmCfg, metav1.ApplyOptions{FieldManager: constants.NimApplyConfigFieldManager, Force: true})
@@ -292,6 +298,8 @@ func (r *NimAccountReconciler) reconcileRuntimeTemplate(ctx context.Context, acc
 			"opendatahub.io/apiProtocol":         "REST",
 			"opendatahub.io/modelServingSupport": "[\"single\"]",
 		}
+
+		template.Labels = labels
 
 		template.Objects = []runtime.RawExtension{
 			{Object: utils.GetNimServingRuntimeTemplate()},
@@ -327,7 +335,7 @@ func (r *NimAccountReconciler) reconcileNimPullSecret(
 		WithData(map[string][]byte{".dockercfg": credsJson}).
 		WithType(corev1.SecretTypeDockercfg).
 		WithOwnerReferences(ownerCfg).
-		WithLabels(map[string]string{"opendatahub.io/managed": "true"})
+		WithLabels(labels)
 
 	secret, err := r.KClient.CoreV1().Secrets(namespace).
 		Apply(ctx, secretCfg, metav1.ApplyOptions{FieldManager: constants.NimApplyConfigFieldManager, Force: true})
--- a/controllers/nim_account_controller.go
+++ b/controllers/nim_account_controller.go
@@ -112,7 +112,7 @@ func (r *NimAccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	targetStatus := &v1.AccountStatus{
 		Conditions: []metav1.Condition{
 			// initial unknown status
-			makeAccountSuccessfulCondition(account.Generation, initialMsg),
+			makeAccountUnknownCondition(account.Generation, initialMsg),
 			makeApiKeyUnknownCondition(account.Generation, initialMsg),
 			makeConfigMapUnknownCondition(account.Generation, initialMsg),
 			makeTemplateUnknownCondition(account.Generation, initialMsg),
@@ -421,6 +421,10 @@ func makeAccountFailureCondition(gen int64, msg string) metav1.Condition {
 
 func makeAccountSuccessfulCondition(gen int64, msg string) metav1.Condition {
 	return utils.MakeNimCondition(utils.NimConditionAccountStatus, metav1.ConditionTrue, gen, "AccountSuccessful", msg)
+}
+
+func makeAccountUnknownCondition(gen int64, msg string) metav1.Condition {
+	return utils.MakeNimCondition(utils.NimConditionAccountStatus, metav1.ConditionUnknown, gen, "AccountNotReconciled", msg)
 }
 
 // API KEY VALIDATION CONDITIONS
--- a/controllers/nim_account_controller.go
+++ b/controllers/nim_account_controller.go
@@ -1,0 +1,424 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/go-logr/logr"
+	"github.com/kuadrant/authorino/pkg/log"
+	"github.com/opendatahub-io/odh-model-controller/api/nim/v1"
+	"github.com/opendatahub-io/odh-model-controller/controllers/constants"
+	"github.com/opendatahub-io/odh-model-controller/controllers/utils"
+	templatev1 "github.com/openshift/api/template/v1"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/reference"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type (
+	NimAccountReconciler struct {
+		client.Client
+		Log logr.Logger
+	}
+)
+
+const (
+	apiKeySpecPath = "spec.apiKeySecret.name"
+)
+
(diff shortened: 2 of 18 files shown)


controllers/nim_account_controller.go
```diff
@@ -0,0 +1,424 @@
...
+
+func (r *NimAccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := r.Log.WithValues("Account", req.Name, "namespace", req.Namespace)
+	ctx = log.IntoContext(ctx, logger)
+
+	account := &v1.Account{}
+	if err := r.Client.Get(ctx, req.NamespacedName, account); err != nil {
+		if k8serrors.IsNotFound(err) {
+			logger.V(1).Info("account deleted")
+			r.cleanupResources(ctx, req.Namespace)
```
lburgazzoli, line 93: it seems that the resources affected by the cleanup have the `Account` as owner, I guess they get cleaned up automatically by k8s gc, so may not be needed ?

controllers/nim_account_controller.go
```diff
@@ -0,0 +1,424 @@
...
+	// initial account status condition and set initial api key status condition
+	meta.SetStatusCondition(&targetStatus.Conditions, makeAccountFailureCondition(account.Generation, "api key secret not available"))
+	meta.SetStatusCondition(&targetStatus.Conditions, makeApiKeyFailureCondition(account.Generation, "api key secret not available"))
+	// initial unknown for the resources conditions
+	meta.SetStatusCondition(&targetStatus.Conditions, makeConfigMapUnknownCondition(account.Generation, "not reconciled yet"))
+	meta.SetStatusCondition(&targetStatus.Conditions, makeTemplateUnknownCondition(account.Generation, "not reconciled yet"))
+	meta.SetStatusCondition(&targetStatus.Conditions, makePullSecretUnknownCondition(account.Generation, "not reconciled yet"))
+
+	defer func() {
+		r.updateStatus(ctx, req.NamespacedName, *targetStatus)
```
lburgazzoli, line 118: given the update status is updated every time the function return, it seems there is a case where the conditions are reporting wrong information. i.e. in the case the function returns as a consequence of `utils.GetAvailableNimRuntimes()` reporting an error, then the condition would report i.e. "api key secret not available", which may not be true.

not sure if that is expected

controllers/nim_account_controller.go
```diff
@@ -0,0 +1,424 @@
...
+
+func (r *NimAccountReconciler) reconcileNimConfig(ctx context.Context, account *v1.Account, apiKey string, runtimes []utils.NimRuntime) (*corev1.ConfigMap, error) {
+	cmap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      constants.NimDataConfigMapName,
+			Namespace: account.Namespace,
+		},
+	}
+
+	if _, err := controllerutil.CreateOrUpdate(ctx, r.Client, cmap, func() error {
```
lburgazzoli, line 238: not sure if it is applicable, but eventually you can use Server Side Apply instead of `CreateOrUpdate` which may reduce calls and conflicts

controllers/nim_account_controller.go
TomerFi: Resolved in https://github.com/opendatahub-io/odh-model-controller/pull/289/commits/7929b53bb3cbda70594b2aa0169f26d34bf92722.

controllers/nim_account_controller.go
TomerFi: Resolved in https://github.com/opendatahub-io/odh-model-controller/pull/289/commits/8f87716a9eec21bf16ce5cf1143cce3ad9295252.

main.go
```diff
@@ -84,6 +84,8 @@ func init() { //nolint:gochecknoinits //reason this way we ensure schemes are al
 // +kubebuilder:rbac:groups=authorino.kuadrant.io,resources=authconfigs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=datasciencecluster.opendatahub.io,resources=datascienceclusters,verbs=get;list;watch
 // +kubebuilder:rbac:groups=dscinitialization.opendatahub.io,resources=dscinitializations,verbs=get;list;watch
+// +kubebuilder:rbac:groups=nim.opendatahub.io,resources=accounts,verbs=get;list;watch;update
```
xieshenzh, line 87: Usually, a controller needs rbac:

```suggestion
// +kubebuilder:rbac:groups=nim.opendatahub.io,resources=accounts,verbs=get;list;watch;create;update;patch;delete
// +kubebuilder:rbac:groups=nim.opendatahub.io,resources=accounts/status,verbs=get;update;patch
// +kubebuilder:rbac:groups=nim.opendatahub.io,resources=accounts/finalizers,verbs=update
```

TomerFi: Is this required?
We'e not creating, patching, or deleting accounts.

controllers/nim_account_controller.go
TomerFi: Resolved in https://github.com/opendatahub-io/odh-model-controller/pull/289/commits/de002501c908763dce09e79acc6551f2ea8e5551.

controllers/nim_account_controller.go
```diff
@@ -0,0 +1,428 @@
...
+		account.Status = *status.DeepCopy()
+		if err = r.Client.Status().Update(ctx, account); err != nil {
+			logger.Error(err, "failed to update account status")
+		}
+	}
+}
+
+// createOwnerReferenceCfg is used to create an owner reference config to use with server side apply
+func (r *NimAccountReconciler) createOwnerReferenceCfg(account *v1.Account) *ssametav1.OwnerReferenceApplyConfiguration {
+	gvks, _, _ := r.Scheme().ObjectKinds(account)
```
lburgazzoli, line 354: nit: given the usage path, at this point probably the GVK of the account should be set so it is probably not needed to resolve it via the scheme.

lburgazzoli: note: this is not a blocker, just a suggestion but it seems there are some issues that prevent to implement it as described

TomerFi: Added a comment referring this in https://github.com/opendatahub-io/odh-model-controller/pull/289/commits/53cf8faa9f79792e0ab12174d53bc67bccc3b3b7.

Basically, what happens is that while working as expected in a live cluster, when running using envtest, the object's TypeMeta is being stripped. See the attached screenshots:

**Live**
image

**EnvTest*
image

TomerFi: This was tested with the latest envtest (0.19.1) using kube 1.26, 1.29, and 1.30.

controllers/nim_account_controller.go
TomerFi: You know what @lburgazzoli, I'm just reliazing something Yes,  you were right, we didn't need to delete the resources when an acount is being deleted, k8s will handle this.
But the func I removed in the above commit, was also used for when the API key secret is deleted. And this still needs to be handled.

So, I need to bring back the cleanup utility func  for cleanups, but only use it when the secret is deleted.

TomerFi: (re)Resolved in https://github.com/opendatahub-io/odh-model-controller/pull/289/commits/15dccb2c6f0d1b9814b7a5fe1ba75c8c7a6d4723.

controllers/constants/constants.go
```diff
@@ -71,3 +71,11 @@ const (
 const (
 	NoSuitableRuntimeError = "not found error: no suitable runtime found."
 )
+
+// NIM
+const (
+	NimApplyConfigFieldManager = "nim-account-controller"
+	NimDataConfigMapName       = "nvidia-nim-data"
+	NimRuntimeTemplateName     = "nvidia-nim-serving-template"
+	NimPullSecretName          = "nvidia-nim-image-pull"
```
TomerFi, line 80: @xieshenzh @etirelli @mpaulgreen

We reconcile a _ConfigMap_, a _Template_, and a pull _Secret_, named _nvidia-nim-data_, _nvidia-nim-serving-template_, and _nvidia-nim-image-pull_ respectively.

If we change the names to something more dynamic, i.e., based on the triggering _Account_ name, please correct me if I'm wrong, but we won't need the admission webhook in #288.

xieshenzh: I thought the names are dynamic, that's why the names are stored in the status of the Account CR.

The name of the Account is supposed to be provided by the user.

controllers/nim_account_controller.go
```diff
@@ -0,0 +1,480 @@
...
+			msg = "failed to fetch api key secret"
+			logger.V(1).Error(err, "failed to fetch api key secret")
+		}
+		meta.SetStatusCondition(&targetStatus.Conditions, makeAccountFailureCondition(account.Generation, msg))
+		r.cleanupResources(ctx, account)
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+	foundApiKeySec := "found api key secret"
+	logger.V(1).Info(foundApiKeySec)
+	meta.SetStatusCondition(&targetStatus.Conditions, makeAccountFailureCondition(account.Generation, foundApiKeySec))
```
terrytangyuan, line 149: nit: Consider making an util function to log + set condition since these are repeated many times

controllers/nim_account_controller.go
```diff
@@ -0,0 +1,480 @@
...
+func (r *NimAccountReconciler) reconcileNimConfig(
+	ctx context.Context, ownerCfg *ssametav1.OwnerReferenceApplyConfiguration,
+	namespace, apiKey string, runtimes []utils.NimRuntime,
+) (*corev1.ConfigMap, error) {
+	data, dErr := utils.GetNimModelData(apiKey, runtimes)
+	if dErr != nil {
+		return nil, dErr
+	}
+
+	cmapCfg := ssacorev1.ConfigMap(constants.NimDataConfigMapName, namespace).
```
terrytangyuan, line 264: nit: cmap -> cm

controllers/nim_account_controller.go
```diff
@@ -0,0 +1,480 @@
...
+	}
+
+	if _, err := controllerutil.CreateOrUpdate(ctx, r.Client, template, func() error {
+		if err := controllerutil.SetControllerReference(account, template, r.Scheme()); err != nil {
+			return err
+		}
+
+		template.Annotations = map[string]string{
+			"opendatahub.io/apiProtocol":         "REST",
+			"opendatahub.io/modelServingSupport": "[\"single\"]",
```
terrytangyuan, line 293: Are there reusable constants?

hack/verify_nvidia_nim_api.go
```diff
@@ -0,0 +1,95 @@
...
+		panic(vErr)
+	}
+	fmt.Println("API Key validated successfully")
+
+	models, dErr := utils.GetNimModelData(apiKey, runtimes)
+	if dErr != nil {
+		panic(dErr)
+	}
+	fmt.Printf("Got %d models info successfully", len(models))
+	fmt.Println()
```
terrytangyuan, line 78: These can be combined to one print?

TomerFi: Do you mean `fmt.Println(fmt.Sprintf...`?
I can try, but I think this will make _golint-ci_ mad 😁 .

controllers/constants/constants.go
TomerFi: That was the goal, but we were asked to keep these identical to what we have in TP. I think you also commented on this part in the ADR. But maybe I'm wrong.

I'll change these to be dynamic, which will mean that we fully support multiple accounts in the same namespace. If that's the case, why do we need an admission webhook validating that only one exists in a namespace?

TomerFi: I missed the _status_ perms, fixed here: https://github.com/opendatahub-io/odh-model-controller/pull/289/commits/974306614cbb9c5c1ad61ec5ec10308e2d7dfdab.

We now have:
```
// +kubebuilder:rbac:groups=nim.opendatahub.io,resources=accounts,verbs=get;list;watch;update
// +kubebuilder:rbac:groups=nim.opendatahub.io,resources=accounts/status,verbs=get;list;watch;update
```

I think this should suffice; please resolve if you agree or comment if you do not. 😄

controllers/nim_account_controller.go
TomerFi: I can move it to a constant if you prefer.

hack/verify_nvidia_nim_api.go
TomerFi: As expected, `fmt.Println(fmt.Sprintf("Got %d models info successfully", len(models)))`, breaks _golangci-lint_:

image

I can do: `fmt.Printf("Got %d models info successfully\n", len(models))`, but I don't usually prefer to include line breaks in string literals.

Your call.

controllers/constants/constants.go
xieshenzh: > That was the goal, but we were asked to keep these identical to what we have in TP. I think you also commented on this part in the ADR. But maybe I'm wrong.
> 
> I'll change these to be dynamic, which will mean that we fully support multiple accounts in the same namespace. If that's the case, why do we need an admission webhook validating that only one exists in a namespace?

We don't need the wehbook.

I don't think we need the operator at all for NIM enablement alone, if everything is static.
We only need to change the old CronJob to run go code instead of shell scripts, so that it would appear more maintainable.

main.go
TomerFi: I missed the _status_ perms, fixed here: https://github.com/opendatahub-io/odh-model-controller/pull/289/commits/974306614cbb9c5c1ad61ec5ec10308e2d7dfdab.

We now have:
```
// +kubebuilder:rbac:groups=nim.opendatahub.io,resources=accounts,verbs=get;list;watch;update
// +kubebuilder:rbac:groups=nim.opendatahub.io,resources=accounts/status,verbs=get;list;watch;update
```

I think this should suffice; please resolve if you agree or comment if you do not. 😄

controllers/constants/constants.go
TomerFi: I'm afraid I have to disagree with the job suggestion.

I agree with you that the webhook is unnecessary. This is exactly the point I was trying to make with the dynamic naming convention, but as pointed out in the parallel Slack thread, we need to involve the dashboard people in this.

controllers/nim_account_controller.go
TomerFi: Resolved in https://github.com/opendatahub-io/odh-model-controller/pull/289/commits/b8293f4d86d0821c78a8b52c56a7720f8ed4228f.

controllers/nim_account_controller.go
TomerFi: Also, while trying this, I noticed I missed an unknown condition. Resolved here https://github.com/opendatahub-io/odh-model-controller/pull/289/commits/b8293f4d86d0821c78a8b52c56a7720f8ed4228f.

TomerFi: Let me know if you want me to do the utility funcs regardless.

main.go
xieshenzh: I am resolving this. But please double check the logs of the operator after it runs for a few hours, and make sure there are not issues related to permissions in the logs.

xieshenzh: And don't forget to test updating and deleting Account CR.
These events could require additional permissions.

controllers/nim_account_controller.go
```diff
@@ -0,0 +1,484 @@
...
+	logger.V(1).Info(runtimesOk)
+	meta.SetStatusCondition(&targetStatus.Conditions, makeAccountFailureCondition(account.Generation, runtimesOk))
+
+	// validate api key
+	if err := utils.ValidateApiKey(apiKeyStr, availableRuntimes[0]); err != nil {
+		msg := "api key failed validation"
+		logger.Error(err, msg)
+		meta.SetStatusCondition(&targetStatus.Conditions, makeAccountFailureCondition(account.Generation, msg))
+		meta.SetStatusCondition(&targetStatus.Conditions, makeApiKeyFailureCondition(account.Generation, msg))
+		r.cleanupResources(ctx, account)
```
israel-hdez, line 184: What is the motivation for the cleanup in case of errors?

I have the thought that this can break momentarily a working environment during a [controller resync](https://github.com/kubernetes-sigs/controller-runtime/blob/8e29c6387e83b25306d59b87c29424f3e59ca255/designs/cache_options.md#L123).

controllers/utils/nim.go
```diff
@@ -0,0 +1,378 @@
...
+	nimGetRuntimeManifestFmt = "https://nvcr.io/v2/%s/manifests/%s"
+	nimGetNgcCatalog         = "https://api.ngc.nvidia.com/v2/search/catalog/resources/CONTAINER"
+	nimGetNgcToken           = "https://authn.nvidia.com/token?service=ngc&"
+	nimGetNgcModelDataFmt    = "https://api.ngc.nvidia.com/v2/org/%s/team/%s/repos/%s?resolve-labels=true"
+)
+
+var NimHttpClient HttpClient
+
+func init() {
+	NimHttpClient = &http.Client{}
```
israel-hdez, line 97: Configure the client with timeouts to prevent requests to hang forever....

controllers/nim_account_controller.go
```diff
@@ -0,0 +1,484 @@
...
+		return err
+	}
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("odh-nim-controller").
+		For(&v1.Account{}).
+		Owns(&corev1.ConfigMap{}).
+		Owns(&corev1.Secret{}).
+		Owns(&templatev1.Template{}).
+		Watches(&corev1.Secret{}, handler.EnqueueRequestsFromMapFunc(
+			func(ctx context.Context, obj client.Object) []reconcile.Request {
```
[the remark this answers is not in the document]
TomerFi: Good point!
The Secret is created by ODH Dashboard.

@olavtar @yzhao583 @andrewballantyne
Can you please confirm the Secret encapsulating the API key had the following label:

```
"opendatahub.io/managed": "true"
```

cc @mpaulgreen @xieshenzh

controllers/nim_account_controller.go
```diff
@@ -0,0 +1,484 @@
...
+			},
+		},
+	}
+
+	credsJson, marshErr := json.Marshal(creds)
+	if marshErr != nil {
+		return nil, marshErr
+	}
+
+	secretCfg := ssacorev1.
```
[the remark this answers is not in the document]
TomerFi: Resolved in https://github.com/opendatahub-io/odh-model-controller/pull/289/commits/85231d0a2756775c6a6f989c4e992f063188da4e.

hack/verify_nvidia_nim_api.go
TomerFi: Resolved in https://github.com/opendatahub-io/odh-model-controller/pull/289/commits/85231d0a2756775c6a6f989c4e992f063188da4e.

controllers/nim_account_controller.go
TomerFi: We don't clean up for all errors, only for errors relating to the API key validation. The motivation is to remove the model info from the cluster in case of a failed validation. We can remove this if you prefer. I don't think it was a requirement. It may be something that we did in TP and may no longer be required here since we use status conditions.

@mpaulgreen, @xieshenzh, Any thoughts?

mpaulgreen: @xieshenzh wrote the code in the cronob. So, this is the code that deleted the associated resources - https://github.com/opendatahub-io/odh-dashboard/pull/2959/files#diff-654ddbcfb120f3fa423302828fb6b0e72b2a680a025a7a09bc24ccfc4ea2abd9R103-R110. I am not sure why we want to keep the associated configmap and template if api key is invalid? There could be downstream impact too... @yzhao583 @@olavtar any issues if we dont remove the secret, cm and template? But @israel-hdez this is a very common pattern.

controllers/nim_account_controller.go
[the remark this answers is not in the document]
olavtar: @TomerFi I don't see the label in the nvidia-nim-access secret.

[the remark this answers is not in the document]
mpaulgreen: @olavtar You have to add this label

controllers/nim_account_controller.go
israel-hdez: > I am not sure why we want to keep the associated configmap and template if api key is invalid?

If the API Key is invalid, I'm OK with the clean-up. Although, my question is general to the clean-up rather than specific to this `if` block; i.e. there are other cleanups before this one. And, also, I'm thinking mainly about re-syncs.

AFAIK, the controller will do a re-sync (i.e. it will run a reconcile cycle) automatically after a 10-hour period even if nothing has changed in the cluster. As this reconcile is using external services, a re-sync may encounter temporary issues (e.g. API Key validation, or fetching runtimes returning 500 status code, or the request timing out). Such temporary errors will lead to a clean-up, despite everything is still valid.

Since the controller is returning an error, controller-framework would trigger another reconcile quickly which may re-create again all resources if it succeeds. The small window of the deleted resources is my concern. I haven't seen how these resources are used later.

So, if you don't see an issue with the missing resources for a ¿short? window of time, then I'm good with the clean-up.

controllers/nim_account_controller.go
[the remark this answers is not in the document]
andrewballantyne: Yu's PR https://github.com/opendatahub-io/odh-dashboard/pull/3455#discussion_r1838285420 is not currently doing this.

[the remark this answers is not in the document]
mpaulgreen: Its being added by @olavtar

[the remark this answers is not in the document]
olavtar: @mpaulgreen Actually Yu is adding it.

controllers/constants/constants.go
TomerFi: For this stage, we're keeping the hardcoded names; further discussion is in [Slack](https://redhat-internal.slack.com/archives/C07CZDA6DE1/p1731435611515029?thread_ts=1731428744.293129&cid=C07CZDA6DE1).

controllers/nim_account_controller.go
TomerFi: > If the API Key is invalid, I'm OK with the clean-up. Although, my question is general to the clean-up rather than specific to this if block; i.e. there are other cleanups before this one. And, also, I'm thinking mainly about re-syncs.

We clean up in four scenarios, all of which qualify as a failure to validate the API key:
1. We failed to fetch the Secret encapsulating the API key; see [here](https://github.com/TomerFi/odh-model-controller/blob/nim-account-controller/controllers/nim_account_controller.go#L149).
2. We got the Secret, but it doesn't have the required key; see [here](https://github.com/TomerFi/odh-model-controller/blob/nim-account-controller/controllers/nim_account_controller.go#L162).
3. We failed to fetch NIM's available runtimes; in this case, we don't have runtime info to validate the key with; see [here](https://github.com/TomerFi/odh-model-controller/blob/nim-account-controller/controllers/nim_account_controller.go#L176).
4.  The actual validation fail; see [here](https://github.com/TomerFi/odh-model-controller/blob/nim-account-controller/controllers/nim_account_controller.go#L189).

> AFAIK, the controller will do a re-sync (i.e. it will run a reconcile cycle) automatically after a 10-hour period even if nothing has changed in the cluster. As this reconcile is using external services, a re-sync may encounter temporary issues (e.g. API Key validation, or fetching runtimes returning 500 status code, or the request timing out). Such temporary errors will lead to a clean-up, despite everything is still valid.

Legitimate concern. Perhaps we should reconsider the cleanups.

> Since the controller is returning an error, controller-framework would trigger another reconcile quickly which may re-create again all resources if it succeeds. The small window of the deleted resources is my concern. I haven't seen how these resources are used later.

The controller only returns an error for scenarios 2-4 from the ones described above. If the Secret is not found, no error is returned; see [here](https://github.com/TomerFi/odh-model-controller/blob/nim-account-controller/controllers/nim_account_controller.go#L150).

> So, if you don't see an issue with the missing resources for a ¿short? window of time, then I'm good with the clean-up.

I think we can live with that, but I would be happy if the dashboard people could comment on this, too.
@olavtar @andrewballantyne

main.go
TomerFi: @xieshenzh You were right, finalzier permissions are also required:
```
cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on
```

controllers/nim_account_controller.go
```diff
@@ -422,6 +425,47 @@ func (r *NimAccountReconciler) cleanupResources(ctx context.Context, account *v1
 
 }
 
+func (r *NimAccountReconciler) TechPreviewCleanups() manager.RunnableFunc {
```
mpaulgreen, line 428: AFAIU this function should also cleanup up the api key secret if present from previous TP release. The other resources are something that is managed by operator so it can remain as you feel.

main.go
TomerFi: @xieshenzh I documented the tests I performed in this [doc](https://docs.google.com/document/d/1GlYxgkyYdIAs0mK2GeuHv3xCpUAQNl34__pxB67P1pw/edit?usp=sharing) (updated the PR body).

controllers/nim_account_controller.go
terrytangyuan: it's ok for now

controllers/nim_account_controller.go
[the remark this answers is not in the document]
israel-hdez: I'm now seeing PR opendatahub-io/odh-dashboard#3455 has the label. I'm resolving this conversation.

israel-hdez (approved):
/lgtm
/hold

PR #300, opened 2024-11-15
[RHOAIENG-15207] Removed the $(mesh_namespace) variable from manifests

mholder6:
Removed the unused variable `$(mesh_namespace)` from the manifests and added the `-opendatahub` suffix to be hardcoded instead. 

## Description
Removed the specification of how the ClusterRoleBinding resource should be defined in the `config/base/kustomization.yaml` file, and removed the variable from the `config/rbac/role_binding.yaml` file.

## How Has This Been Tested?
Ensured that the `-opendatahub` suffix persisted in the RHOAI 2.16 builds where many manifests were changed. The suffix persisting confirmed that the variable could be disposed of. Building the kustomize file with the removed variable resulted in the same manifests that were built with the variable there.

### DIFF
--- a/config/base/kustomization.yaml
+++ b/config/base/kustomization.yaml
@@ -136,17 +136,3 @@ replacements:
           name: validating-webhook-configuration
         fieldPaths:
           - webhooks.0.clientConfig.service.namespace
-  - source:
-      kind: ConfigMap
-      version: v1
-      name: odh-model-controller-parameters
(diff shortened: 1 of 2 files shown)


hdefazio (approved):
Nitpick: Can you update the commit message to specifically say "mesh_namespace"? Otherwise LGTM

mholder6:
@hdefazio Updated commit message, thanks!

PR #308, opened 2024-11-20
update global ca bundle logic and storage-config logic

Jooho:
## Description

ODH Operator changed the default string for customCA: https://github.com/opendatahub-io/opendatahub-operator/pull/1339
With this change, odh-model-controller need to change  some logic

1. customcacert logic 
small change - it simply trim new lines in cert data.
With this change, if there is only newlines in the cert data, it will not create a kserve custom cert cm.

2. storage-config logic
It was checking if global cert cm (odh-trusted-ca-bundle) exist but now it is checking if kserve custom cert cm(odh-kserve-custom-ca-bundle) exist. If so, it will add the certificate information into `storage-config`

ref https://issues.redhat.com/browse/RHOAIENG-16070
## How Has This Been Tested?

Setup
1. Deploy the latest kserve using odh operator
2. Update odh-model-controller image with quay.io/opendatahub/odh-model-controller:pr-308

Test
1. check if the kserve custom cert cm(odh-kserve-custom-ca-bundle) exist. by default, it should not be created.
2. update dsci 
~~~
    trustedCABundle:
      customCABundle: |
        ---TEST---
        TST
        ---TEST---
~~~
then check if  the kserve custom cert cm(odh-kserve-custom-ca-bundle) exist. it should be created.
Plus, storage-config secret should include the information like the following:
```
{"access_key_id":"XXX","bucket":"XXX","cabundle_configmap":"odh-kserve-custom-ca-bundle","certificate":"---TEST---\nTST\n---TEST---","default_bucket":"XXX","endpoint_url":"https://s3.amazonaws.com/","region":"us-east-1","secret_access_key":"XXX","type":"s3"}
```

## Merge criteria:

### DIFF
--- a/controllers/kserve_customcacert_controller.go
+++ b/controllers/kserve_customcacert_controller.go
@@ -18,6 +18,7 @@ package controllers
 import (
 	"context"
 	"reflect"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/opendatahub-io/odh-model-controller/controllers/constants"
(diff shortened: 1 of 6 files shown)


israel-hdez (approved):
/lgtm

PR #312, opened 2024-11-26
Add NIM flag logic

trujillm:
## Description
This will get the nim flag state of removed or managed from params.env and add to the environment variables for odh-model-controller to be utilize

## How Has This Been Tested?
Unit tests

### DIFF
--- a/config/base/params.env
+++ b/config/base/params.env
@@ -4,4 +4,4 @@ caikit-standalone-image=quay.io/opendatahub/caikit-nlp:fast
 tgis-image=quay.io/opendatahub/text-generation-inference:fast
 ovms-image=quay.io/opendatahub/openvino_model_server:2024.3-release
 vllm-image=quay.io/opendatahub/vllm:fast
-
+nim-state=removed
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -71,6 +71,12 @@ spec:
                   name: service-mesh-refs
                   key: MESH_NAMESPACE
                   optional: true
+            - name: NIM_STATE
+              valueFrom:
+                configMapKeyRef:
+                  name: odh-model-controller-parameters
+                  key: nim-state
+                  optional: true
           livenessProbe:
             httpGet:
               path: /healthz
--- a/main.go
+++ b/main.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"flag"
 	"os"
+	"slices"
 	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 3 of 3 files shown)


config/base/params.env
```diff
@@ -4,3 +4,4 @@ caikit-standalone-image=quay.io/opendatahub/caikit-nlp:stable-994ec60
 tgis-image=quay.io/opendatahub/text-generation-inference:stable-eba83ba
 ovms-image=quay.io/opendatahub/openvino_model_server:2024.3-release-4c8c52c	
 vllm-image=quay.io/opendatahub/vllm:stable-849f0f5
+nim-state=removed
```
spolti, line 7: not sure if here is the best place to add this.

trujillm: @spolti I believe this was the agreed upon strategy with @israel-hdez. It ties into the odh operator logic you can see here --> https://github.com/opendatahub-io/opendatahub-operator/pull/1330

spolti: oh, ok, it is fine then :)

spolti (approved):

trujillm:
@spolti do you know if I need to change something for the image tag failure?

israel-hdez (changes requested):
Please, change the target branch to `incubating`.
The `main` branch is now a stable/downstream branch.

config/base/params.env
```diff
@@ -1,6 +1,7 @@
...
-tgis-image=quay.io/opendatahub/text-generation-inference:fast
-ovms-image=quay.io/opendatahub/openvino_model_server:2024.3-release
-vllm-image=quay.io/opendatahub/vllm:fast
+odh-model-controller=quay.io/opendatahub/odh-model-controller:v0.12.0
+caikit-tgis-image=quay.io/opendatahub/caikit-tgis-serving:stable-2d9db23
+caikit-standalone-image=quay.io/opendatahub/caikit-nlp:stable-994ec60
+tgis-image=quay.io/opendatahub/text-generation-inference:stable-eba83ba
+ovms-image=quay.io/opendatahub/openvino_model_server:2024.3-release-4c8c52c	
+vllm-image=quay.io/opendatahub/vllm:stable-849f0f5
+nim-state=removed
```
zdtsw, line 7: so in https://github.com/opendatahub-io/opendatahub-operator/pull/1330,  if not specify nim-state in DSC, it should pass down "managed" as default. 
but here, if not get value from Operator, the default is removed.
isnt that inconsistent? 
if the requirement from UX is, unless user explicitly set "removed" in DSC, then it is always "managed", then it should be `nim-state=managed` here, right?
 in reality, it does not matter managed/remove/ "" in params.env since it will always be overrwritten by Operator. But if run odh-model-controller without Operator,  then set "removed" here is different than "managed" also cause confusion in the code.

trujillm: @zdtsw I think the intent here was to add removed explicitly as a safety in case someone ever decides to change in params.env without taking modlemesh into consideration.

config/base/params-vllm-gaudi.env
```diff
@@ -1 +1 @@
-vllm-gaudi-image=quay.io/opendatahub/vllm:fast-gaudi
```
israel-hdez, line 1: @trujillm may you rebase your PR on `incubation`?
Looks like you did your changes on top of the `main` branch, which is no longer our default branch, and there are a lot of changes because of the different base branch.

israel-hdez (commented):
@trujillm I reviewed the commit and it is OK to me. Just, please, rebase on top of `incubation` to bring only your changes.

config/base/params-vllm-gaudi.env
trujillm: @israel-hdez I believe I have corrected please verify and let me know if you need anything else from my side

israel-hdez (approved):
/lgtm

PR #316, opened 2024-11-26
Use the main branch to build stable image tags, incubating for latest…

hdefazio:
… image tags

## Merge criteria:

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,10 +5,10 @@ on:
   push:
     branches:
       - main
-      - 'release-[0-9]+\.[0-9]+\.[0-9]+'
+      - incubating
     tags:
       - 'latest'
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -5,10 +5,10 @@ on:
   push:
     branches:
       - main
-      - 'release-[0-9]+\.[0-9]+\.[0-9]+'
+      - incubating
     tags:
       - 'latest'
-      - 'v*'
+      - 'odh-v*'
```
spolti, line 11: we need to document it somewhere so we don't miss the correct tag name when doing the releases.

hdefazio: good point - I just updated the odh release process doc to reflect the new tags

PR #326, opened 2024-12-03
feat(mr): enable by default model registry inferenceservice reconcile

Al-Pragliola:
## Description

Ref: [RHOAIENG-15767](https://issues.redhat.com//browse/RHOAIENG-15767)

Recap:

1) The inferenceservice controller function has been moved to model registry and just imported here
2) The controller is enabled by default in the manifests using arg `--model-registry-inference-reconcile`
3) A new label has been added: `modelregistry.opendatahub.io/name` this is the name of the model registry from which we are deploying the model  - this is the name of the service/deployment (should be the same)
4) An annotation has been added: `modelregistry.opendatahub.io/url` this is the URL of the model registry where the controller will get/create the ServingEnvironment and the InferenceService, make sure that it is reachable from the controller's pod - in my tests I use the rest route generated from istio

## How Has This Been Tested?

Prerequisites:
- a model registry created using odh
- a model version deployed

#

- Enable image registry in the openshift cluster:
```shell
# enable default route for openshift image registry
oc patch configs.imageregistry.operator.openshift.io/cluster --type merge -p '{"spec":{"defaultRoute":true}}'
# get image registry host
export REGISTRY=`oc get route default-route -n openshift-image-registry --template='{{ .spec.host }}'`
# login to openshift registry
docker login -u $(oc whoami) -p $(oc whoami -t) $REGISTRY
```

- Build and push the registry:
```shell
docker build . -f ./Containerfile -t ${REGISTRY}/opendatahub/odh-model-controller:latest
docker push ${REGISTRY}/opendatahub/odh-model-controller:latest
```

- Enable devFlags inside the DataScienceCluster:
```yaml
    kserve:
      devFlags:
        manifests:
          - contextDir: config
            sourcePath: overlays/odh
            uri: 'https://github.com/Al-Pragliola/odh-model-controller/tarball/feat/enable-inference-mr'
```

- Patch the image in the odh-model-controller's deployment:
```shell
oc patch deployment odh-model-controller --patch='{"spec": {"template": {"spec":{"containers": [{"name": "manager", "image": "image-registry.openshift-image-registry.svc:5000/opendatahub/odh-model-controller:latest"}]}}}}'
```

- Get external route of MR
```shell
exoprt URL=`echo "https://$(oc get routes -n istio-system -l app.kubernetes.io/name=MODEL-REGISTRY-NAME -o json | jq '.items[].status.ingress[].host | select(contains("-rest"))')" | tr -d '"'`
```

- Add url annotation and name label to the inferenceservice
```shell
oc annotate inferenceservices.serving.kserve.io -n inferenceservice-namespace inferenceservice-name modelregistry.opendatahub.io/url=$URL
oc label inferenceservices.serving.kserve.io -n inferenceservice-namespace inferenceservice-name modelregistry.opendatahub.io/name=MODEL-REGISTRY-NAME
```

- Wait or watch odh-model-controller logs and then check the inferenceservice labels, a label called `modelregistry.opendatahub.io/inference-service-id` should have been added

## Merge criteria:


config/overlays/odh/odh_model_controller_manager_patch.yaml
```diff
@@ -8,5 +8,8 @@ spec:
       containers:
         - args:
             - --leader-elect
-          image: $(odh-model-controller) 
+            - --model-registry-inference-reconcile
+            - --model-registries-namespace
+            - "odh-model-registries"
```
tarilabs, line 13: I have a suspicious this might not be updated downstream (if compared here, to midstream) and to avoid headaches it would be better to read this from the DSC ?

tarilabs (commented):
I have one comment below, really appreciate this coming together @Al-Pragliola !

config/overlays/odh/odh_model_controller_manager_patch.yaml
Al-Pragliola: When do you think makes sense to read this information from DSC?

Midstream we can do that when we instantiate the controller but if anything changes on DSC we won't update this constant until we restart the pod

If we want to do that in the reconcile loop we have a problem because in upstream there's no concept of DSC

tarilabs: I believe doing it at Controller init time to be sufficient imho.

Worst case eventually, default namespace related values could be refactored as a function, so that upstream just return the value, and midstream another function is passed which can do the lookup as needed. But again, I'm missing to identify a case atm where this would be needed. Happy to hear otherwise!

Al-Pragliola: https://github.com/opendatahub-io/odh-model-controller/pull/326/commits/fafd6021f103138cf0779f7585ecf58509583965 implements taking the namespace at controller init time, feels free to resolve this conversation if it looks ok to you

spolti:
Looking at the code itself it looks ok to me, but I would rely on the Model Registry team review, once they give the green lights we can merge :)

spolti (approved):

tarilabs (approved):
thank you @Al-Pragliola for your ongoing effort with this MR-Isvc-reconciler, I'm happy to see this is now refactored to incorporate the code placed and refreshed upstream

as Dhiraj mentioned, we will need to keep to keep go.mod updated, and/or eventually a dedicated controller (for the longer term)

PR #333, opened 2024-12-10
Add metadata with upstream release versions 

heyselbi:
## Description

Addressing:
https://issues.redhat.com/browse/RHOAISTRAT-327
https://issues.redhat.com/browse/RHOAIENG-16305
https://issues.redhat.com/browse/RHOAIENG-14290
https://issues.redhat.com/browse/RHOAIENG-15498

In a nutshell, adding the release versions of upstream projects that our components deliver to ODH and RHOAI. These will be fetched for DSC status and displayed in UI.

This is specifically for upcoming ODH release. The file will need to be updated separately for RHOAI in case any versions differ.

Reviewers please double check that the release version is correct.

## How Has This Been Tested?

No testing yet since it does not affect the functionality of the repo. I will reach out to Feature owner if there is a specific way we should be testing the UI component of it.

## Merge criteria:

### DIFF
--- a/config/component_metadata.yaml
+++ b/config/component_metadata.yaml
@@ -1,0 +1,22 @@
+releases:
+  - name: KServe
+    version: v0.14.0
+    repoUrl: https://github.com/kserve/kserve/
+  - name: ModelMesh Serving
+    version: v0.12.0
+    repoUrl: https://github.com/kserve/modelmesh-serving
+  - name: vLLM
+    version: v0.6.4.post1
+    repoUrl: https://github.com/vllm-project/vllm
+  - name: OpenVINO Model Server
+    version: v2024.5
+    repoUrl: https://github.com/openvinotoolkit/model_server
+  - name: Caikit
+    version: v0.27.7
+    repoUrl: https://github.com/caikit/caikit
+  - name: Caikit-nlp
+    version: v0.5.8
+    repoUrl: https://github.com/caikit/caikit-nlp
+  - name: Text Generation Inference Server
+    version: commit 8a8c55dd0884b8e4e420d22f10bcce6752455edf
+    repoUrl: https://github.com/IBM/text-generation-inference/


config/component_metadata.yaml
```diff
@@ -0,0 +1,22 @@
+releases:
+  - name: KServe
+    version: v0.14.0
+    repoUrl: https://github.com/kserve/kserve/
```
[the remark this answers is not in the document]
heyselbi: @spolti The goal is to show what version of upstream upstream project we have in ODH and RHOAI. We are often asked, what upstream release version is KServe, vLLM, openvino etc. based on in ODH/RHOAI. The PR/feature will display this info in UI. I think it is pretty understandable that ODH and RHOAI have additional commits/changes in addition to what the upstream project is.

We are not to use branch but specific release version. The feature refinement document in the [JIRA](https://issues.redhat.com/browse/RHOAISTRAT-327) provides all detailed descriptions.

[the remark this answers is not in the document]
spolti: okay, so you need to update the branches as I suggested in my last comment.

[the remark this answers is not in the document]
spolti: After a conversation on slack we've agreed to keep 0.14.0 here

config/component_metadata.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - name: ModelMesh Serving
+    version: v0.12.0-rc0
```
spolti, line 6: However, IMHO, it is not a good thing to use the `rc` release tag here, users might think that our product does not use the final / stable release from upstream.

RH-steve-grubb:
OVMS looks good.

config/component_metadata.yaml
```diff
@@ -0,0 +1,22 @@
...
+    repoUrl: https://github.com/kserve/modelmesh-serving
+  - name: vLLM
+    version: v0.6.4.post1
```
tarukumar, line 9: This doesn't seems right `0.6.4.post2.dev115+g0ca5ce7b.d20241204` is the correct version

dtrifiro: Can we add a comment saying that this corresponds to the latest upstream tag merged into opendatahub-io/vllm?

dtrifiro: For example, right now we're at v0.6.4.post1 (https://github.com/opendatahub-io/vllm/releases/tag/v0.6.4.post1) 
other tags are going to be available here:
https://github.com/opendatahub-io/vllm/tags

heyselbi: @dtrifiro @tarukumar I will let the feature lead know that we would want to include a little note in the UI saying that all components have extra commits on top of the upstream community release, so customers/users know that our images, while they include the upstream community runtime up to x.y.z release, we have additional commits on top of it.

heyselbi: But I wouldn't include this comment here, since it is pertinent to all components (not just model serving).

config/component_metadata.yaml
heyselbi: Agreed. Will update the rc to full release.

spolti (approved):
approving the model server side.

terrytangyuan (approved):
Thanks!

terrytangyuan (commented):
How do we make sure that we don't forget to update this during release? Do we have a Jira template for release checklist?

config/component_metadata.yaml
dtrifiro: I think that leaving a note here aswell would do no harm, so whoever is updating this doesn't forget, but I'm fine either way

dtrifiro (approved):

heyselbi:
@terrytangyuan Yes, the next step/JIRA is to automate the process ideally by a bot, but if not, then by JIRA creation.

PR #339, opened 2024-12-16
Automatically inject expected ODH annotations to InferenceGraph and InferenceServices

israel-hdez:
**NOTE:** Requires PR #324 before merging (**fulfilled**)

## Description

On creation of InferenceGraph or InferenceService resources, the following default annotations will be added:
* `serving.knative.openshift.io/enablePassthrough: true`
* `sidecar.istio.io/inject: true`
* `sidecar.istio.io/rewriteAppHTTPProbers: true`

The annotations are added only for Serverless mode, and only if they are missing.

Fixes https://issues.redhat.com/browse/RHOAIENG-15337

## How Has This Been Tested?
* Deploy odh-model-controller from this PR, either by creating a custom build of the operator, or by using dev-flags
* Follow the InferenceGraph example at https://kserve.github.io/website/latest/modelserving/inference_graph/image_pipeline/#deploy-the-inferenceservices (you will need to configure pytorch runtime)
* When you create the InferenceGraph resource, confirm that the three default annotations (mentioned up) are automatically injected to the created InferenceGraph resource.
* Delete the InferenceGraph resource, and re-create it in Raw mode. Confirm that the default annotations are _not_ injected.
* Delete the InferenceGraph resource, and re-create it in Serverless mode, while manually specifying one (or some) of the annotations with a non-default value (e.g. `serving.knative.openshift.io/enablePassthrough: false`). Confirm that the manually specified annotations are honored, while the others are injected to the default values.

Similar tests can be performed on InferenceService resources.

## Merge criteria:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,1 @@
+Containerfile
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -48,6 +48,7 @@ import (
 	"github.com/opendatahub-io/odh-model-controller/internal/controller/utils"
 	webhooknimv1 "github.com/opendatahub-io/odh-model-controller/internal/webhook/nim/v1"
 	webhookservingv1 "github.com/opendatahub-io/odh-model-controller/internal/webhook/serving/v1"
+	webhookservingv1alpha1 "github.com/opendatahub-io/odh-model-controller/internal/webhook/serving/v1alpha1"
 	webhookservingv1beta1 "github.com/opendatahub-io/odh-model-controller/internal/webhook/serving/v1beta1"
 	// +kubebuilder:scaffold:imports
 )
@@ -269,6 +270,13 @@ func main() {
 	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
 		if err = webhookservingv1beta1.SetupInferenceServiceWebhookWithManager(mgr); err != nil {
 			setupLog.Error(err, "unable to create webhook", "webhook", "InferenceService")
+			os.Exit(1)
+		}
+	}
+	// nolint:goconst
+	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
+		if err = webhookservingv1alpha1.SetupInferenceGraphWebhookWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create webhook", "webhook", "InferenceGraph")
 			os.Exit(1)
 		}
 	}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -53,6 +53,10 @@ import (
 	// +kubebuilder:scaffold:imports
 )
 
+const (
+	enableWebhooksEnv = "ENABLE_WEBHOOKS"
+)
+
 var (
 	scheme   = runtime.NewScheme()
 	setupLog = ctrl.Log.WithName("setup")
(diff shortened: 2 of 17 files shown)


israel-hdez:
@spolti This PR is now fully ready, although we need first to merge the migration #324

Dockerfile
```diff
@@ -0,0 +1 @@
+Containerfile
```
spolti, line 1: I am not sure what is this change.

cmd/main.go
```diff
@@ -272,6 +273,13 @@ func main() {
 			os.Exit(1)
 		}
 	}
+	// nolint:goconst
+	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
```
spolti, line 277: maybe move ENABLE_WEBHOOKS to a constant?

internal/webhook/serving/v1alpha1/inferencegraph_webhook.go
```diff
@@ -0,0 +1,73 @@
...
+var inferencegraphlog = logf.Log.WithName("inferencegraph-resource")
+
+// SetupInferenceGraphWebhookWithManager registers the webhook for InferenceGraph in the manager.
+func SetupInferenceGraphWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).For(&servingv1alpha1.InferenceGraph{}).
+		WithDefaulter(&InferenceGraphCustomDefaulter{client: mgr.GetClient()}).
+		Complete()
+}
+
+// +kubebuilder:webhook:path=/mutate-serving-kserve-io-v1alpha1-inferencegraph,mutating=true,failurePolicy=fail,sideEffects=None,groups=serving.kserve.io,resources=inferencegraphs,verbs=create,versions=v1alpha1,name=minferencegraph-v1alpha1.odh-model-controller.opendatahub.io,admissionReviewVersions=v1
```
spolti, line 44: if user updates it and remove the annotations?

internal/webhook/serving/v1alpha1/inferencegraph_webhook_test.go
```diff
@@ -0,0 +1,111 @@
...
+	corev1 "k8s.io/api/core/v1"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	testutils "github.com/opendatahub-io/odh-model-controller/test/utils"
+)
+
+const InferenceServiceConfigPath1 = "../../../controller/serving/testdata/configmaps/inferenceservice-config.yaml"
+
+var _ = Describe("InferenceGraph Webhook", func() {
```
spolti, line 33: maybe add a use case that checks if the IG is updated.

Dockerfile
israel-hdez: Hoping I am remembering correctly, when running `kubebuilder` I got some errors, because it assumes the `Dockerfile` exists and it was trying to validate it.

I solved by creating a symbolic link named `Dockerfile` pointing to the `Containerfile`. GitHub does not make it clear, but this change is a symbolic link.

internal/webhook/serving/v1alpha1/inferencegraph_webhook.go
israel-hdez: The removal will be honored. I'm unsure if we should add back the defaults.

internal/webhook/serving/v1alpha1/inferencegraph_webhook_test.go
israel-hdez: What is the actual assertion you have in mind?

cmd/main.go
israel-hdez: I extracted to a local constant, given it is used only here.

internal/webhook/serving/v1alpha1/inferencegraph_webhook_test.go
spolti: I was thinking in case the IG is updated and the annotations need to be re-added, but that seems not the case.

spolti (approved):

PR #340, opened 2024-12-17
chore: use naming convention for resources created by nim

TomerFi:
## Description

Following the merge of https://github.com/opendatahub-io/odh-dashboard/pull/3535, we can use a naming convention to create the resources owned by NIM runtimes instead of static names.

This will also require cleanups when upgrading; see https://github.com/opendatahub-io/opendatahub-operator/pull/1472.

Jira: [NVPE-29](https://issues.redhat.com/browse/NVPE-29)
Slack: [thread](https://redhat-internal.slack.com/archives/C07CZDA6DE1/p1734382422391069)

## How Has This Been Tested?

```shell
make test
make build
```

## Merge criteria:

### DIFF
--- a/internal/controller/nim/account_controller.go
+++ b/internal/controller/nim/account_controller.go
@@ -285,7 +285,7 @@ func (r *AccountReconciler) reconcileNimConfig(
 		return nil, dErr
 	}
 
-	cmCfg := ssacorev1.ConfigMap(constants.NimDataConfigMapName, namespace).
+	cmCfg := ssacorev1.ConfigMap(fmt.Sprintf("%s-cm", *ownerCfg.Name), namespace).
 		WithData(data).
 		WithOwnerReferences(ownerCfg).
 		WithLabels(labels)
@@ -303,7 +303,7 @@ func (r *AccountReconciler) reconcileNimConfig(
 func (r *AccountReconciler) reconcileRuntimeTemplate(ctx context.Context, account *v1.Account) (*templatev1.Template, error) {
 	template := &templatev1.Template{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      constants.NimRuntimeTemplateName,
+			Name:      fmt.Sprintf("%s-template", account.Name),
 			Namespace: account.Namespace,
 		},
 	}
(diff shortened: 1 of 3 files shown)


controllers/nim_account_controller.go
```diff
@@ -266,7 +266,7 @@ func (r *NimAccountReconciler) reconcileNimConfig(
 		return nil, dErr
 	}
 
-	cmCfg := ssacorev1.ConfigMap(constants.NimDataConfigMapName, namespace).
```
israel-hdez, line 269: Since it is no longer static, would it make sense to add some assertions in the units?

TomerFi: Added in https://github.com/opendatahub-io/odh-model-controller/pull/340/commits/25027d88b599aa8fb9876334b8b570cfe6775c53

internal/controller/nim/account_controller.go
```diff
@@ -303,7 +303,7 @@ func (r *AccountReconciler) reconcileNimConfig(
 func (r *AccountReconciler) reconcileRuntimeTemplate(ctx context.Context, account *v1.Account) (*templatev1.Template, error) {
 	template := &templatev1.Template{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      constants.NimRuntimeTemplateName,
+			Name:      fmt.Sprintf("%s-template", account.Name),
```
spolti, line 306: is there any chance this field or the account object to be nil?

israel-hdez (commented):
The changes are good to me.
But Filippe has a question. So, I'll defer merging to him.

internal/controller/nim/account_controller.go
TomerFi: Nope. This is the account we were triggered for.
We fetch it [here](https://github.com/opendatahub-io/odh-model-controller/blob/incubating/internal/controller/nim/account_controller.go#L119) by name and namespace, and use it [here](https://github.com/opendatahub-io/odh-model-controller/blob/incubating/internal/controller/nim/account_controller.go#L238) to reconcile the template.

spolti (approved):
/lgtm

PR #344, opened 2025-01-08
chore: set nim runtime api call page size to 1000

TomerFi:
## Description

Based on NVIDIA's recommendations and the benchmarks executed in https://github.com/opendatahub-io/odh-model-controller/pull/336, setting the page size to 1000 will reduce API calls but will not affect performance.

## How Has This Been Tested?

```shell
make test
```
Also, see benchmark functions in https://github.com/opendatahub-io/odh-model-controller/pull/336.

## Merge criteria:


spolti (approved):
/lgtm

PR #349, opened 2025-01-16
fix null pointer error (RHOAIENG-18228)

Jooho:
## Description

When knative have some issues and if it did not deploy runtime, Address would not be set. 
We should check this situation.

## Merge criteria:

### DIFF
--- a/internal/controller/serving/reconcilers/kserve_isvc_gateway_reconciler.go
+++ b/internal/controller/serving/reconcilers/kserve_isvc_gateway_reconciler.go
@@ -67,7 +67,7 @@ func (r *KserveGatewayReconciler) Reconcile(ctx context.Context, log logr.Logger
 	_, meshNamespace = utils.GetIstioControlPlaneName(ctx, r.client)
 
 	// return if Address.URL is not set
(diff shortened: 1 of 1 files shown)


internal/controller/serving/reconcilers/kserve_isvc_gateway_reconciler.go
```diff
@@ -67,7 +67,7 @@ func (r *KserveGatewayReconciler) Reconcile(ctx context.Context, log logr.Logger
 	_, meshNamespace = utils.GetIstioControlPlaneName(ctx, r.client)
 
 	// return if Address.URL is not set
-	if isvc.Status.Address != nil && isvc.Status.Address.URL == nil {
+	if isvc.Status.Address == nil || isvc.Status.Address.URL == nil {
 		log.V(1).Info("Waiting for the URL as the InferenceService is not ready yet")
```
hdefazio, line 71: My only thought is - do we need to give a different error message if the Address field isn't set instead of just the URL? It seems like the original check was making sure that the Address field existed first but if the Address and URL field are always set together then this doesn't bother me too much.

PR #358, opened 2025-01-23
Fix for ModelMesh-only setups - Start InferenceGraph controller only if the CRD is present

israel-hdez:
## Description

When ODH platform only has ModelMesh installed without KServe, the InferenceGraph CRD won't be available. This setup was leading to odh-model-controller to crash, because InferenceGraph controller was always trying to start.

This adds conditionals to only start InferenceGraph controller if the CRD is available.

Fixes https://issues.redhat.com/browse/RHOAIENG-18577

## How Has This Been Tested?
Test 1:
* Setup ODH so that only ModelMesh is installed
* Check that odh-model-controller does not enter in CrashLoopbackOff state

Test 2:
* Setup ODH so that KServe is installed (optionally, ModelMesh also enabled)
* Check that odh-model-controller does not enter in CrashLoopbackOff state, and InferenceGraph controller starts.

## Merge criteria:

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -28,8 +28,10 @@ import (
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	"github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
 	istiov1beta1 "istio.io/client-go/pkg/apis/security/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -40,6 +42,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
@@ -53,7 +56,6 @@ import (
 	webhookservingv1 "github.com/opendatahub-io/odh-model-controller/internal/webhook/serving/v1"
 	webhookservingv1alpha1 "github.com/opendatahub-io/odh-model-controller/internal/webhook/serving/v1alpha1"
 	webhookservingv1beta1 "github.com/opendatahub-io/odh-model-controller/internal/webhook/serving/v1beta1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	// +kubebuilder:scaffold:imports
 )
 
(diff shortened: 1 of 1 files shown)


cmd/main.go
```diff
@@ -242,8 +244,73 @@ func main() {
 		}
 	}
 
-	nimState := os.Getenv("NIM_STATE")
 	signalHandlerCtx := ctrl.SetupSignalHandler()
+	setupNim(mgr, signalHandlerCtx, kubeClient)
+
+	// nolint:goconst
+	if os.Getenv(enableWebhooksEnv) != "false" {
```
spolti, line 251: we have 3 `if` doing the same thing, maybe it would be better to move all the webhooks under only one `if` statement, wdyt?

cmd/main.go
```diff
@@ -242,8 +244,73 @@ func main() {
...
+		os.Exit(1)
+	} else if inferenceGraphCrdAvailable {
+		if err = servingcontroller.NewInferenceGraphReconciler(mgr).SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "InferenceGraph")
+			os.Exit(1)
+		}
+	} else {
+		setupLog.Info("controller is turned off", "controller", "InferenceGraph")
+	}
+	// +kubebuilder:scaffold:builder
```
spolti, line 292: is this expected to be added here?

maybe these scaffold instruction can be moved to the beginning of the file.

cmd/main.go
Jooho: yea I agree

cmd/main.go
```diff
@@ -242,8 +244,73 @@ func main() {
...
+	}
+	// nolint:goconst
+	if os.Getenv(enableWebhooksEnv) != "false" {
+		if err = webhookservingv1beta1.SetupInferenceServiceWebhookWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create webhook", "webhook", "InferenceService")
+			os.Exit(1)
+		}
+	}
+	// nolint:goconst
+	if os.Getenv(enableWebhooksEnv) != "false" {
```
Jooho, line 270: Do we need to put this into #284 line? 

if there is no iIG crd, why we add this webhook? However, I think we need to add this because of webhook manifests. So IG webhook manifests will be created anyway. Instead, we should check  if there is IG crd on cluster or not from webhook. wdyt?

cmd/main.go
Jooho: `// nolint:goconst `
I think this should be removed if we gather this to 1 if

cmd/main.go
israel-hdez: I don't think the code of the webhook will be invoked at all if the CRD is missing. In any case, there is a type assertion at the very begging of the code of the webhook.

Jooho: then how to manage manifest for the IG webhook? the manifests would be created anyway. am I right?

cmd/main.go
israel-hdez: It is not added. It has been always here at last.
GitHub diff is not smart enough to find that I extracted some code in the middle of `main()` to the `setupNim` func. So, in the diff it looks like I added a lot, when in reality I moved code around.

You may find a much better diff view here: https://app.semanticdiff.com/gh/opendatahub-io/odh-model-controller/pull/358/changes#cmd/main.go?ignore_comments=false

israel-hdez: BTW, we could move it to the top. It will later move to the middle, since `kubebuilder` is always pushing it down.

spolti: that's fine.

cmd/main.go
israel-hdez: We have only one `if` :-)

spolti (approved):
/approve

spolti:
/lgtm
@Jooho and I tested with @israel-hdez image, `odh-m-c` is not crashing anymore with modelmesh and no `inferenceGraph` CRD available in the cluster.

PR #359, opened 2025-01-30
Move kserve and MM release metadata to corresponding repos

heyselbi:
## Description

Will close: https://issues.redhat.com/browse/RHOAIENG-18612
Per discussion with platform team and @Jooho , it was decided to move references to kserve and MM release versions in metadata from odh-model-controller to kserve/MM repos.
I am also removing references to runtime release versions, since odh-model-controller status is not displayed in DSC. Runtime version display will be covered as part of the https://issues.redhat.com/browse/RHOAIENG-9923

## Merge criteria:


ugiordan:
@heyselbi, I believe this metadata file should be removed from the repo. We don’t display status for `odh-model-controller` since it operates behind the scenes, so it’s not necessary for this component. I would expect the metadata files for both `KServe` and `MM Serving` to mention the `vLLM` release, so users are aware that enabling either one means vLLM is also being used.

heyselbi:
Latest update: I removed references to runtime release versions as well, since odh-model-controller status is not displayed in DSC. Runtime version display will be covered as part of the https://issues.redhat.com/browse/RHOAIENG-9923

PR #360, opened 2025-02-03
Update status post frequency for NIM account controller

trujillm:
#### **Issue**  
Currently, the NIM controller updates the **account status** only **at the end of reconciliation**. This delay causes slower updates, even though reconciliation is an **async process**.  

#### **Fix**  
- Modify the controller to update the account status **while reconciliation is still running**, rather than waiting until the end.  
- This allows status changes to be reflected **sooner**, improving responsiveness.  
- The frontend should still compare timestamps (1) to ensure consistency.  

#### **Impact**  
Faster account status updates.  
- Maintains async processing behavior.  
- Improves UI responsiveness without affecting reconciliation logic. 

### DIFF
--- a/internal/controller/nim/account_controller.go
+++ b/internal/controller/nim/account_controller.go
@@ -111,6 +111,7 @@ func (r *AccountReconciler) SetupWithManager(mgr ctrl.Manager, ctx context.Conte
 //
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.19.1/pkg/reconcile
+// nolint:gocyclo
 func (r *AccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := log.FromContext(ctx).WithValues("Account", req.Name, "namespace", req.Namespace)
 	ctx = log.IntoContext(ctx, logger)
--- a/internal/controller/nim/account_controller.go
+++ b/internal/controller/nim/account_controller.go
@@ -144,9 +144,11 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		},
 	}
 
-	defer func() {
-		r.updateStatus(ctx, req.NamespacedName, *targetStatus)
-	}()
+	accountSubject := types.NamespacedName{Name: req.Name, Namespace: req.Namespace}
+
+	if updateErr := utils.UpdateStatus(ctx, accountSubject, *targetStatus, r.Client); updateErr != nil {
+		logger.Error(updateErr, "failed to create the initial status")
+	}
 
 	// fetch api secret
 	secretNs := account.Spec.APIKeySecret.Namespace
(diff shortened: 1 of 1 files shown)


internal/controller/nim/account_controller.go
```diff
@@ -144,9 +145,11 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		},
 	}
 
-	defer func() {
-		r.updateStatus(ctx, req.NamespacedName, *targetStatus)
-	}()
+	accountSubject := types.NamespacedName{Name: req.Name, Namespace: req.Namespace}
+
+	if updateErr := utils.UpdateStatus(ctx, accountSubject, *targetStatus, r.Client); updateErr != nil {
```
spolti, line 150: are these `updateStatus` added on too many different places really needed?

trujillm: @spolti the use case is we want to post the status any time there is interaction witht he NIM API to give real time feedback to the user. If there is another way to accomplish this please advise otherwise I would say this achieves our goal

trujillm: @spolti did you have a chance to take another look at my response

spolti: I see, I am still not sure if it is needed, I mean, a reconcile loop shouldn't take too long to finish and with the status being updated so many times would be even tough to see everything that is being updated (unless I am missing details on how these updates will be displayed to the user).

One solution would be to put this status into a map and call the statusUpdate function only once a the end of the reconcile loop.

trujillm: @spolti I don't think that would be a valid solution. The use case here is we make dozens of calls to the NVIDIA API for each model currently the list is at 67. Meaning each reconciliation is expected to send roughly 71 consecutive requests. The objective here is get real time feed back on if and when any of those calls fail.

trujillm: @spolti when you have time please see my comment above and advise. We were wanting to get this in for 2.19

spolti:
@israel-hdez do you mind reviewing it as well?

trujillm:
@israel-hdez and @spolti just wanted to bump this again. Trying to get merged for 2.19
Please review when you have some time.

internal/controller/nim/account_controller.go
```diff
@@ -165,53 +168,83 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 		msg := "api key failed validation"
 		logger.Error(err, msg)
 		meta.SetStatusCondition(&targetStatus.Conditions, makeAccountFailureCondition(account.Generation, msg))
+		if updateErr := utils.UpdateStatus(ctx, accountSubject, *targetStatus, r.Client); updateErr != nil {
+			logger.Error(updateErr, "failed to create status for failed api key validation")
+		}
 		meta.SetStatusCondition(&targetStatus.Conditions, makeApiKeyFailureCondition(account.Generation, msg))
+		if updateErr := utils.UpdateStatus(ctx, accountSubject, *targetStatus, r.Client); updateErr != nil {
+			logger.Error(updateErr, "failed to create status for failed api key validation")
+		}
```
israel-hdez, line 234: My only feedback is that I don't see useful the very fine updates here, and on similar instances where two updates are done individually.

I think those `meta.SetStatusCondition` calls would run very, very quickly. The two status updates may make the controller run slower, and one call for both updates may work better.

israel-hdez (approved):
I only had one comment. But it is minor. I'm not blocking this from merging.
/lgtm

internal/controller/nim/account_controller.go
spolti: I was thinking about it as well.
On a cluster that has many different namespaces with nim enabled, how that would behave?

PR #362, opened 2025-02-04
fix: added accept header to nim manifest pull requests accepting oci

TomerFi:
## Description

Currently, 14 runtimes out of the overall 67 runtimes we get from NVIDIA’s API, are OCI-based and support multi-arch. When attempting to pull a manifest for one of the said runtimes, we get a 404 with a message hidden in the body, indicating we need to explicitly accept OCI images:

```json
{
  "errors": [
    {
      "code": "MANIFEST_UNKNOWN",
      "message": "OCI index found, but accept header does not support OCI indexes"
    }
  ]
}
```

We fetch the available runtime list from NVIDIA on reconciling, this list is not ordered, and although the response appears cached, the item slots may differ between calls. We use the first runtime from the list for validating the API key, in cases when an OCI-based image is the first one, the API key validation fails, we cleanup and update the status accordingly.

This can be fixed by adding a header indicating the accepted media type to requests for manifest pulling:
(see [OpenContainers spec](https://github.com/opencontainers/image-spec/blob/main/image-index.md)).
```
Accept: application/vnd.oci.image.index.v1+json
```

Adding this header means that, for OCI images, the manifest pull will now include a body with details about the different underlying architecture-based manifests. In our case, we don't work the body, we only care about the status code, and we treat 200 as a successful validation.

### Without the header

**Successful non-OCI image**:
image

**Faild OCI image**:
image

### With the header

**Successful non-OCI image**:
image

**Successful OCI image**:
image

Jira: https://issues.redhat.com/browse/NVPE-137
Addendum doc: https://docs.google.com/document/d/1MNERaPIS__5eQCh-ad9GmX_KosleRCbYb20F1Wdarss/edit?tab=t.0

## How Has This Been Tested?

Using a slightly modified version of the [verify_nvidia_nim_api script](https://github.com/opendatahub-io/odh-model-controller/blob/incubating/hack/verify_nvidia_nim_api.go), verifying every runtime out of the 67 existing ones, is validated successfully with the new header.

Follow the [Reproduce Steps](https://docs.google.com/document/d/1MNERaPIS__5eQCh-ad9GmX_KosleRCbYb20F1Wdarss/edit?tab=t.0#heading=h.sc9u2xn4gmsv) section in the addendum [doc](https://docs.google.com/document/d/1MNERaPIS__5eQCh-ad9GmX_KosleRCbYb20F1Wdarss/edit?tab=t.0#heading=h.sc9u2xn4gmsv).

## Merge criteria:

### DIFF
--- a/internal/controller/utils/nim.go
+++ b/internal/controller/utils/nim.go
@@ -275,7 +275,6 @@ func attemptToPullManifest(runtime NimRuntime, tokenResp *NimTokenResponse) erro
 	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", tokenResp.Token))
 	req.Header.Add("Accept", "application/vnd.oci.image.index.v1+json")
 
-	fmt.Println(req.Header.Get("Accept"))
 	resp, respErr := NimHttpClient.Do(req)
 	if respErr != nil {
 		return respErr
--- a/internal/controller/utils/nim.go
+++ b/internal/controller/utils/nim.go
@@ -273,7 +273,9 @@ func attemptToPullManifest(runtime NimRuntime, tokenResp *NimTokenResponse) erro
 	}
 
 	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", tokenResp.Token))
+	req.Header.Add("Accept", "application/vnd.oci.image.index.v1+json")
 
+	fmt.Println(req.Header.Get("Accept"))
 	resp, respErr := NimHttpClient.Do(req)
 	if respErr != nil {
 		return respErr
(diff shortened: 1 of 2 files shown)


internal/controller/utils/nim.go
```diff
@@ -273,7 +273,9 @@ func attemptToPullManifest(runtime NimRuntime, tokenResp *NimTokenResponse) erro
 	}
 
 	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", tokenResp.Token))
+	req.Header.Add("Accept", "application/vnd.oci.image.index.v1+json")
 
+	fmt.Println(req.Header.Get("Accept"))
```
Jooho, line 278: is it an intentional one?

Jooho (commented):
small question

internal/controller/utils/nim.go
TomerFi: hehehehe no.

TomerFi: removed in https://github.com/opendatahub-io/odh-model-controller/pull/362/commits/adc835d4f208425da4e0779dd0b0a152beb57da5.

israel-hdez (approved):

PR #365, opened 2025-02-05
[RHOAIENG-16435] Fixed error that was caused when deleting an ISVC using RawDeployment and ServiceMesh is not installed

andresllh:
Added a check to ignore the error if the CRD does not exist.

## Description

Added a check to see if the error being thrown is due to a CRD not being installed when the serverless inference reconciler was cleaning up resources in the namespace.

## How Has This Been Tested?

I installed the OpenDataHub Operator. Created a Serving-Runtime and an InferenceService using RawDeployment annotation. Deleted the inference service and saw that the error was gone.
Used devspace to create a dev environment in my cluster and sync my local files to it.
Ran `make lint` and `make test` and made sure all tests were passing.

## Merge criteria:

### DIFF
--- a/internal/controller/serving/inferenceservice_controller.go
+++ b/internal/controller/serving/inferenceservice_controller.go
@@ -122,8 +122,7 @@ func (r *InferenceServiceReconciler) ReconcileServing(ctx context.Context, req c
 		logger.Info("Stop InferenceService reconciliation")
 		// InferenceService not found, so we check for any other inference services that might be using Kserve/ModelMesh
 		// If none are found, we delete the common namespace-scoped resources that were created for Kserve/ModelMesh.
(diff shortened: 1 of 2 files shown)


devspace.yaml
```diff
@@ -0,0 +1,103 @@
+version: v2beta1
```
spolti, line 1: this is very nice, but moving it to another PR would be better, it is better to add in a commit only changes related to the issue we are addressing.

andresllh: Removed the devspace files. Will create a new PR for it.

andresllh: PR created: https://github.com/opendatahub-io/odh-model-controller/pull/366

spolti (approved):

spolti (approved):

PR #366, opened 2025-02-06
Adding devspace files

andresllh:
## Description

Adding devspace files for developers that want to use devspace.

## How Has This Been Tested?

Installed [devspace](devspace.sh) cli  
Ran `devspace init` to create devspace_start.sh script and devspace.yaml files
Modified devspace.yaml with correct config for the odh-model-controller.
Ran `devspace dev` and terminal came up.
Changed directory to `/app/cmd` and ran `go run main.go` and the operator started successfully

## Merge criteria:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ go.work
 
 # Other
 .envrc
+.devspace
--- a/dev_tools/devspace.yaml
+++ b/dev_tools/devspace.yaml
@@ -1,0 +1,103 @@
+version: v2beta1
+name: odh-model-controllergit
+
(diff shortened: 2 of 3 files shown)


spolti:
For this, maybe we could create a new directory like tools, or put it inside hack, would it make sense?

andresllh:
> For this, maybe we could create a new directory like tools, or put it inside hack, would it make sense?

Created `dev_tools` folder that lives at the root of the repository and moved the files in  there. Updated config to reflect the new directory.

israel-hdez (approved):
Dev tooling :-) 
Let's merge it. I'm not familiar, but this can be useful for people.
/lgtm

PR #368, opened 2025-02-06
fix: cleanup fails for disabled NIM

TomerFi:
## Description

When setting _DataScienceCluster.spec.kserve.nim.managedState_ to _Removed_ and restarting _odh-model-controller_, NIM accounts require cleanups. The cleanup process is performed before the manager starts, hence the cache is not yet available, which ends up with the cleanup process failing and logging the following error:
```
ERROR setup failed to fetch accounts \{"error": "the cache is not started, can not read objects"}
```

This PR fixes this by moving the cleanup process to a runnable managed by the manager, the cleanup will be executed after the manager starts and the cache will be available. The new cleanup logic was moved from _main_ to the _utils_ and includes tests.

Jira: https://issues.redhat.com/browse/NVPE-147

## How Has This Been Tested?

- Added unit test cases.
- Built an intermediate image as `quay.io/tomerfi/odh-model-controller:fix-cleanup` and used it as in our development environment. Testing process:
  - Successfully enable NIM
  - Set _DataScienceCluster.spec.kserve.nim.managedState_ to _Removed_
  - Verify _odh-model-controller-parameters.data.nim-state_ was reconciled to _removed_
  - Restart _odh-model-controller_
  - Verify the error is not logged
  - Verify the supported resources were removed (ConfigMap, Secret, Template)
  - Verify the Account status was updated

nim-disabled

## Merge criteria:

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -31,10 +31,8 @@ import (
 	"github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
 	istiov1beta1 "istio.io/client-go/pkg/apis/security/v1beta1"
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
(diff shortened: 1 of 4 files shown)


TomerFi:
> the only way to get the nim account correctly removed is by restarting odh?

@spolti Yeah, this is by design. This specific cleanup process and the flag triggering it are meant to be used when the goal is to disable the controller entirely, which requires restarting the deployment. Other types of cleanup are handled in the controller logic when the flag is "managed" and the controller is running.

spolti (approved):

spolti:
okay, thanks for the explanation.

Can you please fix the conflict?

hdefazio:
@TomerFi Does this need to be in 2.18? If so, we need to land this today?

TomerFi:
Although it would be nice to see this in 2.18, this is fixing a specific end case, not breaking, and has a clear workaround. We can wait for 2.19 if the 2.18 process has already started.

TomerFi:
@hdefazio I just noticed in Jira that this is in fact scheduled for 2.18. Is that still possible?

hdefazio:
@TomerFi the code freeze for RHOAI 2.18 is today, so you'll need to land this, and then cherry pick it to https://github.com/opendatahub-io/odh-model-controller/tree/main and https://github.com/red-hat-data-services/odh-model-controller before EOD

spolti (commented):
/lgtm

PR #369, opened 2025-02-07
feat(mr): re-enable model registry reconciler and add svc annotation to it

Al-Pragliola:
## Description

Requires https://github.com/opendatahub-io/opendatahub-operator/pull/1682

This PR aims to re-enable the model registry inference reconciler, which was disabled by default after the recent refactoring https://github.com/opendatahub-io/odh-model-controller/pull/324, and add an additional parameter to the reconciler function.

Changes:
- removed `config/base/odh_model_controller_manager_patch.yaml` file as it is no longer used
- ~~added `--model-registry-inference-reconcile` to the manager manifest~~
- added `ModelRegistryServiceAnnotation` 
- updated dep to `github.com/kubeflow/model-registry`
- fixed dsc mr namespace retrieval function

## How Has This Been Tested?

Prerequisites:
- a model registry created using odh
- a model version deployed

#

- Enable image registry in the openshift cluster:
```shell
# enable default route for openshift image registry
oc patch configs.imageregistry.operator.openshift.io/cluster --type merge -p '{"spec":{"defaultRoute":true}}'
# get image registry host
export REGISTRY=`oc get route default-route -n openshift-image-registry --template='{{ .spec.host }}'`
# login to openshift registry
docker login -u $(oc whoami) -p $(oc whoami -t) $REGISTRY
```

- Build and push the registry:
```shell
docker build . -f ./Containerfile -t ${REGISTRY}/opendatahub/odh-model-controller:latest
docker push ${REGISTRY}/opendatahub/odh-model-controller:latest
```

- Scale down opendatahub operator to prevent reconciliation
```shell
oc scale deployment --replicas=0 -n openshift-operators opendatahub-operator-controller-manager
```

- Patch the image in the odh-model-controller's deployment:
```shell
oc patch deployment odh-model-controller -n opendatahub --patch='{"spec": {"template": {"spec":{"containers": [{"name": "manager", "image": "image-registry.openshift-image-registry.svc:5000/opendatahub/odh-model-controller:latest", "args":["--metrics-bind-address=:8080", "--leader-elect", "--health-probe-bind-address=:8081", "--model-registry-inference-reconcile"]}]}}}}'
```

- Add name label to the inferenceservice
```shell
oc label inferenceservices.serving.kserve.io -n inferenceservice-namespace inferenceservice-name modelregistry.opendatahub.io/name=MODEL-REGISTRY-NAME
```

- Wait or watch odh-model-controller logs and then check the inferenceservice labels, a label called `modelregistry.opendatahub.io/inference-service-id` should have been added

- Restore opendatahub operator replicas
```shell
oc scale deployment --replicas=1 -n openshift-operators opendatahub-operator-controller-manager
```

## Merge criteria:


config/base/odh_model_controller_manager_patch.yaml
```diff
@@ -1,13 +0,0 @@
-apiVersion: apps/v1
```
spolti, line 1: why remove it?

@zdtsw is this required from the opendatahub-operator side?

spolti: oh, nevermind :D

go.mod
```diff
@@ -12,7 +12,7 @@ require (
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/kserve/kserve v0.14.0
 	github.com/kuadrant/authorino v0.18.1
-	github.com/kubeflow/model-registry v0.2.11
+	github.com/kubeflow/model-registry v0.2.7-alpha.0.20250210125605-b5509328e038
```
spolti, line 15: why downgrading the version?

internal/controller/constants/constants.go
```diff
@@ -65,6 +65,7 @@ const (
 	ModelRegistryModelVersionIdLabel     = "modelregistry.opendatahub.io/model-version-id"
 	ModelRegistryRegisteredModelIdLabel  = "modelregistry.opendatahub.io/registered-model-id"
 	ModelRegistryFinalizer               = "modelregistry.opendatahub.io/finalizer"
+	ModelRegistryServiceAnnotation       = "routing.opendatahub.io/external-address-rest"
```
spolti, line 68: serviceAnnotation points to an external route?

Isn't better to rename this constant to something else that matches `external-address-rest`?

also, is `routing.opendatahub.io` required? Or can we keep the naming convention like the others?
`modelregistry.opendatahub.io/external-route`

internal/controller/serving/reconcilers/model_registry_reconciler.go
```diff
@@ -29,12 +29,16 @@ func NewModelRegistryInferenceServiceReconciler(client client.Client, log logr.L
 	if len(dscList.Items) > 0 {
 		dsc = dscList.Items[0]
 
-		ns, found, err := unstructured.NestedFieldCopy(dsc.Object, "spec", "components", "modelRegistry", "registriesNamespace")
-		if err == nil && found {
+		ns, found, err := unstructured.NestedFieldCopy(dsc.Object, "spec", "components", "modelregistry", "registriesNamespace")
+		if err != nil || !found {
+			log.Error(err, "Failed to get Model Registry Namespace from DataScienceCluster")
```
spolti, line 34: maybe log as a warning if it is not a blocker and inform the user what namespace will be set instead.

go.mod
Al-Pragliola: This comes from a commit rather than a tag, specifically:

```shell
go get github.com/kubeflow/model-registry@b5509328e038700bb186a554858ba7b55c58a22a
```

If this concerns you, we can do an ad hoc release upstream and point to the new tag here

internal/controller/serving/reconcilers/model_registry_reconciler.go
spolti: wait, is there no default NS if it isn't found in the DSC?
What happens if there is no NS set? If that's the case we should return a error to the reconciler.

internal/controller/constants/constants.go
Al-Pragliola: To give a bit of context, in each k8s service associated with a model registry instance, we annotate it with 
`routing.opendatahub.io/external-address-rest` (and also another one for grpc) with the model registry operator with the urls exposed by istio, this info is used by the dashboard and any other component that needs it to communicate with the model registry via http calls. Now, I think changing it introduces breaking changes with the clients, so if possible I would like to not do that, but yes I can rename the variable if it makes it easier to think about it

spolti: okay, thanks for the explanation.

internal/controller/serving/reconcilers/model_registry_reconciler.go
Al-Pragliola: I think there should not be a case when the field is empty, looking at the definition in the DSC type:

```go
	// Namespace for model registries to be installed, configurable only once when model registry is enabled, defaults to "odh-model-registries"
	// +kubebuilder:default="odh-model-registries"
	// +kubebuilder:validation:Pattern="^([a-z0-9]([-a-z0-9]*[a-z0-9])?)?$"
	// +kubebuilder:validation:MaxLength=63
	RegistriesNamespace string `json:"registriesNamespace,omitempty"`
```

but yeah I agree with you, it's better to add the possibility to return an error, maybe in the case of modelregistry not being in the "Managed" state

spolti: oh, ok, if there is the default one being set it is fine, we just need to make sure at the code level that it is set.

Al-Pragliola: added a check in https://github.com/opendatahub-io/odh-model-controller/pull/369/commits/eb2017532b14bbb735d5f15b7a665d22332d1fb4

go.mod
spolti: so, are you going to update it?

internal/controller/constants/constants.go
gabemontero, line 68: @Al-Pragliola is this PR intending to address the need for setting the `modelregistry.opendatahub.io/name` label on the inferenceservice like I've had to do with the v2.23 based testing we have been collaborating on?

I do see that label listed a few lines above ... is it possible the need for setting that label was addressed by a prior PR that did not make it into v2.23 ?

thanks

hdefazio:
@gabemontero Does this need to be in 2.18? If so, we need to land this today

gabemontero:
> @gabemontero Does this need to be in 2.18? If so, we need to land this today

I'll defer to @Al-Pragliola on what versions it needs to go into .... I'm still a bit new to ODH, but I was informed we were talking about getting this in 2.24 (which is about to go out) or 2.24.x / 2.25 @hdefazio ; apologies based on 2.24 I'm not sure where putting in 2.18 comes from.

internal/controller/constants/constants.go
Al-Pragliola: Hey @gabemontero, let me give you an overview:

- PRE PR[#326](https://github.com/opendatahub-io/odh-model-controller/pull/326)

inferenceservice controller was disabled by default -- code was outdated and not working properly

- PR[#326](https://github.com/opendatahub-io/odh-model-controller/pull/326)

The controller was enabled by default, the label modelregistry.opendatahub.io/name was added, also the annotation 
modelregistry.opendatahub.io/url has also been added.
The internal logic of the controller has been moved to the upstream repository https://github.com/kubeflow/model-registry/ .
This PR was merged into ODH 2.23 release

- PR[#324](https://github.com/opendatahub-io/odh-model-controller/pull/324)

This PR migrates odh-model-controller (which includes our inferenceservice controller) to a newer version of kubebuilder and does a refactoring of how manifests are handled, long story short, the inferenceservice controller is disabled again by default.

- ODH 2.23

This release includes #326 and #324 -- #324 is merged after #326, so the inferenceservice controller is disabled in this release.

- Current PR [#369](https://github.com/opendatahub-io/odh-model-controller/pull/369)

In this PR, we enable the controller again, and we get some improvements that came out of the testing we did together, like removing the need for the URL annotation, and also a retrocompatibility option to possibly omit the name label if there is only one model registry in the namespace.

As things stand, #369 will not make it to the 2.24 release of ODH, and probably not to the 2.18 release of RHOAI either - we can target ODH 2.25, or if you need to drop the use of devFlags, we can try a 2.24.1 and 2.19 for RHOAI. I hope things are clearer, if you have any questions don't hesitate to 🙏🏼

hdefazio:
@gabemontero No worries - I was talking about RHOAI 2.18, which has a code freeze today so you would need to cherry pick this in a few places. But we can totally wait for ODH 2.25, which will be released in about a week (and therefore it would end up in RHOAI 2.20 based on how we do releases in the team)

gabemontero:
> @gabemontero No worries - I was talking about RHOAI 2.18, which has a code freeze today so you would need to cherry pick this in a few places. But we can totally wait for ODH 2.25, which will be released in about a week (and therefore it would end up in RHOAI 2.20 based on how we do releases in the team)

thanks for the clarification @hdefazio  ... and fwiw @Al-Pragliola just clarified it for me in slack as well :-)

fwiw absolutely long term my team (the Backstage.io based project RHDH (Developer Hub) will want to prereq RHOAI and not ODH but we'll work with @Al-Pragliola and team to determine when it is appropriate to do that

go.mod
Al-Pragliola: done

internal/controller/constants/constants.go
gabemontero: thanks for the detail @Al-Pragliola  .... I think the key bit for me is that with **this PR** I now no longer need to add the `modelregistry.opendatahub.io/name` to my kserve `InferenceService` because `and also a retrocompatibility option to possibly omit the name label if there is only one model registry in the namespace.` 

Where in my testing at least there is only one model registry.

Have I parsed it correctly?

Al-Pragliola: yes, correct

internal/controller/serving/reconcilers/model_registry_reconciler.go
```diff
@@ -12,7 +14,12 @@ import (
 	"github.com/opendatahub-io/odh-model-controller/internal/controller/utils"
 )
 
-func NewModelRegistryInferenceServiceReconciler(client client.Client, log logr.Logger, skipTLSVerify bool, bearerToken string) *infrctrl.InferenceServiceController {
+var (
+	errGetDSC                = errors.New("failed to get DataScienceCluster")
+	errGetMRNamespaceFromDSC = errors.New("failed to get Model Registry Namespace from DataScienceCluster")
+)
```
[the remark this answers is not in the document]
Al-Pragliola: No, it's not needed, but any errors from this function will be joined into the parent function:

```go
		if err != nil {
			return reconcileResult, errors.Join(reconcileErr, err)
		}
```

So I think it's more future proof to extract them as errors to have the ability to customize the handling with `errors.Is`

config/manager/manager.yaml
```diff
@@ -63,6 +63,7 @@ spec:
         args:
           - --leader-elect
           - --health-probe-bind-address=:8081
+          - --model-registry-inference-reconcile
```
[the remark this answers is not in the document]
Al-Pragliola: that's correct, do you think it's a blocker for merging this PR? cc @spolti

internal/controller/serving/reconcilers/model_registry_reconciler.go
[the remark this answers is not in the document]
zdtsw: sure. just try to understand that

internal/controller/serving/reconcilers/model_registry_reconciler.go
```diff
@@ -23,18 +30,30 @@ func NewModelRegistryInferenceServiceReconciler(client client.Client, log logr.L
...
+		if !isNsOk {
+			return nil, fmt.Errorf("%w: invalid namespace", errGetMRNamespaceFromDSC)
 		}
 	}
 
+	if mrNamespaceFromDSC == "" {
+		return nil, fmt.Errorf("%w: empty namespace", errGetMRNamespaceFromDSC)
+	}
+
+	log.Info("Model Registry Namespace from DataScienceCluster", "Namespace", mrNamespaceFromDSC)
```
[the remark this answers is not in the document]
zdtsw: right, did not mean to get the len check away, but might be good to have a check on == 0 as well

config/manager/manager.yaml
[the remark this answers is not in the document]
zdtsw: dont take my words as gating your PR if it is approved by serving guys.
i was more on the thinking of, we have built a cross-component depdency between modelreg and odh-model-controller.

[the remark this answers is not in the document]
Al-Pragliola: Spoiler alert - it's a work in progress to extract this controller and put it under the wings of the model registry (upstream)

internal/controller/serving/reconcilers/model_registry_reconciler.go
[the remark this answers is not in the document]
Al-Pragliola: do you mean on `dscList.Items` ?

config/manager/manager.yaml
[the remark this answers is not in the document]
zdtsw: shocks taken- if that is already on the agenda,  good to let us know asap, so we can plan it from ourside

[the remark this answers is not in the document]
Al-Pragliola: yeah as soon as community approves we will let you know 👍🏼

internal/controller/serving/reconcilers/model_registry_reconciler.go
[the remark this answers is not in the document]
zdtsw: yeah, you already have 
	if len(dscList.Items) > 0 { // positive case} but it should be len(dscList) == 1
and the negatvie should cover == 0 ||  >1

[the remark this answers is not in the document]
Al-Pragliola: done, thanks for the suggestion

config/manager/manager.yaml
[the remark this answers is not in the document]
israel-hdez: > that's correct, do you think it's a blocker for merging this PR?

I think it is a blocker if odh-model-controller is going to fail if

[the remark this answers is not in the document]
Al-Pragliola: thanks for you feedback! I have moved the controller under a feature gate like NIM in the latest commit, please let me know if it's the correct approach @israel-hdez @spolti @zdtsw 

also I have added the need PR in opendatahub-operator https://github.com/opendatahub-io/opendatahub-operator/pull/1682

internal/controller/serving/reconcilers/model_registry_reconciler.go
```diff
@@ -23,18 +28,31 @@ func NewModelRegistryInferenceServiceReconciler(client client.Client, log logr.L
 
 	err := client.List(context.Background(), &dscList)
 	if err != nil {
-		log.Error(err, "Failed to list DataScienceCluster")
+		return nil, fmt.Errorf("%w: %w", errGetDSC, err)
+	}
+
+	if len(dscList.Items) == 0 || len(dscList.Items) > 1 {
```
spolti, line 34: Doesn't the opendatahub-operator already checks it?

internal/controller/serving/reconcilers/model_registry_reconciler.go
```diff
@@ -23,18 +28,31 @@ func NewModelRegistryInferenceServiceReconciler(client client.Client, log logr.L
...
 
-		ns, found, err := unstructured.NestedFieldCopy(dsc.Object, "spec", "components", "modelRegistry", "registriesNamespace")
-		if err == nil && found {
-			mrNamespaceFromDSC = ns.(string)
-		}
+	if mrNamespaceFromDSC == "" {
+		return nil, fmt.Errorf("%w: empty namespace", errGetMRNamespaceFromDSC)
 	}
 
+	log.Info("Model Registry Namespace from DataScienceCluster", "Namespace", mrNamespaceFromDSC)
```
spolti, line 54: is the context of this message to log that MR was enabled?

internal/controller/serving/reconcilers/model_registry_reconciler.go
Al-Pragliola: yes it does, for reference of this change a link to the discussion with @zdtsw https://github.com/opendatahub-io/odh-model-controller/pull/369#discussion_r1961814344

internal/controller/serving/reconcilers/model_registry_reconciler.go
Al-Pragliola: no if MR is not enabled the reconcile should not run as requested here https://github.com/opendatahub-io/odh-model-controller/pull/369#discussion_r1962871162 and worked in https://github.com/opendatahub-io/opendatahub-operator/pull/1682

this message is just to inform the user about which namespace is set to default if the inferenceservice does not specify a `modelregistry.opendatahub.io/namespace` label

spolti (approved):

hdefazio:
@dhirajsb So sorry for the late reply, yes this should be in the odh nightlies

PR #374, opened 2025-02-16
add kserve-state and modelmesh-state as args

VedantMahabaleshwarkar:
## Description

Fixes: [RHOAIENG-18974](https://issues.redhat.com/browse/RHOAIENG-18974)
- add `kserve-state` and `modelmesh-state` in odh-model-controller-parameters 
- pass `kserve-state` and `modelmesh-state` values as env variables in the odh-model-controller deployment
- use `kserve-state` value to determine whether to start InferenceGraph controller

## How Has This Been Tested?

- run `kustomize build config/base > temp.txt`
- change `kserve-state` and `modelmeshserving-state` in `config/base/params.env` from `managed` to `unmanaged` in any combination. 
- For each combination run `kustomize build config/base > temp1.txt` etc 
- Compare the kustomize build outputs for each case
- each resulting deployment yaml should differ (thus leading to a new deployment on a cluster) 

## Merge criteria:

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -137,13 +137,18 @@ func main() {
 		os.Exit(1)
 	}
 
+	kserveState := os.Getenv("KSERVE_STATE")
+	modelMeshState := os.Getenv("MODELMESH_STATE")
+	setupLog.Info("Installation status of Serving Components",
+		"kserve-state", kserveState, "modelmesh-state", modelMeshState)
+
 	kserveWithMeshEnabled, kserveWithMeshEnabledErr := utils.VerifyIfComponentIsEnabled(
 		context.Background(), mgr.GetClient(), utils.KServeWithServiceMeshComponent)
 	if kserveWithMeshEnabledErr != nil {
 		setupLog.Error(kserveWithMeshEnabledErr, "could not determine if kserve have service mesh enabled")
 	}
 
-	if err := setupReconcilers(mgr, setupLog, kubeClient, cfg); err != nil {
+	if err := setupReconcilers(mgr, setupLog, kubeClient, cfg, kserveState, modelMeshState); err != nil {
 		os.Exit(1)
 	}
 	// +kubebuilder:scaffold:builder
@@ -310,7 +315,7 @@ func setupWebhooks(mgr ctrl.Manager, setupLog logr.Logger, kserveWithMeshEnabled
 }
 
 func setupReconcilers(mgr ctrl.Manager, setupLog logr.Logger,
-	kubeClient kubernetes.Interface, cfg *rest.Config) error {
+	kubeClient kubernetes.Interface, cfg *rest.Config, kserveState string, _ string) error {
 	if err := setupInferenceServiceReconciler(mgr, kubeClient, cfg); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "InferenceService")
 		return err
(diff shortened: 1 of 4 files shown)


VedantMahabaleshwarkar:
@israel-hdez @zdtsw PTAL. Not sure if I'm missing anything but from what I saw the args are passed correctly when the odh-model-controller pod is created initially. However if you later change the DSC (for e.g: change Modelmesh from `Removed` to `Managed`), the odh-model-controller pod is not restarted. Maybe we need restart the pod from odh operator?

zdtsw:
> @israel-hdez @zdtsw PTAL. Not sure if I'm missing anything but from what I saw the args are passed correctly when the odh-model-controller pod is created initially. However if you later change the DSC (for e.g: change Modelmesh from `Removed` to `Managed`), the odh-model-controller pod is not restarted. Maybe we need restart the pod from odh operator?

two things:
1. seem the value of modelserving is not really used by setupReconcilers()?
2. the behavior i think should be the same for nim, right? have you waited for long enough 1-2mins and still nothing happened. I think when params.env is updated, then odh operator will try to deploy manifests again, then in turn configmap is changed, but for deployment to get notified it could take 1-2mins before it restarts pod.

cmd/main.go
```diff
@@ -332,21 +337,26 @@ func setupReconcilers(mgr ctrl.Manager, setupLog logr.Logger,
...
+		if igCrdErr != nil {
+			setupLog.Error(igCrdErr, "unable to check if InferenceGraph CRD is available", "controller", "InferenceGraph")
+			return igCrdErr
+		} else if inferenceGraphCrdAvailable {
+			if err := setupInferenceGraphReconciler(mgr); err != nil {
+				setupLog.Error(err, "unable to create controller", "controller", "InferenceGraph")
+				return err
+			}
+		} else {
+			setupLog.Info("crds unavailable, skipping controller", "controller", "InferenceGraph")
```
israel-hdez, line 354: IMO, the CRD check is no longer needed and we should always start the InferenceGraph controller when KServe is enabled. It is expected that the CRD is present. If it isn't present, there is a setup issue.

config/base/kustomization.yaml
```diff
@@ -124,6 +124,39 @@ replacements:
...
+      kind: ConfigMap
+      version: v1
+      name: odh-model-controller-parameters
+      fieldPath: data.nim-state
+    targets:
+      - select:
+          kind: Deployment
+          name: odh-model-controller
+        fieldPaths:
+          - spec.template.spec.containers.0.env.[name=NIM_STATE].value
```
israel-hdez, line 137: May you also here use `containers.[name=manager]` instead of `containers.0`? Just to prevent unexpected results if the order ever changes.

config/manager/manager.yaml
```diff
@@ -101,11 +101,11 @@ spec:
                 key: MESH_NAMESPACE
                 optional: true
           - name: NIM_STATE
-            valueFrom:
-              configMapKeyRef:
-                name: odh-model-controller-parameters
-                key: nim-state
-                optional: false
+            value: $(nim-state)
```
israel-hdez, line 104: I think it is better to hard-code some value, rather than using this legacy `$(variable)` syntax. Just to not confuse people with variables which, IIRC, we have moved away from.

zdtsw:
if the PR includes fix for nim-state too, maybe ask someone from NIM team as reviewer too.
also update description

VedantMahabaleshwarkar:
@TomerFi Can you take a quick look at this? I changed the way `nim-state` is attached to the deployment because I didn't find that odh-model-controller was restarting when the state changed.

zdtsw:
> > I didn't find that odh-model-controller was restarting when the state changed
> 
> @VedantMahabaleshwarkar, Before this PR, we were aware the deployment required a manual restart because changing values in ConfigMaps doesn't reflect or trigger a reconciliation of deployments using them as sources for environment variables.
> 
> So, when we modified the nim state in DataScienceCluster, that triggered an update to the odh-model-controller-parameters ConfigMap. But the odh-model-controller, needed to be restarted manually to see the updated environment variables.
> 
> Now that you changed the kustomization to patch the deployment directly, instead of having the deployment use the ConfigMap as a source, looks like this will trigger a restart of the deployment, and this is an improvement. Thank you!
> 😃 

I think with this PR then the manual restart is not required. Unless it is intention to keep it, as allow user decide when they wanna it take that effects, then we need revert the change done on nim part.

israel-hdez (approved):
Approving, assuming all are good.
/lgtm

israel-hdez:
Thanks @VedantMahabaleshwarkar

TomerFi:
@israel-hdez This goes into 2.19 right?

zdtsw:
> @israel-hdez This goes into 2.19 right?

operator side code is in 2.18, but serving part not, so will be 2.19

israel-hdez:
@TomerFi As Wen said :-)

zdtsw:
> > operator side code is in 2.18, but serving part not, so will be 2.19
> 
> @zdtsw to clarify, by operator side you mean odhoperator? and by serving side, do you mean this?

correct.
ODH/ RHOAI operator has the changes for geting kserve-state and modelmesh-state which is in ODH 2.24.0 and to-be RHOAI 2.18.0
but it wont be called till this PR is in odh-model-controller. which will be planned ODH 2.25.0 and to-be RHOAI 2.19.0

as for the nim-state, because of this PR, the behavior would change, follow the same logic as the other two newly introduced flags, 
i think if you have anywhere in the document telling user to manually restart odh-model-controller pod after they switch nim-state flag, it is not needed any more (after 2.18). unless you still perfer the old way as in 2.17, then we need to revert it on nim-state because this PR had been merged last night :D

PR #345, opened 2025-01-08
Authorization for InferenceGraph (Serverless)

israel-hdez:
## Description

This adds a new controller for KServe InferenceGraph resources. This new controller will have the responsibility of creating Authorino AuthConfig resources (similarly to InferenceServices case), when authorization is available in ODH platform.

InferenceGraphs can now be annotated with `security.opendatahub.io/enable-auth: "true"` to secure InferenceGraphs and only serve requests that are authorized.

Requires PRs:
* opendatahub-io/kserve#463
* opendatahub-io/opendatahub-operator#1491

Fixes: https://issues.redhat.com/browse/RHOAIENG-13449

## How Has This Been Tested?
* Deploy (since odh-operator is changed, it would be easier to use custom builds of everything):
  * odh-model-controller from this PR
  * KServe using code of dependent PR
  * operator using code of dependent PR
* Deploy InferenceGraph example of KServe website: https://kserve.github.io/website/latest/modelserving/inference_graph/image_pipeline/
* Check that the InferenceGraph is accessible
* Annotate the InferenceGraph with `security.opendatahub.io/enable-auth: "true"`
* Check that the InferenceGraph is accessible only to authorized requests

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #377, opened 2025-02-20
fix: when in Raw Mode as no Authorino is available, should not own AuthConfig

zdtsw:
when introduced PR https://github.com/opendatahub-io/odh-model-controller/pull/345 , an old logic to dynamically add Owns on AuthConfig is removed.
with latest official build, on a cluster without Authorino previously installed, Raw mode wont work , pod keep restart with inferencegraph controller not ready

## Description

ref: https://issues.redhat.com/browse/RHOAIENG-20304

## How Has This Been Tested?

local build: quay.io/wenzhou/odh-model-controller:20239

## Merge criteria:

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -146,7 +146,13 @@ func main() {
 		setupLog.Error(kserveWithMeshEnabledErr, "could not determine if kserve have service mesh enabled")
 	}
 
-	if err := setupReconcilers(mgr, setupLog, kubeClient, cfg, kserveState, modelMeshState); err != nil {
+	if err := setupReconcilers(
+		mgr,
+		setupLog,
+		kubeClient,
+		cfg,
+		kserveWithMeshEnabled,
+		kserveState, modelMeshState); err != nil {
 		os.Exit(1)
 	}
 	// +kubebuilder:scaffold:builder
@@ -313,7 +319,7 @@ func setupWebhooks(mgr ctrl.Manager, setupLog logr.Logger, kserveWithMeshEnabled
 }
 
 func setupReconcilers(mgr ctrl.Manager, setupLog logr.Logger,
-	kubeClient kubernetes.Interface, cfg *rest.Config, kserveState string, _ string) error {
+	kubeClient kubernetes.Interface, cfg *rest.Config, kserveWithMeshEnabled bool, kserveState string, _ string) error {
 	if err := setupInferenceServiceReconciler(mgr, kubeClient, cfg); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "InferenceService")
 		return err
@@ -336,7 +342,7 @@ func setupReconcilers(mgr ctrl.Manager, setupLog logr.Logger,
 	}
 
 	if kserveState == "managed" {
-		if err := setupInferenceGraphReconciler(mgr); err != nil {
+		if err := setupInferenceGraphReconciler(mgr, kserveWithMeshEnabled); err != nil {
 			setupLog.Error(err, "unable to create controller", "controller", "InferenceGraph")
 			return err
 		}
@@ -389,6 +395,6 @@ func setupServingRuntimeReconciler(mgr ctrl.Manager) error {
 	}).SetupWithManager(mgr)
 }
 
-func setupInferenceGraphReconciler(mgr ctrl.Manager) error {
-	return servingcontroller.NewInferenceGraphReconciler(mgr).SetupWithManager(mgr)
+func setupInferenceGraphReconciler(mgr ctrl.Manager, isServingMode bool) error {
+	return servingcontroller.NewInferenceGraphReconciler(mgr).SetupWithManager(mgr, isServingMode)
 }
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -25,14 +25,14 @@ import (
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
-	_ "k8s.io/client-go/plugin/pkg/client/auth"
-	"k8s.io/client-go/rest"
-
+	authorinov1beta2 "github.com/kuadrant/authorino/api/v1beta2"
 	istiov1beta1 "istio.io/client-go/pkg/apis/security/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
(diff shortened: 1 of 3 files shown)


cmd/main.go
```diff
@@ -337,7 +337,15 @@ func setupReconcilers(mgr ctrl.Manager, setupLog logr.Logger,
 	}
 
 	if kserveState == "managed" {
-		if err := setupInferenceGraphReconciler(mgr); err != nil {
+		authConfigCrdAvailable, authCrdErr := utils.IsCrdAvailable(
```
lburgazzoli, line 340: nit: since `authConfigCrdAvailable` does not seems to be used elsewhere, wonder if this check should be moved to `SetupWithManager` function where it is actually in use

zdtsw: that can be done
i first added it in `InferenceGraphReconciler.SetupWithManager()`, but then seems they wanna have the check move as higher as possible, so it was eventually done in the main

internal/controller/serving/inferencegraph_controller.go
```diff
@@ -167,11 +167,22 @@ func (r *InferenceGraphReconciler) getExistingAuthConfig(ctx context.Context, ig
...
+		Named("serving-inferencegraph")
+	// dynamic add authconfig to watchlist based on serving mode
+	authConfigCrdAvailable, authCrdErr := utils.IsCrdAvailable(
+		mgr.GetConfig(),
+		authorinov1beta2.GroupVersion.String(),
+		"AuthConfig")
+	if authCrdErr != nil {
+		return fmt.Errorf("failed to check if  AuthConfig CRD in the cluster: %w", authCrdErr)
+	}
+	if authConfigCrdAvailable {
```
spolti, line 181: isn't better to check if the raw deployment is configured?
This way it will make authorino optional for serverless mode, no?

zdtsw: you mean like, even authorino operator is installed in the cluster or authconfig CRD existis in the cluster, still user opt-in Raw then Serverless mode? 
lets say, if kserve is removed, modelmesh is managed. does inferencegraph need authconfig? i guess not. but it is by default has Serverless mode from kserve, right?

spolti: > you mean like, even authorino operator is installed in the cluster or authconfig CRD exists in the cluster, still user opt-in Raw then Serverless mode?
More or less, yes. But as the user can use or not auth in Serverless mode what I suggested might not be a good option.

>  if kserve is removed, modelmesh is managed. Does inferencegraph need authconfig? i
I don't think so, IIRC MM does not support IG. So auth should not be added anyway.

Thinking better. The way you proposed would be probably better because:
- when a user enables auth, but no CRD is available, it will throw errors anyway, so the user will know that he initially opted out for auth but wants to use it now, so he needs to install it.
- it will not expect that Authorino is always there, but if it is, register for use.


But, I guess there are other places that will fail in case Auth is not present, here, for example:
https://github.com/opendatahub-io/odh-model-controller/blob/incubating/internal/controller/serving/inferencegraph_controller.go#L115

Did you spot any issue in this part when you tested this patch?

@israel-hdez ^

zdtsw: from what @rnetser and Milland test, once AuthConfig CRD is in place, nothing went wrong after that.
i think in this case, it will call `reconcileAuthConfig()` but with return nil in the end unless it hits `capabilityErr`


btw,  i passed `kserveWithMeshEnabled` from the `main`, which i think can be a 2nd guard in the if checks before dynamically adding it

spolti: sounds good :)
Let's wait @israel-hdez for another review

internal/controller/serving/inferencegraph_controller.go
```diff
@@ -166,12 +166,23 @@ func (r *InferenceGraphReconciler) getExistingAuthConfig(ctx context.Context, ig
...
+		Named("serving-inferencegraph")
+	// dynamic add authconfig to watchlist based on serving mode + if authconfig crd is available
+	authConfigCrdAvailable, authCrdErr := utils.IsCrdAvailable(
+		mgr.GetConfig(),
+		authorinov1beta2.GroupVersion.String(),
+		"AuthConfig")
+	if authCrdErr != nil {
+		return fmt.Errorf("failed to check if  AuthConfig CRD in the cluster: %w", authCrdErr)
+	}
+	if authConfigCrdAvailable && isServingMode {
```
Jooho, line 181: I think a fast-fail way would be better. so if isServingMode is false then it should return right away.

cmd/main.go
```diff
@@ -390,6 +395,6 @@ func setupServingRuntimeReconciler(mgr ctrl.Manager) error {
 	}).SetupWithManager(mgr)
 }
 
-func setupInferenceGraphReconciler(mgr ctrl.Manager) error {
-	return servingcontroller.NewInferenceGraphReconciler(mgr).SetupWithManager(mgr)
+func setupInferenceGraphReconciler(mgr ctrl.Manager, isServingMode bool) error {
```
Jooho, line 398: `isServingMode` is vague to me. serverlessMode sounds better to me

internal/controller/serving/inferencegraph_controller.go
Jooho: >https://github.com/opendatahub-io/odh-model-controller/blob/incubating/internal/controller/serving/inferencegraph_controller.go#L115
Did you spot any issue in this part when you tested this patch?

This should be ok because there is condition to check authConfig : https://github.com/opendatahub-io/odh-model-controller/blob/incubating/internal/controller/serving/inferencegraph_controller.go#L98

internal/controller/serving/inferencegraph_controller.go
zdtsw: thanks, did some updated

internal/controller/serving/inferencegraph_controller.go
```diff
@@ -166,12 +166,26 @@ func (r *InferenceGraphReconciler) getExistingAuthConfig(ctx context.Context, ig
...
-func (r *InferenceGraphReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
+func (r *InferenceGraphReconciler) SetupWithManager(mgr ctrl.Manager, isServerlessMode bool) error {
+	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&servingv1alpha1.InferenceGraph{}).
-		Owns(&authorinov1beta2.AuthConfig{}).
-		Named("serving-inferencegraph").
-		Complete(r)
+		Named("serving-inferencegraph")
+	if !isServerlessMode {
```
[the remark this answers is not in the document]
israel-hdez: Raw doesn't use authorino, but oauth-proxy.

israel-hdez (approved):
This implementation is basically the same we had for InferenceServices controller before to the kubebuilder v4 refactor: https://github.com/opendatahub-io/odh-model-controller/blob/odh-v2.22/controllers/inferenceservice_controller.go#L175-L195. This code needs to come back (I created https://issues.redhat.com/browse/RHOAIENG-20356)

Effectively being the same solution:
/lgtm

spolti (approved):

PR #383, opened 2025-02-28
chore: added logs for nim api responses

TomerFi:
## Description

Added debug logs for nvidia nim API responses.
Jira: [NVPE-143](https://issues.redhat.com/browse/NVPE-143)

## How Has This Been Tested?

- Executed the hack script verifying NIM access.
- Enabled debug level and modified the URL to a faulty one in order to verify the debug log.
- Verified the test cases and linting.
- Created an intermediate image and deployed it in an environment running enablement issues and verified the logs.

## Merge criteria:

### DIFF
--- a/hack/verify_nvidia_nim_api.go
+++ b/hack/verify_nvidia_nim_api.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	"github.com/opendatahub-io/odh-model-controller/internal/controller/utils"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
 // Use this script for validating NVIDIA API access used by the NIM Account Controller.
@@ -42,6 +43,7 @@ import (
 // ** add -verbose to the script name to print the runtime images and models (it's a lot of data).
 func main() {
 	verbose := flag.Bool("verbose", false, "verbose")
+	logger := zap.New()
 
 	flag.Parse()
 	if len(flag.Args()) < 1 {
@@ -49,7 +51,7 @@ func main() {
 	}
 	apiKey := flag.Arg(0)
 
-	runtimes, rErr := utils.GetAvailableNimRuntimes()
+	runtimes, rErr := utils.GetAvailableNimRuntimes(logger)
 	if rErr != nil {
 		panic(rErr)
 	}
@@ -65,12 +67,12 @@ func main() {
 		}
 	}
 
-	if vErr := utils.ValidateApiKey(apiKey, runtimes[0]); vErr != nil {
+	if vErr := utils.ValidateApiKey(logger, apiKey, runtimes[0]); vErr != nil {
 		panic(vErr)
 	}
 	fmt.Println("API Key validated successfully")
 
-	models, dErr := utils.GetNimModelData(apiKey, runtimes)
+	models, dErr := utils.GetNimModelData(logger, apiKey, runtimes)
 	if dErr != nil {
 		panic(dErr)
 	}
@@ -90,5 +92,4 @@ func main() {
 			fmt.Println()
 		}
 	}
-
 }
--- a/internal/controller/nim/account_controller.go
+++ b/internal/controller/nim/account_controller.go
@@ -68,7 +68,7 @@ var (
 
 func (r *AccountReconciler) SetupWithManager(mgr ctrl.Manager, ctx context.Context) error {
 	// TODO: Copied from original main.go... Should it be FromContext?
-	logger := ctrl.Log.WithName("controllers").WithName("ModelRegistryInferenceService")
+	logger := ctrl.Log.WithName("controllers").WithName("AccountControllerSetup")
 	log.IntoContext(ctx, logger)
 
 	if err := mgr.GetFieldIndexer().IndexField(ctx, &v1.Account{}, apiKeySpecPath, func(obj client.Object) []string {
--- a/internal/controller/nim/account_controller.go
+++ b/internal/controller/nim/account_controller.go
@@ -69,6 +69,7 @@ var (
 func (r *AccountReconciler) SetupWithManager(mgr ctrl.Manager, ctx context.Context) error {
 	// TODO: Copied from original main.go... Should it be FromContext?
 	logger := ctrl.Log.WithName("controllers").WithName("ModelRegistryInferenceService")
+	log.IntoContext(ctx, logger)
 
 	if err := mgr.GetFieldIndexer().IndexField(ctx, &v1.Account{}, apiKeySpecPath, func(obj client.Object) []string {
 		return []string{obj.(*v1.Account).Spec.APIKeySecret.Name}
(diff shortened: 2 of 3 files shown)


internal/controller/nim/account_controller.go
```diff
@@ -69,6 +69,7 @@ var (
 func (r *AccountReconciler) SetupWithManager(mgr ctrl.Manager, ctx context.Context) error {
 	// TODO: Copied from original main.go... Should it be FromContext?
```
Jooho, line 70: yes we are using log fromContext if there are no issues with it.  Do you have any issues to use the way?

hack/verify_nvidia_nim_api.go
```diff
@@ -65,12 +67,12 @@ func main() {
 		}
```
spolti, line 67: Why not replace this print with a logger?

internal/controller/nim/account_controller.go
```diff
@@ -69,6 +69,7 @@ var (
...
 	logger := ctrl.Log.WithName("controllers").WithName("ModelRegistryInferenceService")
```
spolti, line 71: is this logger name correct?

internal/controller/utils/nim.go
```diff
@@ -189,12 +190,26 @@ func getNimRuntimes(runtimes []NimRuntime, page, pageSize int) ([]NimRuntime, er
...
 	return runtimes, nil
 }
 
+func logApiResponse(logger logr.Logger, resp *http.Response) {
+	logger.V(1).Info(fmt.Sprintf("got api response %s", resp.Status))
+	if resp.StatusCode != http.StatusOK {
+		if resp.ContentLength > 0 {
+			body, err := io.ReadAll(resp.Body)
+			if err != nil {
+				logger.V(1).Error(fmt.Errorf("failed to parse body"), "")
```
spolti, line 205: append the error here?

internal/controller/utils/nim.go
```diff
@@ -189,12 +190,26 @@ func getNimRuntimes(runtimes []NimRuntime, page, pageSize int) ([]NimRuntime, er
...
 	runtimes = append(runtimes, mapNimCatalogResponseToRuntimeList(catRes)...)
 	if catRes.Params.Page < catRes.ResultPageTotal-1 {
-		return getNimRuntimes(runtimes, page+1, pageSize)
+		return getNimRuntimes(logger, runtimes, page+1, pageSize)
 	}
 
 	return runtimes, nil
 }
 
+func logApiResponse(logger logr.Logger, resp *http.Response) {
```
spolti, line 199: for this func, I am not sure if it is really necessary to use the in context logger, instead, create a new logger here, it would avoid passing the logger as parameter everywhere.

hack/verify_nvidia_nim_api.go
TomerFi: I thought about that. But I didn't want it to be a part of this PR. The script is just something for us to work with, not consumed by anyone else.
I don't want to block this PR with modifications to this script.

internal/controller/nim/account_controller.go
TomerFi: Probably not, but this was not introduced in this PR. This was introduced when the project migrated to kubebuilder v4. See https://github.com/opendatahub-io/odh-model-controller/pull/324.

I can fix it in this PR if you prefer or do it in a different PR. We can even include the TODO comment @Jooho was referring to here in the same PR. Both were introduced in the migration PR and not here.

internal/controller/utils/nim.go
TomerFi: Resolved in https://github.com/opendatahub-io/odh-model-controller/pull/383/commits/bd22ff20f386d309f68bfb795866b9db9c4452cc.

internal/controller/utils/nim.go
TomerFi: Why? We need to log to be consistent.

internal/controller/nim/account_controller.go
spolti: please udpate it.

internal/controller/utils/nim.go
spolti: The logger you are passing as parameter has nothing special: https://github.com/opendatahub-io/odh-model-controller/pull/383/files#diff-ac9aee3dd976038fa2ac5fc8ba9dddd1f98bb8bf71e7a924926351981a35f07aR46

so, why not instantiate it here inside this func and avoid passing the logger as parameter?

TomerFi: @spolti You're looking at the hack script, see the previous comment, https://github.com/opendatahub-io/odh-model-controller/pull/383#discussion_r1975482886.

Check the code for the actual controller, the logger is proeperly instantiated by the controller, see: https://github.com/opendatahub-io/odh-model-controller/pull/383/files#diff-7617d7be763ad3db3310e0dcecb5978ebbc608f8ff2767970a01efe2fa941c70R206.

internal/controller/nim/account_controller.go
TomerFi: @Jooho About this TODO comment, I can't seem to find anywhere else we use `FromContext` at setup stage.

internal/controller/nim/account_controller.go
TomerFi: Resolved in https://github.com/opendatahub-io/odh-model-controller/pull/383/commits/835ba0b37ba8095adf1b2200bcaf990f0361491b.

internal/controller/utils/nim.go
spolti: Ah, got it, thanks, it seems I overlooked that :D

spolti (approved):

spolti:
@Jooho are you ok with these changes?

TomerFi:
@spolti @Jooho After the approval, we used an intermediate image for this PR to debug some other issue, we noticed we needed the request URI logged too, so I pushed another commit https://github.com/opendatahub-io/odh-model-controller/pull/383/commits/a2b168fcbf359cf3b0190d15486c238c68542718 and rebased.

internal/controller/nim/account_controller.go
TomerFi: Also, we use the SignalHandler context; see [here](https://github.com/opendatahub-io/odh-model-controller/blob/incubating/cmd/main.go#L170). I don't think it has a logger at the point we create it.

Jooho: In the main.go file, the way the reconciler is registered for NIM is somewhat different from the others. Is there a specific reason for this? It would be better to keep the structure consistent with the rest if possible. current structure in odh-model-controller missed to print out the error when reconciller failed(I need to follow up this) but the error log corrected in the main.go level(https://github.com/opendatahub-io/odh-model-controller/blob/incubating/cmd/main.go#L147)
However, it is not related to this pr so we can discuss it later.

Jooho (commented):
/lgtm

internal/controller/nim/account_controller.go
TomerFi: Hi @Jooho 

I switched to using the setup logger here: https://github.com/opendatahub-io/odh-model-controller/pull/391.

hack/verify_nvidia_nim_api.go
TomerFi: @spolti I refactored the hack script, see https://github.com/opendatahub-io/odh-model-controller/pull/392.

PR #380, opened 2025-02-24
revert: revert "Update status post frequency for NIM account controll…

TomerFi:
## Description

This reverts commit e705ebab3432c63fe4d521b72753b8867d1997ef. (#360 )
Frequent updates of the status had some unexpected results, fixing them requires some refactoring of the account controller, which requires some work and will block current running tasks. We will address the refactoring and the status updated in a different PR.

Jira: https://issues.redhat.com/browse/NVPE-174

## How Has This Been Tested?

Created a custom build image and deployed it in our development cluster.

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #384, opened 2025-02-28
chore: filter events for the nim controller

TomerFi:
## Description

Filter update events for the nim account controller to minimize redundant reconciliation operations.
Jira: [NVPE-91](https://issues.redhat.com/browse/NVPE-91)

## How Has This Been Tested?

I created an intermediate image and deployed it in our development environment.
I tested this by modifying the resources as well as deleting and recreating them.

## Merge criteria:

### DIFF
--- a/internal/controller/nim/account_controller.go
+++ b/internal/controller/nim/account_controller.go
@@ -104,13 +104,11 @@ func (r *AccountReconciler) SetupWithManager(mgr ctrl.Manager, ctx context.Conte
 			})).
 		WithEventFilter(predicate.Funcs{
 			UpdateFunc: func(e event.UpdateEvent) bool {
-				oldSecret, isSecret := e.ObjectOld.(*corev1.Secret)
-				if isSecret {
+				if oldSecret, isSecret := e.ObjectOld.(*corev1.Secret); isSecret {
 					newSecret := e.ObjectNew.(*corev1.Secret)
 					return !semantic.EqualitiesOrDie().DeepDerivative(oldSecret.Data, newSecret.Data)
 				}
-				oldAccount, isAccount := e.ObjectOld.(*v1.Account)
-				if isAccount {
+				if oldAccount, isAccount := e.ObjectOld.(*v1.Account); isAccount {
 					newAccount := e.ObjectNew.(*v1.Account)
 					return !semantic.EqualitiesOrDie().DeepEqual(oldAccount.Spec, newAccount.Spec)
 				}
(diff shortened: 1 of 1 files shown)


internal/controller/nim/account_controller.go
```diff
@@ -99,6 +102,21 @@ func (r *AccountReconciler) SetupWithManager(mgr ctrl.Manager, ctx context.Conte
 				}
 				return requests
 			})).
+		WithEventFilter(predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
```
Jooho, line 106: are you sure you only need the update event for the object? with this, I believe create/delete will be ignrored.

TomerFi: Are you sure? I thought it was the other way around.
When #380 gets merged, I'll rebase and verify.

zdtsw: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.19.1/pkg/predicate#TypedPredicate
if it is only overwritten Update() here, others should preserve the default logic

spolti: yes, AFAIK it will not touch the other logic.

internal/controller/nim/account_controller.go
```diff
@@ -99,6 +102,21 @@ func (r *AccountReconciler) SetupWithManager(mgr ctrl.Manager, ctx context.Conte
...
+				oldSecret, isSecret := e.ObjectOld.(*corev1.Secret)
+				if isSecret {
+					newSecret := e.ObjectNew.(*corev1.Secret)
+					return !semantic.EqualitiesOrDie().DeepDerivative(oldSecret.Data, newSecret.Data)
+				}
+				oldAccount, isAccount := e.ObjectOld.(*v1.Account)
```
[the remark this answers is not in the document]
zdtsw: probably i missed something.
if the watch is only on secret, but the filter is on secret and account both?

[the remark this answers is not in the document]
TomerFi: Resolved in https://github.com/opendatahub-io/odh-model-controller/pull/384/commits/a049d71e24e428b09f84d83c24208aee27ac5125.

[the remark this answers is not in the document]
TomerFi: @zdtsw yep. There are two if statements, one for the account and one for the secret. I think @spolti just used account one as an example of keeping the object in the context of the if statement.
I fixed it for both statements.

internal/controller/nim/account_controller.go
TomerFi: I tested this using an intermediate image in our development environment; it works as expected. I'm only filtering update events; other events work without any filtering.

spolti (approved):

internal/controller/nim/account_controller.go
Jooho: ok my bad :)

PR #390, opened 2025-03-07
fix: legally blocked content should not break NIM integration

TomerFi:
## Description

In clusters deployed in the EU, we noticed some models are considered blocked content and will return a 451 response code.
In this PR, we:
- Modified the API validation process to try with another image when hit with 451.
- Modified the metadata scraping process to exclude failing models instead of breaking the process.

Jira: https://issues.redhat.com/browse/NVPE-193

## How Has This Been Tested?

- Added a third model to the stub for runtime images. Tests assert that only two models exist without failing.
- Created an intermediate image and tested it in an EU cluster, `quay.io/tomerfi/odh-model-controller:fix-eu`. After deploying, verify the integration is enabled successfully by examining the Account status. This image was tested simultaneously in an EU and US cluster, making sure it works.

## Merge criteria:

### DIFF
--- a/internal/controller/utils/nim.go
+++ b/internal/controller/utils/nim.go
@@ -119,18 +119,27 @@ func GetAvailableNimRuntimes(logger logr.Logger) ([]NimRuntime, error) {
 }
 
 // ValidateApiKey is used for validating the given API key by retrieving the token and pulling the given custom runtime
-func ValidateApiKey(logger logr.Logger, apiKey string, runtime NimRuntime) error {
-	tokenResp, tokenErr := getRuntimeRegistryToken(logger, apiKey, runtime.Resource)
-	if tokenErr != nil {
-		return tokenErr
-	}
+func ValidateApiKey(logger logr.Logger, apiKey string, runtimes []NimRuntime) error {
+	for _, runtime := range runtimes {
+		tokenResp, tokenErr := getRuntimeRegistryToken(logger, apiKey, runtime.Resource)
+		if tokenErr != nil {
+			return tokenErr
+		}
 
-	manifestErr := attemptToPullManifest(logger, runtime, tokenResp)
-	if manifestErr != nil {
-		return manifestErr
-	}
+		manifestResp, manifestErr := attemptToPullManifest(logger, runtime, tokenResp)
+		if manifestErr != nil {
+			return manifestErr
+		}
 
-	return nil
+		switch manifestResp.StatusCode {
+		case http.StatusUnavailableForLegalReasons:
+			continue
+		case http.StatusOK:
+			return nil
+		}
+		break
+	}
+	return fmt.Errorf("failed to validate api key")
 }
 
 // GetNimModelData is used for fetching the model info for the given runtimes, returns configmap data
@@ -141,20 +150,15 @@ func GetNimModelData(logger logr.Logger, apiKey string, runtimes []NimRuntime) (
 		return data, tokenErr
 	}
 
-	var modelErr error
 	for _, runtime := range runtimes {
-		model, unmarshaled, err := getModelData(logger, runtime, tokenResp)
-		if err != nil {
-			modelErr = err
-			break
+		if model, unmarshaled, err := getModelData(logger, runtime, tokenResp); err == nil {
+			data[model.Name] = unmarshaled
 		}
-		data[model.Name] = unmarshaled
 	}
 
-	if modelErr != nil {
-		return nil, modelErr
+	if len(data) < 1 {
+		return nil, fmt.Errorf("no models found")
 	}
-
 	return data, nil
 }
 
@@ -172,11 +176,10 @@ func getNimRuntimes(logger logr.Logger, runtimes []NimRuntime, page, pageSize in
 
 	req.URL.RawQuery = query.Encode()
 
-	resp, respErr := NimHttpClient.Do(req)
+	resp, respErr := handleRequest(logger, req)
 	if respErr != nil {
 		return runtimes, respErr
 	}
-	logApiResponse(logger, req, resp)
 
 	body, bodyErr := io.ReadAll(resp.Body)
 	if bodyErr != nil {
@@ -194,22 +197,6 @@ func getNimRuntimes(logger logr.Logger, runtimes []NimRuntime, page, pageSize in
 	}
 
 	return runtimes, nil
-}
-
-func logApiResponse(logger logr.Logger, req *http.Request, resp *http.Response) {
-	logger.V(1).Info(fmt.Sprintf("sending api request %s", req.URL))
-
-	logger.V(1).Info(fmt.Sprintf("got api response %s", resp.Status))
-	if resp.StatusCode != http.StatusOK {
-		if resp.ContentLength > 0 {
-			body, err := io.ReadAll(resp.Body)
-			if err != nil {
-				logger.V(1).Error(err, "failed to parse body")
-			} else {
-				logger.V(1).Info(fmt.Sprintf("got body %s", string(body)))
-			}
-		}
-	}
 }
 
 // mapNimCatalogResponseToRuntimeList is used for parsing the ngc catalog response to a list of available runtimes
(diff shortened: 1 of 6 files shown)


internal/controller/utils/nim.go
```diff
@@ -285,56 +271,80 @@ func requestToken(logger logr.Logger, req *http.Request) (*NimTokenResponse, err
 }
 
 // attemptToPullManifest is used for pulling a runtime for verifying access
-func attemptToPullManifest(logger logr.Logger, runtime NimRuntime, tokenResp *NimTokenResponse) error {
+func attemptToPullManifest(logger logr.Logger, runtime NimRuntime, tokenResp *NimTokenResponse) (*http.Response, error) {
 	req, reqErr := http.NewRequest("GET", fmt.Sprintf(nimGetRuntimeManifestFmt, runtime.Resource, runtime.Version), nil)
 	if reqErr != nil {
-		return reqErr
+		logger.Error(reqErr, "failed to construct request")
```
spolti, line 277: imho, in this cases there is no need to log the error as it is being returned.

zdtsw:
i  guess it was unlucky that the first one happened with 451 which stopped any other model to be validated, and the fix is to skip such and continue till the end. and only use the 2XX ones.

internal/controller/utils/nim.go
```diff
@@ -119,18 +119,27 @@ func GetAvailableNimRuntimes(logger logr.Logger) ([]NimRuntime, error) {
...
+		}
 
-	return nil
+		switch manifestResp.StatusCode {
+		case http.StatusUnavailableForLegalReasons:
+			continue
+		case http.StatusOK:
+			return nil
+		}
+		break
```
zdtsw, line 140: just replace `break` with line142? 
maybe even print what code is return now for better visibility

spolti:
While these changes look good, I don't know how to test it.
Maybe,  can we have someone else from the NIM team look at it?

internal/controller/utils/nim.go
TomerFi: Good catch!
Resolved in https://github.com/opendatahub-io/odh-model-controller/pull/390/commits/ee47edf344e97e497f6604ebf90fc802cf8147a9.

TomerFi:
> i guess it was unlucky that the first one happened with 451 which stopped any other model to be validated, and the fix is to skip such and continue till the end. and only use the 2XX ones.

@zdtsw That didn't happen. The initial bug was about 451 for the model info fetch part, not the validation part. But we tested this specically as part of the bug fix and discovered that potentially, 451 will also be returned for the validation process in cases where a blocked model is the first in the list ("lucky").

TomerFi:
> While these changes look good, I don't know how to test it.
Maybe, can we have someone else from the NIM team look at it?

@spolti We have an EU that was showing this, and it was immediately fixed once pushed with the fixed image.

internal/controller/utils/nim.go
TomerFi: > just replace break with line142?
Can you clarify, please? Line 142 is outside the context of the loop.

> maybe even print what code is return now for better visibility
Good catch. We were logging it before, and I accidentally removed that code in this PR. Resolved in https://github.com/opendatahub-io/odh-model-controller/pull/390/commits/350b7059f52457fb25a5bda236b3690bef38fb98.

zdtsw: i was thinking if you wanna break the for...loop then hit the return error
you could directly return error here. 
or i misunderstood this intention?

zdtsw: in general, i am fine with this change.
i let others make the finall call 
/LGTM

TomerFi: > i was thinking if you wanna break the for...loop then hit the return error
you could directly return error here.

Oh yeah, I see what you mean now. This should work, but Go blocks the build for that, so I'll have to throw a `return nil` there. So I think what we have now is better, more clear.

internal/controller/utils/nim.go
```diff
@@ -285,56 +271,80 @@ func requestToken(logger logr.Logger, req *http.Request) (*NimTokenResponse, err
...
-
-	return nil
+	return handleRequest(logger, req)
 }
 
 // getModelData is used for fetching NIM model data for the given runtime
 func getModelData(logger logr.Logger, runtime NimRuntime, tokenResp *NimTokenResponse) (*NimModel, string, error) {
 	req, reqErr := http.NewRequest("GET", fmt.Sprintf(nimGetNgcModelDataFmt, runtime.Org, runtime.Team, runtime.Image), nil)
 	if reqErr != nil {
+		logger.Error(reqErr, "failed to construct request")
```
spolti, line 290: The caller of the getModelData in this PR is not checking returned errors, maybe this part can be improved.

TomerFi: Yes. The [caller](https://github.com/TomerFi/odh-model-controller/blob/fix-eu-content-unavailable/internal/controller/utils/nim.go#L154) uses the error as a Boolean flag, we can probably move the logger statement to the caller. This might make more sense.

TomerFi: We can't move it up to the caller. If we do, we'll end up with repeating log records. The next if statement from where you're looking, [line 297](https://github.com/TomerFi/odh-model-controller/blob/fix-eu-content-unavailable/internal/controller/utils/nim.go#L297), returns an error without logging because the handleRequest function does the logging. So, in this case, if the caller does logging as well, we'll have double records.

So wait a sec, let me see if I can refactor this, we can probably live without this error being propagated back up.

TomerFi: That was a very good catch! Resolved here: https://github.com/opendatahub-io/odh-model-controller/pull/390/commits/4c5b04d41c900cd9ca72935d75da1aa84dbcdcf9.

internal/controller/utils/nim.go
```diff
@@ -285,54 +271,78 @@ func requestToken(logger logr.Logger, req *http.Request) (*NimTokenResponse, err
...
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		sErr := fmt.Errorf("got response %s", resp.Status)
+		if resp.StatusCode == http.StatusUnavailableForLegalReasons {
+			logger.Info("content not available for legal reasons")
+		} else {
+			logger.Error(sErr, "unexpected response status code")
+		}
+		return nil, ""
```
israel-hdez, line 308: Without the errors, this is looking more like best effort for getting the data. Right?
I think the only disadvantage I see is that, in case of errors, for retrying users would need to restart odh-model-controller.

If  you are OK with it, I can approve. The rest of the code is fine for me.

TomerFi: Yes, I agree. It does look like "best effort for getting the data", but I don't think we had much of a choice here; we rely on NVIDIA servers for a lot of consecutive requests. We already have more than 80 models, we can't have one fail request botch the entire process of metadata scraping.

About the reconciliation, users won't have to restart the operator; we reconcile twice a day regardless, so we'll try again eventually. And if they can't wait, the quickest workaround is to modify the API key Secret or the Account, these actions will trigger a full reconciliation. It's also important to note that this doesn't affect running models; the result of this change will be that the models that failed the metadata scraping request will be missing from the drop-down list in the deploy-modal. Which is better than the current state of disabling the entire integration.

That said, we've been discussing refactoring the Account controller for multiple reasons. I'll keep this discussion in mind, we can consider requeuing for unknown errors.

internal/controller/utils/nim.go
spolti: yep :)
thanks

spolti (approved):
Approving, waiting others for their approval

TomerFi:
> Approving, waiting others for their approval

@spolti Based on _tide_ we need the _approved_ label as well.

internal/controller/utils/nim.go
israel-hdez: OK, all good then.

israel-hdez (approved):
/lgtm

PR #405, opened 2025-04-01
Change `Account` spec to support NIM model list selection

xieshenzh:
## Description

Allow users to create a config map with a list of IDs of the models that are supposed to be selected for deployment.
Then users can set the reference of the config map in the spec of an Account CR to enable the model selection.
The data in the config map `nimConfig` and the models in the deployment list should be updated accordingly.

Jira: https://issues.redhat.com/browse/NVPE-215
ADR: https://docs.google.com/document/d/16Ui585IE31igU6bOM-MgoNXfVmaZ76Mfxrcu7d3Lm9w/

## How Has This Been Tested?

1. Test creating Account CR without the model selection config map
2. Test creating Account CR with the model selection config map
2.1 Test updating the data in the config map
2.2 Test deleting reference to the config map from the spec of the Account CR 

Demo: https://www.youtube.com/watch?v=FX1uxKDpSvk

## Merge criteria:


internal/controller/nim/account_controller.go
```diff
@@ -383,6 +448,41 @@ func (r *AccountReconciler) reconcileNimPullSecret(
...
+
+	// get the config map that contains the selected model list
+	cmNs := cmRef.Namespace
+	if cmNs == "" {
+		cmNs = namespace
+	}
+	modelListCm := &corev1.ConfigMap{}
+	modelListCmSubject := types.NamespacedName{Name: cmRef.Name, Namespace: cmNs}
+	if err := r.Client.Get(ctx, modelListCmSubject, modelListCm); err != nil {
+		return nil, err
```
TomerFi, line 468: We should account for NotFound errors here. The statement invoking this function is requeueing for errors, we'll be stuck here forever. We should log, but return an empty slice and a nil to avoid breaking the rest of the flow.

Example scenario: a user modified the Account spec but hadn't yet created the actual ConfigMap or created it with a wrong name.

internal/controller/nim/account_controller.go
```diff
@@ -383,6 +448,41 @@ func (r *AccountReconciler) reconcileNimPullSecret(
...
+			var selectedModelIds []string
+			err := json.Unmarshal([]byte(models), &selectedModelIds)
+			if err != nil {
+				return nil, err
+			}
+			return selectedModelIds, nil
+		}
+	}
+
+	return nil, fmt.Errorf("failed to get the selected model list from the config map %+v", modelListCmSubject)
```
TomerFi, line 483: Much like the comment above, if the user had a wrong syntax in the ConfigMap, returning an error will requeueu and we'll be right back here. It will be better to log the error and not return it, so the rest of the flow will work as if this ConfigMap didn't exist.

internal/controller/nim/account_controller.go
```diff
@@ -213,6 +245,39 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	if selectedModelList, err := r.getSelectedModelList(ctx, account.Spec.ModelListConfig, account.Namespace); err != nil {
+		logger.V(1).Error(err, "failed to get the selected model list")
+		return ctrl.Result{}, err
+	} else if len(selectedModelList) > 0 {
+		if marshIds, marshErr := json.Marshal(selectedModelList); marshErr != nil {
+			logger.V(1).Error(runtimesErr, "failed to marshal the model Ids")
+			return ctrl.Result{}, marshErr
+		} else {
+			logger.V(1).Info("got the selected NIM model list", "model Ids", json.RawMessage(marshIds))
+		}
```
TomerFi, line 271: If the selected models list is created from the ConfigMap, why do we need to go to such length as marshalling again just so we could log it. Do we need the content of the ConfigMap in the logs?

TomerFi (changes requested):

internal/controller/nim/account_controller.go
```diff
@@ -213,6 +245,39 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	modelIds := make([]string, size)
+	for i := 0; i < size; i++ {
+		modelIds[i] = availableRuntimes[i].Name
+	}
+	if marshIds, marshErr := json.Marshal(modelIds); marshErr != nil {
+		logger.V(1).Error(runtimesErr, "failed to marshal the model Ids")
+		return ctrl.Result{}, marshErr
+	} else {
+		logger.V(1).Info("fetched the available NIM models", "model Ids", json.RawMessage(marshIds))
+	}
```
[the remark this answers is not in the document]
xieshenzh: It logs all the model ids as a json array. I can copy the data and quickly create config map, like what happens in the demo https://www.youtube.com/watch?v=FX1uxKDpSvk

internal/controller/nim/account_controller.go
xieshenzh: I need the content in the logs, so that I can copy the data and quickly reproduce the scenario without the actual config map.

internal/controller/nim/account_controller.go
```diff
@@ -75,7 +77,18 @@ func (r *AccountReconciler) SetupWithManager(mgr ctrl.Manager, ctx context.Conte
 	if err := mgr.GetFieldIndexer().IndexField(ctx, &v1.Account{}, apiKeySpecPath, func(obj client.Object) []string {
 		return []string{obj.(*v1.Account).Spec.APIKeySecret.Name}
 	}); err != nil {
-		logger.Error(err, "failed to set cache index")
+		logger.Error(err, "failed to set apiKey secret cache index")
+		return err
```
spolti, line 81: in returns like this, isn't it better to return the error message instead of logging it and later returning the error?
e.g.:

```go
return fmt.Errorf("failed to set apiKey secret cache index: %w", err)
```

internal/controller/nim/account_controller.go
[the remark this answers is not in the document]
spolti: In this case, can we avoid this manual step?
Why can't we fetch these model list and automatically update the configmap? Or create a new one if there is nothing created yet?

internal/controller/nim/account_controller.go
```diff
@@ -213,6 +246,39 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	} else {
+		logger.V(1).Info("fetched the available NIM models", "model Ids", json.RawMessage(marshIds))
+	}
+
+	// check the selected runtimes
+	if selectedModelList, err := r.getSelectedModelList(ctx, account.Spec.ModelListConfig, account.Namespace); err != nil {
+		logger.V(1).Error(err, "failed to get the selected model list")
+		return ctrl.Result{}, err
+	} else if len(selectedModelList) > 0 {
+		if marshIds, marshErr := json.Marshal(selectedModelList); marshErr != nil {
```
spolti, line 267: cant we export this marshal section to a func as it is used a few lines above?

internal/controller/nim/account_controller.go
```diff
@@ -383,6 +449,52 @@ func (r *AccountReconciler) reconcileNimPullSecret(
...
+	}
+
+	// convert the selected model list to a string array
+	if modelListCm.Data != nil {
+		if models, ok := modelListCm.Data["models"]; ok {
+			var selectedModelIds []string
+			if err := json.Unmarshal([]byte(models), &selectedModelIds); err != nil {
+				logger.Error(err, "failed to unmarshal the selected mode list",
+					"model list", models)
+				return []string{}, nil
```
spolti, line 487: shouldn't the error be returned here?

TomerFi: Returning an error from here, will re-queue the reconcilation request, and we'll be stuck here in a loop until the user fixes the issue with the ConfigMap. But if we log the error and return `[]string{}, nil` the controller will procceed as if the custom user ConfigMap doesn't exist.

TomerFi: See comment: https://github.com/opendatahub-io/odh-model-controller/pull/405#discussion_r2024896907

internal/controller/nim/account_controller.go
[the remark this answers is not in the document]
xieshenzh: @spolti Please take a look at the discussion in the ADR: https://docs.google.com/document/d/16Ui585IE31igU6bOM-MgoNXfVmaZ76Mfxrcu7d3Lm9w/edit?disco=AAABfRz8MTQ

internal/controller/nim/account_controller.go
spolti: ok

internal/controller/nim/account_controller.go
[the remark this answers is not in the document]
spolti: ok, I see. 
So it will be totally up tot he user configure this new configMap.
This will require things to be reconciled, right? 
Once you edit the NIM's CR, will it trigger the odh restart?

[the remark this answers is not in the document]
xieshenzh: Yes. Once the user edits the `Account` CR or edits the new ConfigMap (which is configured by the user and has the list of user-selected model Ids), it will trigger controller to reconcile.

Then, the data in the other ConfigMap `nimConfig`, which is used by dashboard for deployment NIM models, will be updated to contain only the models that are selected by the user.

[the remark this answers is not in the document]
spolti: ok, thanks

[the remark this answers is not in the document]
TomerFi: @xieshenzh Nice work on the recording, it make this clear.

But now I'm wondering, why do we need the selected-ConfigMap's models to be a string of an array?
These are just names, can we just use a standard string of names seperated by a comma?

IMO, it looks like we're doing a lot of of work just to log all models as an array so it will be easier to copy into the configmap, but it would have been just as easy to copy as a standard string. And it would be much simpler to log it as a string.

Am I missing something with the choice of using an array?

internal/controller/nim/account_controller.go
TomerFi: I think this [comment](https://github.com/opendatahub-io/odh-model-controller/pull/405#discussion_r2027249674) is also relevant here.

internal/controller/nim/account_controller.go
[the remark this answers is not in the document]
xieshenzh: Yes, I can do that. Use `strings` package to complete the operation, instead of the `json` package.

xieshenzh:
@spolti @TomerFi I have resolved the comments. Could you please take a look again? Thanks.

api/nim/v1/account_types.go
```diff
@@ -11,6 +11,8 @@ import (
 type AccountSpec struct {
 	// A reference to the Secret containing the NGC API Key.
 	APIKeySecret corev1.ObjectReference `json:"apiKeySecret"`
+	// A reference to the ConfigMap containing the list of NIM models that are allowed to be deployed.
+	ModelListConfig *corev1.ObjectReference `json:"modelListConfig,omitempty"`
```
TomerFi, line 15: Maybe we should set a default name instead if _omitempty_, so admins will have to create the ConfigMap but can skip updating the Account?

internal/controller/nim/account_controller.go
```diff
@@ -383,6 +440,52 @@ func (r *AccountReconciler) reconcileNimPullSecret(
...
+			return []string{}, nil
+		}
+		return nil, err
+	}
+
+	// convert the selected model list to a string array
+	if modelListCm.Data != nil {
+		if models, ok := modelListCm.Data["models"]; ok {
+			var selectedModelIds []string
+			if err := json.Unmarshal([]byte(models), &selectedModelIds); err != nil {
```
TomerFi, line 475: If we use a comma-seperated string, should we still unmarshall?

api/nim/v1/account_types.go
xieshenzh: I don't think so.
There will be an error, when it is set in the spec, but the config map doesn't exist. Although, the controller will not requeue in this case.
I don't think we should leave an `Account` in an erroneous situation by setting a default value in the spec. 

You suggested to not requeue when there is anything wrong with the config map, in order to avoid infinite loop. And I did that. 
But, IMHO, it should be requeued if there is anything wrong with the spec. That's how the controller is supposed to work.
The dashboard should deal with the infinite loop, not the backend. But I don't want to insist on it.

internal/controller/nim/account_controller.go
xieshenzh: I am still using json array. 
But I don't use json marshal to create the logs any more, to avoid additional complexity as you mentioned in the comments.

IMHO, it is easier to edit 100 elements in a json array with a json editor, than editing a long comma delimited string.

api/nim/v1/account_types.go
TomerFi: IMO if we change this [log record](https://github.com/xieshenzh/odh-model-controller/blob/nim-model-list/internal/controller/nim/account_controller.go#L464) from _Error_ to _Info_, we should be good with setting default file name, and if we don't already cache this ConfigMap, we can do that too.

Anyway, it was just a thought. Feel free to resolve this if you disagree. :-)

internal/controller/nim/account_controller.go
TomerFi: Gotcha.

xieshenzh:
@spolti @zdtsw @Jooho @israel-hdez 
We would like to target the PR for 2.20.
I have addressed the change requests.
Could you please take a look again? Thanks.

PR #413, opened 2025-04-15
Adding Both Custom and Trusted Certs to Kserve CA Configmap

brettmthompson:
## Description
[RHOAIENG-21641](https://issues.redhat.com/browse/RHOAIENG-21641)

This PR updates the ConfigMapReconciler to add both the cluster-wide and custom CA bundles from the odh-trusted-ca-bundle configmap to the odh-kserve-custom-ca-bundle configmap. 
 

We cannot use the `config.openshift.io/inject-trusted-cabundle` annotation to inject the cluster-wide CA bundle into the `odh-kserve-custom-ca-bundle` configmap. Per the [RHOAI](https://docs.redhat.com/en/documentation/red_hat_openshift_ai_self-managed/2.8/html/installing_and_uninstalling_openshift_ai_self-managed/working-with-certificates_certs#how_ca_bundles_are_injected) documentation, this annotation will update the configmap with the `ca-bundle.crt` key. However, [kserve](https://kserve.github.io/website/0.13/modelserving/certificate/kserve/) expects the key in the `odh-kserve-custom-ca-bundle` configmap to be `cabundle.crt`. Therefore, if any custom CA bundle certs are present in the `odh-trusted-ca-bundle` configmap, we combine the custom CA bundle certs with the cluster-wide CA bundle certs when adding them to the `odh-kserve-custom-ca-bundle` configmap.

At the same time, we keep the existing behavior of only including the custom CA bundle certs in the certificate for ModelMesh in the `storage-config` secret, by pulling the custom CA bundle cert values directly from the `odh-trusted-ca-bundle` configmap.

In addition to the core changes listed above, this PR also fixes existing unit tests in the `secret_controller_test.go` file. These tests target scenarios in which the `odh-kserve-custom-ca-bundle` configmap is updated and check that the `storage-config` secret is also updated to reflect the changes. **This is not the proper behavior of the odh-model-controller**. The odh-model-controller will always reconcile any changes made to the `odh-kserve-custom-ca-bundle` configmap and restore it to only reflect the values found in the `odh-trusted-ca-bundle` configmap. Therfore, these tests should target scenarios where the `odh-trusted-ca-bundle` configmap is updated, not the `odh-kserve-custom-ca-bundle` configmap.

## How Has This Been Tested?

1. In a cluster with the opendatahub operator installed, create a default DSCI resource.
2. Replace the odh-model-controller image with an image built from this branch.
3. No `odh-kserve-custom-ca-bundle` configmaps should exist.
4. Add a custom CA cert to the DSCI resource like below:
```
spec:
  trustedCABundle:
    managementState: Managed
    customCABundle: |
      -----BEGIN CERTIFICATE-----
      examplebundle123
      -----END CERTIFICATE-----
```
5. An `odh-kserve-custom-ca-bundle` should be created in each opendatahub-operator managed namespace that includes all cluster-wide and custom CA certs under the key `cabundle.crt`.
6. Remove the custom CA cert from the DSCI resource like below:
```
spec:
  trustedCABundle:
    managementState: Managed
    customCABundle: ''
```
7. All `odh-kserve-custom-ca-bundle` configmaps should be deleted.

## Merge criteria:


Jooho (approved):
Great job! It looks good to me. 

/approve

PR #428, opened 2025-05-05
[RHOAIENG-21207] - Publish the Grafana Dashboard to the ODH repository

spolti:
## Description

It provides the basic instructions to have the Grafana Dashboard deployed in your cluster to plot vLLM and NVIDIA accelerator metrics.

It assumes that you already have Prometheus and Grafana running in your cluster.

## How Has This Been Tested?

It was tested in a cluster with Grafana configured by following the commands described in the newly added README.md file.

## Merge criteria:

### DIFF
--- a/monitoring/README.md
+++ b/monitoring/README.md
@@ -41,7 +41,8 @@ so it can correctly show the metrics deployed in your current namespace for the 
 There are two parameters that needs to be updated to reflect it:
 
 - **NAMESPACE**: the target namespace where the model will be deployed.
-- **MODEL_NAME**: the model name as defined in your InferenceService, note, it is a regex, so just the starting string of your model name would be enough.
+- **MODEL_NAME**: the model name as defined in your InferenceService, note that, it will also be used to filter the pod name
+  in the Grafana dashboard, which is a filter using regex.
 
 To correctly override it, we will use a helper called `envsubst`, it is available through the `gettext` GNU package.
 
@@ -49,6 +50,8 @@ We will use the `envsubst` command to replace the variables in the JSON string w
 [inputs.env](inputs.env) file, example:
 
 ```bash
+# the following command reads the environment variables from the inputs.env file and exports them to the current shell
+# so it can be used by the envsubst command
 export $(cat config/grafana/nvidia/inputs.env | xargs)
 envsubst '${NAMESPACE} ${MODEL_NAME}' < grafana/nvidia/nvidia-vllm-dashboard.yaml > /tmp/nvidia-vllm-dashboard-replaced.yaml
 ```
@@ -59,7 +62,9 @@ NAMESPACE=granite
 MODEL_NAME=granite318b
 ```
 
-The part that needs most attention is:
+These values will be replaced by its placeholders in the [dashboard](grafana/nvidia/nvidia-vllm-dashboard.yaml), in the end of
+the generated file, pay attention to this section, where the `namespace` and `model_name` inputs are defined:
+
 ```json
        {
             "current": {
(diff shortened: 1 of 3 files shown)


monitoring/README.md
```diff
@@ -0,0 +1,98 @@
...
+        ]
+      }
+...
+#remaining of the JSON definition
+```
+
+With this approach, it makes difficult to use kustomize to correctly override the needed parameters so the Dashboard
+can correctly show the metrics deployed in your current namespace. There are two parameters that needs to be updated:
+
+- **NAMESPACE**: the targe namespace where the target model will be deployed.
```
hdefazio, line 41:
```suggestion
- **NAMESPACE**: the target namespace where the target model will be deployed.
```

monitoring/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+Notice that the model_name and namespace dashboards variables were updated to the values from `inputs.env` file.
+
+Now deploy the Dashboard:
+```bash
+envsubst '${NAMESPACE} ${MODEL_NAME}' < grafana/nvidia/nvidia-vllm-dashboard.yaml > /tmp/nvidia-vllm-dashboard-replaced.yaml
+```    
+
+
+Note, ti does not covers Grafana or Prometheus configurations, this is just the dashboard object that will be deployed in the Grafana instance.
```
hdefazio, line 97:
```suggestion
Note, it does not covers Grafana or Prometheus configurations, this is just the dashboard object that will be deployed in the Grafana instance.
```

monitoring/README.md
```diff
@@ -0,0 +1,98 @@
...
+- **NAMESPACE**: the targe namespace where the target model will be deployed.
+- **MODEL_NAME**: the same as defined in your InferenceService, note, it is a regex, so just the starting string of your model name would be enough.
+
+To correctly override it, we will use a helper called `envsubst`, it is available through the `gettext` GNU package.
+
+We will use the `envsubst` command to replace the variables in the JSON string with the actual values provided in the 
+[inputs.env]() file, example:
+
+```bash
+export $(cat config/grafana/nvidia/inputs.env | xargs)
```
hdefazio, line 50: Can you add a quick sentence describing what this does? It reads a bit confusingly when you say we're going to use `envsubt` and then the next command doesn't do that.

monitoring/README.md
```diff
@@ -0,0 +1,98 @@
...
+#remaining of the JSON definition
+```
+
+With this approach, it makes difficult to use kustomize to correctly override the needed parameters so the Dashboard
+can correctly show the metrics deployed in your current namespace. There are two parameters that needs to be updated:
+
+- **NAMESPACE**: the targe namespace where the target model will be deployed.
+- **MODEL_NAME**: the same as defined in your InferenceService, note, it is a regex, so just the starting string of your model name would be enough.
+
+To correctly override it, we will use a helper called `envsubst`, it is available through the `gettext` GNU package.
```
hdefazio, line 44: Should we add an install command?

hdefazio (changes requested):

monitoring/README.md
spolti: I did update it before your review, sry

monitoring/README.md
spolti: here as well.

monitoring/README.md
spolti: sure.

monitoring/README.md
hdefazio: true, I just think the requirement is a bit hidden without an install command to draw attention to it - what do we usually do for non python requirements?

hdefazio (approved):

monitoring/README.md
spolti: It depends. Usually, we add some reference to where that command (in this case, GNU) can be found, and the user installs it based on their system.

spolti:
replaced by https://github.com/rh-aiservices-bu/rhoai-uwm/pull/3

PR #447, opened 2025-05-16
RHOAIENG-25652: Bump KServe to 0.15

pierDipi:
## Description

- Bump KServe dependency to 0.15
- Bump KServe CRDs to 0.15

## How Has This Been Tested?
`make test`

## Merge criteria:


go.mod
```diff
@@ -1,150 +1,157 @@
 module github.com/opendatahub-io/odh-model-controller
 
-go 1.22.7
-
-toolchain go1.22.9
+go 1.23.6
```
spolti, line 3: I think @VedantMahabaleshwarkar opened a PR for it, may we remove it from here?
Or close his PR?

Makefile
```diff
@@ -44,7 +46,16 @@ help: ## Display this help.
 
 ##@ Development
 
-.PHONY: manifests
+.PHONY: manifests-update
```
spolti, line 49: Nice :)

go.mod
pierDipi: Rebased

Makefile
```diff
@@ -81,7 +92,8 @@ vet: ## Run go vet against code.
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" POD_NAMESPACE=default \
-		MESH_NAMESPACE=istio-system CONTROL_PLANE_NAME=istio-system go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
+		MESH_NAMESPACE=istio-system CONTROL_PLANE_NAME=istio-system go test -count=1 $$(go list ./... | grep -v /e2e) \
```
bartoszmajsak, line 95: If `-count=1` (as well as changes to code coverage) is important, that should be another PR explaining why :) Otherwise, it will be just lost under the cover of the innocent dep upgrade commit :)

Makefile
```diff
@@ -44,8 +46,17 @@ help: ## Display this help.
...
 ##@ Development
 
+.PHONY: manifests-update
+manifests-update:
+	wget -O - https://raw.githubusercontent.com/kserve/kserve/$(KSERVE_MANIFESTS_REVISION)/config/crd/full/serving.kserve.io_inferencegraphs.yaml \
+		| tail -n +2 > config/crd/external/serving.kserve.io_inferencegraphs.yaml
+	wget -O - https://raw.githubusercontent.com/kserve/kserve/$(KSERVE_MANIFESTS_REVISION)/config/crd/full/serving.kserve.io_inferenceservices.yaml \
+		| tail -n +2 > config/crd/external/serving.kserve.io_inferenceservices.yaml
+	wget -O - https://raw.githubusercontent.com/kserve/kserve/$(KSERVE_MANIFESTS_REVISION)/config/crd/full/serving.kserve.io_servingruntimes.yaml \
+		| tail -n +2 > config/crd/external/serving.kserve.io_servingruntimes.yaml
```
bartoszmajsak, line 56: I like this thinking a lot. Automate all the things! It deserves its own PR in my eyes. It's an improvement to a build process which is out of scope of simply bumping KServe :)

Another thing to note is that this will be executed every time, regardless of the actual changes to the revision we rely on. This might (or might not) be a problem if we ever move it to Konflux and use "hermetic builds". From the top of my head some "crd.sum" check of some sort might help (or even to rely on built-in Make checks with some guard files instead).

bartoszmajsak (commented):
Not requesting changes per se, but wanted to bring up two things below.

Makefile
bartoszmajsak: hermetic - offline

pierDipi:
@bartoszmajsak removed Makefile changes, I'll open follow up PRs once this is merged

Makefile
pierDipi: > Another thing to note is that this will be executed every time, regardless of the actual changes to the revision we rely on. This might (or might not) be a problem if we ever move it to Konflux and use "hermetic builds". From the top of my head some "crd.sum" check of some sort might help (or even to rely on built-in Make checks with some guard files instead).

I think we can follow up at optimizing when, I believe this one is already an improvement, at least to have a consistent revision from which CRDs are taken, currently they are all based on different upstream revisions and it's too easy to miss an update

Makefile
pierDipi: https://github.com/opendatahub-io/odh-model-controller/pull/449

Makefile
pierDipi: https://github.com/opendatahub-io/odh-model-controller/pull/450 (also with the guarding)

bartoszmajsak: I was thinking of the same - let's bring the automation/unification first, as it's valuable on its own, and then we can improve.

bartoszmajsak (approved):

spolti (approved):
/lgtm

PR #452, opened 2025-05-20
RHOAIENG-25111: Add KEDA reconciler for raw deployment mode

pierDipi:
## Description

Integrate KEDA-based autoscaling for KServe InferenceServices in RawDeployment mode.

This allows ISVCs to autoscale on custom Prometheus metrics via KEDA, with secure OpenShift Monitoring access. The reconciler automates KEDA/RBAC resource lifecycle.

Key changes:
- New KServeKEDAReconciler manages KEDA-specific resources (ServiceAccount, Secret, Role, RoleBinding, TriggerAuthentication) for Prometheus-based autoscaling.
  - Creates resources if ISVC uses Prometheus external metric.
  - Adds ISVC as non-controlling owner to shared resources.
  - Removes ISVC owner reference if KEDA metrics unused or ISVC deleted.
  - Cleans up KEDA resources from namespace if no ISVCs use KEDA.
- InferenceService controller now:
  - Watches KEDA TriggerAuthentication, Role, RoleBinding.
  - Uses MatchEveryOwner for ServiceAccount, Secret for sharing.
  - Conditionally watches TriggerAuthentication if CRD present.
- KEDA API (keda.sh/v1alpha1) registered with scheme.
- A few tests added for KServe KEDA reconciler, covering:
  - Creation/ownership of KEDA resources for multiple ISVCs.
  - Owner reference removal on ISVC deletion/metric config change.
  - Handling updates to managed KEDA resources.
  - Scenarios with shared/unrequired KEDA resources.
  - Namespace cleanup.

## How Has This Been Tested?
`make test`

## Merge criteria:

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -78,6 +78,18 @@ rules:
   - list
   - watch
 - apiGroups:
+  - keda.sh
+  resources:
+  - triggerauthentications
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
   - maistra.io
   resources:
   - servicemeshcontrolplanes
@@ -105,6 +117,15 @@ rules:
   - list
   - patch
   - update
+  - watch
+- apiGroups:
+  - metrics.k8s.io
+  resources:
+  - nodes
+  - pods
+  verbs:
+  - get
+  - list
   - watch
 - apiGroups:
   - monitoring.coreos.com
@@ -183,8 +204,8 @@ rules:
   - rbac.authorization.k8s.io
   resources:
   - clusterrolebindings
-  - roles
   - rolebindings
+  - roles
   verbs:
   - create
   - delete
@@ -301,24 +322,3 @@ rules:
   - patch
   - update
   - watch
-- apiGroups:
-    - keda.sh
-  resources:
-    - triggerauthentications
-  verbs:
-    - create
-    - delete
-    - get
-    - list
-    - patch
-    - update
-    - watch
-- apiGroups:
-    - metrics.k8s.io
-  resources:
-    - pods
-    - nodes
-  verbs:
-    - get
-    - list
-    - watch
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -183,6 +183,7 @@ rules:
   - rbac.authorization.k8s.io
   resources:
   - clusterrolebindings
+  - roles
   - rolebindings
   verbs:
   - create
@@ -300,3 +301,24 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+    - keda.sh
+  resources:
+    - triggerauthentications
+  verbs:
+    - create
+    - delete
+    - get
+    - list
+    - patch
+    - update
+    - watch
+- apiGroups:
+    - metrics.k8s.io
+  resources:
+    - pods
+    - nodes
+  verbs:
+    - get
+    - list
+    - watch
--- a/internal/controller/serving/inferenceservice_controller_test.go
+++ b/internal/controller/serving/inferenceservice_controller_test.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"time"
 
+	kedaapi "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
 	kservev1alpha1 "github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
 	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
 	authorinov1beta2 "github.com/kuadrant/authorino/api/v1beta2"
@@ -31,6 +32,7 @@ import (
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gstruct"
 	gomegatypes "github.com/onsi/gomega/types"
+	"github.com/opendatahub-io/odh-model-controller/internal/controller/serving/reconcilers"
 	routev1 "github.com/openshift/api/route/v1"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	istioclientv1beta1 "istio.io/client-go/pkg/apis/networking/v1beta1"
@@ -39,12 +41,15 @@ import (
 	v1 "k8s.io/api/networking/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apiserver/pkg/storage/names"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/utils/ptr"
 	"knative.dev/pkg/apis"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 	maistrav1 "maistra.io/api/core/v1"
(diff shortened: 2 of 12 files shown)


bartoszmajsak (changes requested):
I have more through review in the making, but this thing needs a small fix :)

config/rbac/role.yaml
```diff
@@ -300,3 +301,24 @@ rules:
...
+    - watch
+- apiGroups:
+    - metrics.k8s.io
+  resources:
+    - pods
+    - nodes
+  verbs:
+    - get
+    - list
+    - watch
```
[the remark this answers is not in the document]
pierDipi: Done, I would've expected a test to verify codegen though

bartoszmajsak (commented):
Just a few "cosmetic" things, but overall LGTM!

internal/controller/serving/inferenceservice_controller_test.go
```diff
@@ -1289,6 +1294,567 @@ var _ = Describe("InferenceService Controller", func() {
 			})
 		})
 	})
+
+	Describe("KServe KEDA Reconciler", func() {
+		var (
+			testNs         string
+			kedaReconciler *reconcilers.KserveKEDAReconciler
+			ctx            context.Context
```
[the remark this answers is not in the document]
bartoszmajsak: Just as a side note - having global ctx as we have at the moment is not really the best approach.

internal/controller/serving/inferenceservice_controller_test.go
```diff
@@ -1289,6 +1294,567 @@ var _ = Describe("InferenceService Controller", func() {
...
+		Context("when InferenceServices are configured to scale on KEDA metrics", func() {
+			var isvc *kservev1beta1.InferenceService
+			var isvc2 *kservev1beta1.InferenceService
+
+			const fake = "fake" // for the love of the linter :)
+
+			BeforeEach(func() {
+				isvc = makeKedaTestISVC(testNs, names.SimpleNameGenerator.GenerateName("keda-isvc"), true)
+				Expect(k8sClient.Create(ctx, isvc)).Should(Succeed())
+				Expect(k8sClient.Get(ctx, types.NamespacedName{Name: isvc.Name, Namespace: isvc.Namespace}, isvc)).Should(Succeed())
```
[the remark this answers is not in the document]
pierDipi: hmm, is it eventually? I don't expect it to be eventually consistent as it's using a real client ?

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
```diff
@@ -0,0 +1,629 @@
...
+		} else {
+			newReferences = append(newReferences, ref)
+		}
+	}
+
+	if !found {
+		newReferences = append(newReferences, expected)
+	}
+	return newReferences
+}
```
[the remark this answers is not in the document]
pierDipi: I'm not sure if the client is returning cached (shared) objects and mutability might cause races, we can afford copies here, so I'd leave it at staying immutable?

https://github.com/opendatahub-io/odh-model-controller/blob/a61aea930e97eb1146310589ac094321cbc08408/cmd/main.go#L264-L282

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
```diff
@@ -0,0 +1,629 @@
...
+		return fmt.Errorf("failed to reconcile trigger authentication: %w", err)
+	}
+	log.Info("Successfully reconciled KEDA resources")
+	return nil
+}
+
+func (k *KserveKEDAReconciler) Delete(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+
+	log = log.WithName("KserveKEDAReconciler")
+	log.V(2).Info("KserveKEDAReconciler.Delete called, removing owner reference.")
```
brettmthompson, line 131: Does it make sense to move the "removing owner reference" part of this log into the `removeOwnerRef` method?

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
```diff
@@ -0,0 +1,629 @@
...
+			// There are still InferenceServices with Prometheus autoscaling configured, nothing to remove.
+			return nil
+		}
+	}
+	return k.cleanupNamespace(ctx, log, namespace)
+}
+
+func (k *KserveKEDAReconciler) cleanupNamespace(ctx context.Context, log logr.Logger, namespace string) error {
+	log = log.WithName("KserveKEDAReconciler")
+	log.V(2).Info("KserveKEDAReconciler.Cleanup called.", "Namespace", namespace)
```
brettmthompson, line 159: I think this log could be a little misleading since this method is called by other methods other than `Cleanup`, such as `Delete`. Maybe it would be better as just "Cleaning up keda related resources"?

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
```diff
@@ -0,0 +1,629 @@
...
+// --- TriggerAuthentication ---
+func (k *KserveKEDAReconciler) reconcileTriggerAuthentication(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+	expected := k.expectedTriggerAuthentication(isvc)
+	curr := &kedaapi.TriggerAuthentication{}
+	key := client.ObjectKey{Namespace: expected.Namespace, Name: expected.Name}
+
+	err := k.client.Get(ctx, key, curr)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			log.Info("Creating TriggerAuthentication")
```
brettmthompson, line 183: Wdyt about emitting the logs for creating resources within their creation methods instead of in the calling method like it is now. Also done on lines 250, 303, 370, and 435.

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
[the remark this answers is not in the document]
bartoszmajsak: By default every object that we get `Get` from the ctrl-runtime informer cache is deep copied (there's a flag though called `UnsafeDisableDeepCopy` that let you open the pandora box). As discussed offline - perhaps it's not worth making such micro optimization.

Reference https://github.com/kubernetes-sigs/controller-runtime/blob/71f7db556ca57ce7ea6563f77d739f0d2a54233a/pkg/cache/internal/cache_reader.go#L84-L90

internal/controller/serving/inferenceservice_controller_test.go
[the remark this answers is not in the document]
pierDipi: ahah, all good, better to double check these things

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
```diff
@@ -0,0 +1,629 @@
...
+	if err := k.client.Update(ctx, obj); err != nil && !errors.IsNotFound(err) && !meta.IsNoMatchError(err) {
+		return fmt.Errorf("failed to update %s %s after removing owner reference: %w", obj.GetObjectKind(), key.String(), err)
+	}
+	log.Info("Successfully removed ISVC owner reference and updated resource",
+		"resourceKind", obj.GetObjectKind(), "resourceName", obj.GetName(), "namespace", obj.GetNamespace())
+
+	return nil
+}
+
+func AsOwnerRef(isvc *kservev1beta1.InferenceService) metav1.OwnerReference {
```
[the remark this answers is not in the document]
brettmthompson: I could see some benefit to adding methods similar to this one to the kserve repo itself as struct methods associated with the InferenceService struct. Then we could just call .AsOwnerRef() on inference service objects.

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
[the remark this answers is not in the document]
brettmthompson: @bartoszmajsak I think in the case of your suggestion, the `refs` variable will be pointing to the same underlying array in memory as the `obj` variable's owner references. So when you set `refs[i] = expected` it will update the owner references of the client object as well. This would then lead to issues with the reconciliation of resources since `equality.Semantic.DeepDerivative(expected.OwnerReferences, curr.OwnerReferences)` would then return true even when there is a change made to the owner references.

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
brettmthompson: Additionally, does it make sense to decrease the verbosity level of any logs with level 2? I currently see level 2 only used in one other place, with all other logs being level 0 or 1. Right now no logs will be emitted for keda reconciler deletion calls unless the level is 2.

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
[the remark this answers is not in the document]
bartoszmajsak: This is what I was thinking of too, though we can have just one function for that.

```golang
func AsOwnerRef(owner client.Object) metav1.OwnerReference {
	gvk := owner.GetObjectKind().GroupVersionKind()
	ownerRef := *metav1.NewControllerRef(owner, gvk)
	// Setting Controller to false makes this a non-controlling owner reference.
	// The owned object will be garbage collected when the owner is deleted,
	// but the ISVC doesn't "control" its lifecycle in other controller-runtime senses.
	ownerRef.Controller = ptr.To(false)
	return ownerRef
}
````

File: internal/controller/serving/reconcilers/kserve_keda_reconciler.go (hunk 4 above)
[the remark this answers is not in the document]
Comment (bartoszmajsak) [reply internal/controller/serving/reconcilers/kserve_keda_reconciler.go:606]: Good catch! I missed that.

File: internal/controller/serving/reconcilers/kserve_keda_reconciler.go [hunk 9]
```diff
@@ -0,0 +1,629 @@
...
+		client: client,
+	}
+}
+
+func (k *KserveKEDAReconciler) Reconcile(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
+
+	log = log.WithName("KserveKEDAReconciler")
+	log.V(2).Info("Reconciling InferenceService", "InferenceService", isvc)
+
+	if !hasPrometheusExternalAutoscalingMetric(isvc) {
```
spolti, line 96: so, it is only enabled when the autoscaler has Prometheus-based metrics?

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
spolti: It seems that the keda resources are tied to multiple isvcs, is this correct?
If so, is this safe when talking about service accounts and role bindings?

Apparently it does not, since the keda resources are to properly allow the given metric to be queried, just want to double check.

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
```diff
@@ -0,0 +1,629 @@
...
+					Parameter: "ca",
+					Name:      KEDAPrometheusAuthTriggerSecretName,
+					Key:       "ca.crt",
+				},
+			},
+		},
+	}
+}
+
+// --- ServiceAccount ---
```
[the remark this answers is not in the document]
spolti: I think that is a good idea.

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
```diff
@@ -0,0 +1,629 @@
...
+	or := *metav1.NewControllerRef(isvc, kservev1beta1.SchemeGroupVersion.WithKind("InferenceService"))
+	// Setting Controller to false makes this a non-controlling owner reference.
+	// The owned object will be garbage collected when the ISVC is deleted,
+	// but the ISVC doesn't "control" its lifecycle in other controller-runtime senses.
+	or.Controller = ptr.To(false)
+	return or
+}
+
+func hasPrometheusExternalAutoscalingMetric(isvc *kservev1beta1.InferenceService) bool {
+
```
spolti, line 576:
```suggestion
```

internal/controller/serving/inferenceservice_controller_test.go
```diff
@@ -1289,6 +1294,567 @@ var _ = Describe("InferenceService Controller", func() {
...
+				}
+			})
+
+			It("Must update managed role when it diverges", func() {
+				role := &rbacv1.Role{}
+				Expect(k8sClient.Get(ctx, types.NamespacedName{Namespace: isvc.Namespace, Name: reconcilers.KEDAPrometheusAuthMetricsReaderRoleName}, role)).Should(Succeed())
+				role.Rules[0].Resources = []string{fake}
+				Expect(k8sClient.Update(ctx, role)).To(Succeed())
+
+				Expect(kedaReconciler.Reconcile(ctx, GinkgoLogr, isvc)).To(Succeed())
```
Jooho, line 1584: I see that this pattern of directly calling Reconcile is used in several places. My impression is that while it helps tests run quickly and deterministically, it might not catch potential conflicts or race conditions that can occur when reconciles run concurrently or interact asynchronously with other controllers.

Given that, I wonder if we might be missing some concurrency or interaction issues that only show up in a more realistic environment. What do you think about complementing these tests with some integration-style tests that run the manager and allow reconciles to trigger automatically? That way, we might be able to catch those subtle race conditions or conflicts better.

Would love to hear your thoughts on this approach.

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
```diff
@@ -0,0 +1,629 @@
...
+			SecretTargetRef: []kedaapi.AuthSecretTargetRef{
+				{
+					Parameter: "bearerToken",
+					Name:      KEDAPrometheusAuthTriggerSecretName,
+					Key:       "token",
+				},
+				{
+					Parameter: "ca",
+					Name:      KEDAPrometheusAuthTriggerSecretName,
+					Key:       "ca.crt",
```
spolti, line 234: We need to document it so user don't miss the needed keys.

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
```diff
@@ -0,0 +1,629 @@
...
+		return fmt.Errorf("failed to get ServiceAccount %s: %w", key.String(), err)
+	}
+
+	expected.OwnerReferences = upsertOwnerReference(AsOwnerRef(isvc), curr)
+	expected.ResourceVersion = curr.ResourceVersion
+
+	if equality.Semantic.DeepDerivative(expected.Labels, curr.Labels) &&
+		equality.Semantic.DeepDerivative(expected.Annotations, curr.Annotations) &&
+		equality.Semantic.DeepDerivative(expected.OwnerReferences, curr.OwnerReferences) {
+		log.V(1).Info("ServiceAccount is up-to-date")
```
spolti, line 262: I guess messages like this could be debug level.

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
```diff
@@ -0,0 +1,629 @@
...
+		return nil
+	}
+
+	log.Info("Updating ServiceAccount")
```
spolti, line 266: these too

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
```diff
@@ -0,0 +1,629 @@
...
+			log.Info("Creating Secret", "namespace", key.Namespace, "name", key.Name)
+			return k.createSecret(ctx, isvc)
+		}
+		return fmt.Errorf("failed to get Secret %s: %w", key.String(), err)
+	}
+
+	expected.OwnerReferences = upsertOwnerReference(AsOwnerRef(isvc), curr)
+	expected.ResourceVersion = curr.ResourceVersion
+
+	if expected.Type == curr.Type &&
```
spolti, line 312: For the secret, we could also check the fields that are necessary, like the `ca.crt` and the `token`, sothe  user can be notified if he misses it.

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
```diff
@@ -0,0 +1,629 @@
...
+	return nil
+}
+
+func (k *KserveKEDAReconciler) expectedRole(isvc *kservev1beta1.InferenceService) *rbacv1.Role {
+	return &rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            KEDAPrometheusAuthMetricsReaderRoleName,
+			Namespace:       isvc.Namespace,
+			OwnerReferences: []metav1.OwnerReference{AsOwnerRef(isvc)},
+			Labels:          getKedaLabels(isvc.Labels),
```
spolti, line 408: Is it a good idea to copy all labels from isvc?

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
```diff
@@ -0,0 +1,629 @@
...
+
+const (
+	KEDAResourcesPrefix                            = "inference-"
+	KEDAPrometheusPrefix                           = KEDAResourcesPrefix + "prometheus-"
+	KEDAPrometheusAuthResourceName                 = KEDAPrometheusPrefix + "auth"
+	KEDAPrometheusAuthServiceAccountName           = KEDAPrometheusAuthResourceName
+	KEDAPrometheusAuthTriggerSecretName            = KEDAPrometheusAuthResourceName
+	KEDAPrometheusAuthMetricsReaderRoleName        = KEDAPrometheusAuthResourceName
+	KEDAPrometheusAuthMetricsReaderRoleBindingName = KEDAPrometheusAuthResourceName
+	KEDAPrometheusAuthTriggerAuthName              = KEDAPrometheusAuthResourceName
```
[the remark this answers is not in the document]
pierDipi: This is mostly for decoupling the value and the implementation from the constant(s)

internal/controller/serving/inferenceservice_controller_test.go
```diff
@@ -1289,6 +1294,567 @@ var _ = Describe("InferenceService Controller", func() {
...
+obj: %#v
+isvc: %#v
+`, kind, obj, isvc))
+					Expect(hasOwnerReferenceByUID(obj, isvc2.UID)).To(BeTrue(), fmt.Sprintf(`
+%s should be owned by ISVC
+obj: %#v
+isvc: %#v
+`, kind, obj, isvc2))
+				}
+			})
```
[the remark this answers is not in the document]
pierDipi: yes, it's a fancy Go multiline string and made the output clearer to read as the `obj` is quite large and deserves its own line, using newlines `\n` works but it's less readable in my opinion

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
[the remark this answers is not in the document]
pierDipi: > split files so each resource we reconcile sits in their own file

I can move to a separate package, but each resource per file looks too scattered ?

internal/controller/serving/inferenceservice_controller_test.go
```diff
@@ -1289,6 +1294,567 @@ var _ = Describe("InferenceService Controller", func() {
 			})
 		})
 	})
+
+	Describe("KServe KEDA Reconciler", func() {
```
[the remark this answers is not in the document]
pierDipi: Yes, agreed, I think we can go capture these improvements as issues or follow up PRs

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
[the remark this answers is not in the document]
pierDipi: the only concern I have is that I've seen objects returning empty when calling GetObjectKind().GroupVersionKind() as TypeMeta is usually empty on Go objects

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
pierDipi: Yes, in the other cases, we don't really need it

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
pierDipi: Mind the gap :)

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
pierDipi: There is no debug level, maybe you mean level 2+

pierDipi: Used level 2

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
pierDipi: this is all automatically done by OCP

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
pierDipi: Done

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
spolti: haha nice gif!

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
spolti: nice, didn't know about it.

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
spolti: > There is no debug level, maybe you mean level 2+?

yes

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
pierDipi: Adjusted the log text and added another one for clean up action.

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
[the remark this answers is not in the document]
bartoszmajsak: > GetObjectKind().GroupVersionKind() as TypeMeta is usually empty on Go objects

Right, seen that too. Param is good, though it's unfortunate shortcoming.

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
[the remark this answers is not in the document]
bartoszmajsak: If that's more involved it's worth tackling separately. I will capture it.

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
spolti: cool, thanks for confirming.

test/matchers/gomega_uid_matcher.go
bartoszmajsak, line 1: thank-you

internal/controller/serving/reconcilers/kserve_keda_reconciler.go
```diff
@@ -0,0 +1,596 @@
...
+	return k.maybeCleanupNamespace(ctx, log, isvc.GetNamespace())
+}
+
+func (k *KserveKEDAReconciler) Cleanup(ctx context.Context, log logr.Logger, isvcNs string) error {
+	log = log.WithName("KserveKEDAReconciler")
+	log.V(2).Info("KserveKEDAReconciler.Cleanup called.", "namespace", isvcNs)
+	return k.cleanupNamespace(ctx, log, isvcNs)
+}
+
+func (k *KserveKEDAReconciler) maybeCleanupNamespace(ctx context.Context, log logr.Logger, namespace string) error {
```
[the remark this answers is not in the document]
pierDipi: `maybe...` is to convey that only in some cases the resources in a given namespace will be removed

internal/controller/serving/reconcilers/meta.go
```diff
@@ -0,0 +1,38 @@
...
+	"k8s.io/utils/ptr"
+)
+
+// AsOwnerRef returns the non-controlling OwnerReference representation of the given object.
+func AsOwnerRef(obj metav1.Object, gvk schema.GroupVersionKind) metav1.OwnerReference {
+	or := *metav1.NewControllerRef(obj, gvk)
+	// Setting Controller to false makes this a non-controlling owner reference.
+	// The owned object will be garbage collected when the ISVC is deleted,
+	// but the ISVC doesn't "control" its lifecycle in other controller-runtime senses.
+	or.Controller = ptr.To(false)
```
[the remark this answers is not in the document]
pierDipi: do you find that more readable? It's not necessary though

internal/controller/serving/inferenceservice_controller_test.go
```diff
@@ -1289,6 +1293,343 @@ var _ = Describe("InferenceService Controller", func() {
...
+		var (
+			testNs         string
+			kedaReconciler *reconcilers.KserveKEDAReconciler
+		)
+
+		BeforeEach(func() {
+			ctx = context.Background()
```
bartoszmajsak, line 1304:
```suggestion
```

No need to overwrite that global variables here :)

bartoszmajsak (approved):
Two small nits, but otherwise LGTM.

Thanks for considering my earlier feedback.

internal/controller/serving/inferenceservice_controller_test.go
pierDipi: Done

PR #459, opened 2025-05-21
[RHOAIENG-25522] Route that is not managed by the InferenceService gets deleted

spolti:
While the reported issue was not reproducible in the current code base, I've made an improvement in the serverless mode route deletion, it was deleting the route based on the isvc name, whereas it should be deleting based on the owner reference, similar to how raw route handling is done.

Plus, a unit test was added to verify that the route is still deployed upon Inference Service deletion.

## Merge criteria:


pierDipi (approved):
/lgtm
/approve

PR #465, opened 2025-05-28
Adding Timeout To Openshift Routes Created For Inference Services

brettmthompson:
## Description

[RHOAIENG-24705](https://issues.redhat.com/browse/RHOAIENG-24705)

1. Implement default behavior for openshift route timeout by adding the `haproxy.router.openshift.io/timeout` annotation to the route. Now when openshift routes are created for an inference service the timeout will be set to the sum of all existing component timeouts (predictor + transformer + explainer). If a component exists, but does not have a timeout value set, then the default openshift route timeout will be used as the timeout for that component.
2. Support overriding the default timeout behavior by setting the `haproxy.router.openshift.io/timeout` annotation directly in the ISVC.

**Note:**
In order to handle this change the route comparator methods must be updated to also check against the desired and existing route annotations. By doing this, any change to inference service annotations will cause the route created by the `KserveRouteReconciler` to be updated.

## How Has This Been Tested?

- Unit tests added
- To test manually, create an ISVC with the timeout value specified in the transformer or predictor component, then observe the route is created with the appropriate `haproxy.router.openshift.io/timeout` annotation. Alternatively, add the `haproxy.router.openshift.io/timeout` annotation directly to the ISVC and observe that the route is created with the same `haproxy.router.openshift.io/timeout` annotation.

## Merge criteria:


internal/controller/serving/reconcilers/kserve_route_reconciler.go
brettmthompson, line 1: Is it necessary to add all the annotations and labels included in an inference service to the route created in this file? By doing so, we run the risk of unnecessarily updating  this route when irrelevant annotations or labels are added to the inference service.

In the `KserveRawRouteReconciler` and `MMRouteReconciler` the annotations and labels are explicitly set in the route rather than copying everything from the inference service.

internal/controller/utils/utils.go
```diff
@@ -436,3 +438,28 @@ func SubstituteVariablesInQueries(data string, namespace string, name string) st
...
+		if route.Annotations == nil {
+			route.Annotations = make(map[string]string)
+		}
+		route.Annotations[constants.RouteTimeoutAnnotationKey] = strconv.Itoa(int(*isvc.Spec.Transformer.TimeoutSeconds))
+	} else if isvc.Spec.Predictor.TimeoutSeconds != nil {
+		if route.Annotations == nil {
+			route.Annotations = make(map[string]string)
+		}
+		route.Annotations[constants.RouteTimeoutAnnotationKey] = strconv.Itoa(int(*isvc.Spec.Predictor.TimeoutSeconds))
+	}
```
pierDipi, line 464: Tranformer timeout is always greater than predictor timeout and it "includes" the predictor processing latency?

pierDipi: This is also technically a breaking change because if I have a predictor timeout set and not transformer timeout, I would now have a top-level timeout = predictor timeout (I'm not saying we shouldn't do it but just so we're aware and release note it accordingly)

internal/controller/serving/reconcilers/kserve_route_reconciler.go
```diff
@@ -162,6 +162,10 @@ func (r *KserveRouteReconciler) createDesiredResource(ctx context.Context, isvc
 				WildcardPolicy: v1.WildcardPolicyNone,
 			},
 		}
+
+		// Set route timeout
+		utils2.SetOpenshiftRouteTimeoutForIsvc(route, isvc)
```
spolti, line 167: Why call it twice?

internal/controller/utils/utils.go
```diff
@@ -436,3 +438,28 @@ func SubstituteVariablesInQueries(data string, namespace string, name string) st
...
+	if _, ok := isvc.Annotations[constants.RouteTimeoutAnnotationKey]; ok {
+		if route.Annotations == nil {
+			route.Annotations = make(map[string]string)
+		}
+		route.Annotations[constants.RouteTimeoutAnnotationKey] = isvc.Annotations[constants.RouteTimeoutAnnotationKey]
+		return
+	}
+
+	// By default openshift route timeout will be set to the timeout value of the target service's respective component, if any.
+	if isvc.Spec.Transformer != nil && isvc.Spec.Transformer.TimeoutSeconds != nil {
```
spolti, line 454: Doesn't the explainer have a timeout as well?

internal/controller/serving/inferenceservice_controller_test.go
```diff
@@ -1107,6 +1171,45 @@ var _ = Describe("InferenceService Controller", func() {
...
+				if err := k8sClient.Create(ctx, inferenceService); err != nil && !k8sErrors.IsAlreadyExists(err) {
+					Expect(err).NotTo(HaveOccurred())
+				}
+
+				// Checking that the controller has created the Route with the haproxy.router.openshift.io/timeout annotation added
+				Eventually(func() error {
+					return checkRouteTimeout(types.NamespacedName{Name: inferenceService.Name, Namespace: inferenceService.Namespace}, "45")
+				}, timeout, interval).Should(Succeed())
+
+				By("Updating an existing inference service with the haproxy.router.openshift.io/timeout annotation")
```
spolti, line 1200: so the annotation has precedence over the spec.predictor.timeout?

brettmthompson: Yes the annotation can be set in the isvc to override the default behavior of using the timeout defined in the isvc components spec.

internal/controller/utils/utils.go
brettmthompson: I am aware of this. 

I based this implementation on what is currently done for RAW deployments [here](https://github.com/opendatahub-io/odh-model-controller/blob/incubating/internal/controller/serving/reconcilers/kserve_raw_route_reconciler.go#L113). The transformer service seems to take precedent over the predictor and the explainer is not even checked. So I tried to follow that flow elsewhere.

I am definitely open to changing this approach though. IMO, setting the timeout equivalent to all component timeouts combined makes sense.

internal/controller/serving/inferenceservice_controller_test.go
```diff
@@ -1107,6 +1171,45 @@ var _ = Describe("InferenceService Controller", func() {
 					return k8sClient.Get(ctx, key, route)
 				}, timeout, interval).ShouldNot(HaveOccurred())
 			})
+			It("Should create a route with timeout if isvc has the label to expose route", func() {
+				By("Creating an inference service with a timeout value defined in the component spec")
```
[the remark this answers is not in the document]
brettmthompson: Each of these tests validates a different route reconciler. `KserveRawRouteReconciler`, `KserveRouteReconciler`, and `ModelMeshRouteReconciler`

internal/controller/utils/utils.go
brettmthompson: @spolti @pierDipi Wdyt about updating the `SetOpenshiftRouteTimeoutForIsvc` method to set the route timeout to a combination of predictor + transformer + explainer timeout?

The question would then become should we include a default value in that summation if any of these components exist, but do not have a timeout set?

internal/controller/serving/inferenceservice_controller_test.go
```diff
@@ -138,6 +138,44 @@ var _ = Describe("InferenceService Controller", func() {
 				}, timeout, interval).Should(Succeed())
 			})
 
+			It("Should create a Route with timeout", func() {
+				By("Creating an inference service with a timeout value defined in the component spec")
```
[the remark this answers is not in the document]
brettmthompson: Added tests to cover inference services created with all 3 components with/without timeout explicitly set.

internal/controller/serving/inferenceservice_controller_test.go
```diff
@@ -1107,6 +1208,71 @@ var _ = Describe("InferenceService Controller", func() {
...
+					Expect(err).NotTo(HaveOccurred())
+				}
+
+				// Checking that the controller has created the Route with the haproxy.router.openshift.io/timeout annotation added
+				Eventually(func() error {
+					return checkRouteTimeout(types.NamespacedName{Name: inferenceService.Name, Namespace: inferenceService.Namespace}, "135s")
+				}, timeout, interval).Should(Succeed())
+
+				By("Updating an existing inference service with the haproxy.router.openshift.io/timeout annotation")
+				deployedInferenceService := &kservev1beta1.InferenceService{}
```
spolti, line 1238: Do you think that it is also possible to add an e2e test that forces the inference service to wait for the timeout to happen?

spolti (approved):
Left one major comment for a test addition.
If we don't have enough time for now, let's do it on another PR.

internal/controller/serving/inferenceservice_controller_test.go
brettmthompson: I have created an e2e test in opendatahub-test to cover this. This test creates an inference service and validates inference is performed successfully. Then I add the timeout annotation to the inference service with a very small value (1 microsecond) and validate that inference fails since the timeout is exceeded.

spolti: great :D
Have you sent the PR yet?
I am not seeing it here https://github.com/opendatahub-io/opendatahub-tests/pulls

brettmthompson: @spolti e2e test added to this PR: https://github.com/opendatahub-io/opendatahub-tests/pull/332

PR #487, opened 2025-07-23
Updating the secret controller to limit model mesh certificates

brettmthompson:
Given that `kserve` sets the data in the storage-config secret as an env variable in the `storage-initializer` container, if the size of the data in the `storage-config` secret is very large it is possible that the execution of the `initializer-entrypoint` script on container start up can throw the following error:
```
'message': 'exec container process `/storage-initializer/scripts/initializer-entrypoint`: Argument list too long\n',
```
This results from the total size of the environmental variables, combined with the command-line arguments exceeding the kernel's `ARG_MAX` limit.

## Description

After merging https://github.com/opendatahub-io/odh-model-controller/pull/486, we now set all global + custom + serving certificates in the `odh-kserve-custom-ca-bundle` configmap. This means that with the current logic all of those certificates will be added to the 'certificate' field of the `storage-config` secret for each data connection that utilizes `https`. To prevent this error from arising due to an excess of certificates, this PR only adds the custom certificates present under the `odh-ca-bundle.crt` field in the `odh-trusted-ca-bundle` configmap to the 'certificate' field of the `storage-config` secret. These certificates represent custom certificates manually added to the DSC by end users.

## How Has This Been Tested?

Unit tests and kserve fvt tests

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected handling of custom certificate data for storage secrets, ensuring both ODH and KServe certificates are processed and included appropriately.

* **Tests**
  * Updated test data to reflect new certificate handling logic and refreshed certificate values in test secrets.


mwaykole (approved):

shalberd:
@brettmthompson Yes, I can confirm global odh-trusted-ca-bundle certs are integrated fine with Kserve Raw in ODH 2.33.0. Also good to see KServe, as opposed to model-mesh, references CA trust variables by volume mount / path from configmap in CA-trust-related env vars, so the issue mentioned above with CA-trust env vars containing the CAs themselves is really only limited to model-mesh.
Personally, we will keep going with KServe Raw at our organization and have decided not to use model mesh at all. Due to product strategy / roadmap WatsonX / RHODS combo as well, which seems to go with KServe Raw.

PR #531, opened 2025-09-26
chore: pins go version using go.mod

bartoszmajsak:
Instead of using explicit version in github action we can rely on what is defined in `go.mod` file. This way we will have jobs up-to-date using single place to define toolchain version.

## Summary by CodeRabbit

* **Chores**
  * CI build pipeline now also runs for branches matching stable-*, ensuring release branch builds are automatically validated.
  * Lint and end-to-end test pipelines now derive the Go toolchain version from the project configuration, improving consistency and reducing manual version maintenance.
  * Overall CI reliability improved by aligning workflows to the declared project version and expanding branch coverage. No changes to application behavior or user-facing features.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -6,6 +6,7 @@ on:
     branches:
       - incubating
       - main
+      - stable-*
     tags:
       - 'latest'
       - 'odh-v*'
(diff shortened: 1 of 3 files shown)


hdefazio:
Nice! 
/lgtm

hdefazio (changes requested):
@bartoszmajsak Nice! there is also https://github.com/opendatahub-io/odh-model-controller/blob/incubating/.github/workflows/test.yml#L18 that directly sets the go version

PR #532, opened 2025-09-26
ci(manifests): validates kustomize manifest

bartoszmajsak:
To ensure that our kustomize manifests are always valid, this PR brings a validation script and gh action to ensure that.

> [!IMPORTANT]
> This only checks if kustomize can process YAMLs, it does not mean it
can be successfully applied to the cluster.

## Summary by CodeRabbit

* **Chores**
  * Introduced an automated validation workflow for configuration manifests that runs on pushes and pull requests.
  * Ensures required tooling is installed and validates all manifests consistently.
  * Provides clear, aggregated results with success/failure indicators and detailed error output for easier troubleshooting.
  * Supports excluding specified paths from validation when needed.
  * Fails the check if any manifest is invalid, improving reliability before merge.

### DIFF
--- a/.github/workflows/validate-manifests.yml
+++ b/.github/workflows/validate-manifests.yml
@@ -1,0 +1,29 @@
+name: Validation
+
+on:
+  pull_request:
+  push:
+
+jobs:
+  validate-manifests:
+    name: Validate Kustomize manifests
+    timeout-minutes: 10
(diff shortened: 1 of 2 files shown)


mwaykole (approved):

spolti (approved):
/lgtm

PR #533, opened 2025-09-26
test(utils): temporarily accept 'Standard' and 'Knative' deploymentMode

vivekk16:
Added temporary support in GetDeploymentModeForKServeResource:
- "Standard" → mapped to RawDeployment
- "Knative" → mapped to Serverless

This change is only to ensure KServe e2e RawDeployment tests pass. It is not intended as a permanent feature and should be removed or revised once upstream fixes are in place.

## Summary by CodeRabbit

* **New Features**
  * Added support for “Standard” and “Knative” values in deployment mode annotations/configuration. “Standard” maps to raw deployment and “Knative” maps to serverless, offering clearer, user-friendly aliases. Behavior for defaults and invalid values remains unchanged.

### DIFF
--- a/internal/controller/utils/utils.go
+++ b/internal/controller/utils/utils.go
@@ -56,6 +56,7 @@ func GetDeploymentModeForKServeResource(ctx context.Context, cli client.Client, 
 			return constants.Serverless, nil
 		case string(constants.RawDeployment):
 			return constants.RawDeployment, nil
+		// TODO: Remove temporary alias mapping ("Standard"→RawDeployment, "Knative"→Serverless) post-RHOAI 3.0, with code refactor for new deployment modes.
 		case "Standard":
 			return constants.RawDeployment, nil
 		case "Knative":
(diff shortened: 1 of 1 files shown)


internal/controller/utils/utils.go
```diff
@@ -56,6 +56,10 @@ func GetDeploymentModeForKServeResource(ctx context.Context, cli client.Client,
 			return constants.Serverless, nil
 		case string(constants.RawDeployment):
 			return constants.RawDeployment, nil
+		case "Standard":
```
VedantMahabaleshwarkar, line 59: If this is temporary can you create a follow up JIRA to remove this and add it here as a comment so that we don't lose track?

VedantMahabaleshwarkar (changes requested):

internal/controller/utils/utils.go
vivekk16: https://issues.redhat.com/browse/RHOAIENG-35081
@VedantMahabaleshwarkar  I've created this JIRA ticket to track the removal.

PR #561, opened 2025-10-16
feat: add BaseRefs support for LLMInferenceService gateway resolution

pierDipi:
This adds support for LLMInferenceService.Spec.BaseRefs, which allows referencing LLMInferenceServiceConfig objects to inherit gateway configurations. The implementation includes:

- Add getConfig() method to retrieve LLMInferenceServiceConfig from service namespace or system namespace (POD_NAMESPACE env var), prioritizing the service namespace
- Enhance getGatewayInfo() to merge specs from BaseRefs configs before applying the service's own spec
- Add nil check for config retrieval to prevent panics when configs are not found
- Always append service spec to ensure it takes precedence in the merge

## Summary by CodeRabbit

* **Chores**
  * Updated infrastructure dependencies to support enhanced gateway management capabilities.

* **Improvements**
  * Enhanced configuration resolution with improved error handling and fallback mechanisms for gateway discovery across namespaces.
  * Expanded test coverage for configuration-driven gateway management and AuthPolicy generation scenarios.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -164,6 +164,7 @@ require (
 	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.31.2 // indirect
 	sigs.k8s.io/external-dns v0.14.0 // indirect
 	sigs.k8s.io/json v0.0.0-20241014173422-cfa47c3a1cc8 // indirect
+	sigs.k8s.io/lws v0.6.2 // indirect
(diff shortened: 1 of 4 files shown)


Jooho (commented):
This isn’t a critical issue that would block merging.
I think the comment I left can be addressed later as a follow-up if needed.

PR #568, opened 2025-10-21
Delete network policies, rely on ODH operator ones

pierDipi:
I've noticed a cluster where the odh-model-controller network policy prevents the odh-model-controller webhook from responding, having NetworkPolicy working by default is important for a good UX.

```
[pierdipi] kserve -> $ k annotate llmisvc deepseek-r1-0528 x=x2 --overwrite
Error from server (InternalError): Internal error occurred: failed calling webhook "validating.llmisvc.odh-model-controller.opendatahub.io": failed to call webhook: Post "https://odh-model-controller-webhook-service.opendatahub.svc:443/validate-serving-kserve-io-v1alpha1-llminferenceservice?timeout=10s": dial tcp 10.128.4.55:9443: i/o timeout
```
```
[pierdipi] kserve -> $ k get ns -l=webhook=enabled
No resources found
```

## Summary by CodeRabbit

* **Chores**
  * Removed previously included network policies that allowed metrics and webhook ingress traffic.
  * Top-level configuration no longer references the network-policy resource group, simplifying network policy deployment.


config/network-policy/allow-metrics-traffic.yaml
```diff
@@ -1,26 +1,28 @@
...
+# Only Pods on those namespaces are able to gathering data from the metrics endpoint.
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
   labels:
     app.kubernetes.io/name: odh-model-controller
     app.kubernetes.io/managed-by: kustomize
-  name: allow-metrics-traffic
-  namespace: system
+  name: odh-model-controller-allow-metrics-traffic
```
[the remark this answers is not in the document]
israel-hdez: IIRC, this came from `kubebuilder init`. From Kubebuilder book, these are optional ([ref](https://book.kubebuilder.io/reference/metrics#optional-by-using-network-policy-disabled-by-default)).

config/network-policy/allow-webhook-traffic.yaml
```diff
@@ -1,26 +1,31 @@
...
       - namespaceSelector:
           matchLabels:
-            webhook: enabled # Only from namespaces with this label
+            webhook: enabled
+      - namespaceSelector:
+          matchLabels:
+            kubernetes.io/metadata.name: openshift-kube-apiserver
+      - namespaceSelector:
+          matchLabels:
+            kubernetes.io/metadata.name: openshift-apiserver
```
israel-hdez:
```suggestion
      - namespaceSelector:
          matchExpressions:
            - key: policy-group.network.openshift.io/host-network
              operator: Exists
```

config/network-policy/allow-metrics-traffic.yaml
bartoszmajsak: Do we need this network-policy at all?  How is it applied? I see it being commented out from kustomize chain:

https://github.com/opendatahub-io/odh-model-controller/blob/0127a4910c8c6bfa1845b97a9205278999b3205a/config/network-policy/kustomization.yaml#L3

Can we delete it to avoid confusion?

config/network-policy/allow-webhook-traffic.yaml
bartoszmajsak: This is a solution for a specific environment, I wonder if we need this in the "default settings" or should this be just documented? https://access.redhat.com/solutions/7008681


@israel-hdez where can I learn about "webhook: enabled" label? It's the first time I see this solution.

config/network-policy/allow-webhook-traffic.yaml
```diff
@@ -1,26 +1,31 @@
...
   ingress:
-    # This allows ingress traffic from any namespace with the label webhook: enabled
     - from:
       - namespaceSelector:
           matchLabels:
-            webhook: enabled # Only from namespaces with this label
+            webhook: enabled
+      - namespaceSelector:
+          matchLabels:
+            kubernetes.io/metadata.name: openshift-kube-apiserver
```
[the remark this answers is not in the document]
israel-hdez: IMO odh-model-controller is quite OpenShift-specific. So, I see little reason to worry about k8s.

PR #584, opened 2025-10-27
[s390x] Update vLLM spyre on Z ServingRuntime template

nikheal2:
Update vLLM spyre on Z ServingRuntime template

## Description

for IBM Z , only VF and continuous batching is supported. Updated serving-runtime accordingly. 

## How Has This Been Tested?
Locally Tested on Z cluster.

## Summary by CodeRabbit

* **Chores**
  * Updated vLLM deployment configuration including accelerator settings and runtime environment variables.
  * Enhanced shared memory allocation (2Gi) for improved resource management.
  * Refined container startup process and model path handling for better compatibility.

### DIFF
--- a/config/runtimes/vllm-spyre-s390x-template.yaml
+++ b/config/runtimes/vllm-spyre-s390x-template.yaml
@@ -22,7 +22,7 @@ objects:
       name: vllm-spyre-s390x-runtime
       annotations:
         openshift.io/display-name: vLLM Spyre s390x ServingRuntime for KServe
-        opendatahub.io/recommended-accelerators: '["ibm.com/spyre_pf"]'
+        opendatahub.io/recommended-accelerators: '["ibm.com/spyre_vf"]'
         opendatahub.io/runtime-version: 'v0.10.2.0'
       labels:
         opendatahub.io/dashboard: 'true'
@@ -34,37 +34,39 @@ objects:
         - image: $(vllm-spyre-s390x-image)
           name: kserve-container
           command:
-            - python3
-            - '-m'
-            - vllm_tgis_adapter
+            - /bin/bash
+            - -lc
+            - exec python3 -m vllm.entrypoints.openai.api_server "$@"
+            - --
           args:
-            - /mnt/models
+            - '--model=/mnt/models'
             - '--port=8000'
             - '--served-model-name={{.Name}}'
-            - '--grpc-port=8033'
           env:
             - name: HF_HOME
               value: /tmp/hf_home
             - name: FLEX_COMPUTE
               value: SENTIENT
             - name: FLEX_DEVICE
-              value: PF
+              value: VF
             - name: TOKENIZERS_PARALLELISM
               value: 'false'
             - name: DTLOG_LEVEL
               value: error
             - name: TORCH_SENDNN_LOG
               value: CRITICAL
-            - name: VLLM_SPYRE_WARMUP_BATCH_SIZES
-              value: '4'
-            - name: VLLM_SPYRE_WARMUP_PROMPT_LENS
-              value: '1024'
-            - name: VLLM_SPYRE_WARMUP_NEW_TOKENS
-              value: '256'
           ports:
             - containerPort: 8000
               protocol: TCP
+          volumeMounts:
+            - name: shm
+              mountPath: /dev/shm
       multiModel: false
       supportedModelFormats:
         - autoSelect: true
           name: vLLM
+      volumes:
+        - name: shm
+          emptyDir:
+            medium: Memory
+            sizeLimit: 2Gi
--- a/config/runtimes/vllm-spyre-s390x-template.yaml
+++ b/config/runtimes/vllm-spyre-s390x-template.yaml
@@ -35,8 +35,8 @@ objects:
           name: kserve-container
           command:
             - /bin/bash
-            - -lc
-            - exec python3 -m vllm.entrypoints.openai.api_server "$@"
+            - -c
+            - source /etc/profile.d/ibm-aiu-setup.sh && exec python3 -m vllm.entrypoints.openai.api_server "$@"
             - --
           args:
             - '--model=/mnt/models'
@@ -45,8 +45,6 @@ objects:
           env:
             - name: HF_HOME
               value: /tmp/hf_home
-            - name: FLEX_COMPUTE
-              value: SENTIENT
             - name: FLEX_DEVICE
               value: VF
             - name: TOKENIZERS_PARALLELISM
@@ -55,6 +53,12 @@ objects:
               value: error
             - name: TORCH_SENDNN_LOG
               value: CRITICAL
+            - name: VLLM_SPYRE_USE_CB
+              value: "1"
+            - name: VLLM_SPYRE_REQUIRE_PRECOMPILED_DECODERS
+              value: "1"
+            - name: TORCH_SENDNN_CACHE_ENABLE
+              value: "1"
           ports:
             - containerPort: 8000
               protocol: TCP


config/runtimes/vllm-spyre-s390x-template.yaml
```diff
@@ -34,37 +34,43 @@ objects:
         - image: $(vllm-spyre-s390x-image)
           name: kserve-container
           command:
-            - python3
-            - '-m'
-            - vllm_tgis_adapter
+            - /bin/bash
+            - -c
+            - source /etc/profile.d/ibm-aiu-setup.sh && exec python3 -m vllm.entrypoints.openai.api_server "$@"
```
edwardquarm: @nikheal2  do we need the first part of command since it's already implemented in dockerfile as startup command? https://github.com/vllm-project/vllm-spyre/blob/main/docker/simple_vllm_serve.sh#L4
``source /etc/profile.d/ibm-aiu-setup.sh``. It seems redundant because the ServingRuntime works fine without it  because it is implemented in entrypoint https://github.com/vllm-project/vllm-spyre/blob/main/docker/Dockerfile.amd64#L103

edwardquarm (changes requested):

spolti:
@edwardquarm is this still needed?

edwardquarm:
@spolti this PR is not needed

mwaykole (changes requested):
## PR Risk Assessment - Review by mwaykole

**PR**: https://github.com/opendatahub-io/odh-model-controller/pull/584
**File**: `config/runtimes/vllm-spyre-s390x-template.yaml`
**Overall Risk**: 🟡 **Medium → REQUEST_CHANGES**

The changes are directionally correct (PF→VF, tgis→OpenAI API server, continuous batching env vars, /dev/shm volume), but there are several functional correctness gaps that must be addressed before merge. See inline comments for details.

### Key Issues
1. **Wrong file path** — this PR modifies `config/runtimes/vllm-spyre-s390x-template.yaml` but on `incubating` that file no longer exists; it moved to `config/runtimes/vllm/vllm-spyre-s390x-template.yaml`. The file modified here is NOT in the kustomization manifest, so these changes have **zero effect** on deployed clusters.
2. **Missing `schedulerName: spyre-scheduler`** — without this, pods land on nodes without Spyre hardware and the runtime fails.
3. **Stale runtime version** — `v0.10.2.0` should be `v0.21.0` (matching the `incubating` canonical version).
4. **Missing image parameter declaration** — `$(vllm-spyre-s390x-image)` is referenced but there is no `parameters:` section in the template.
5. **Missing `opendatahub.io/kserve-runtime: vllm` annotation** — required for runtime identification by odh-model-controller.
6. **Env var quoting inconsistency** — new vars use double-quotes while existing ones use single-quotes.

config/runtimes/vllm-spyre-s390x-template.yaml
```diff
@@ -22,7 +22,7 @@ objects:
       name: vllm-spyre-s390x-runtime
```
mwaykole, line 22: **🚨 Critical: Wrong file path — changes have zero effect on deployed clusters**

This PR modifies `config/runtimes/vllm-spyre-s390x-template.yaml` at the root of `config/runtimes/`, but on the `incubating` base branch **this file no longer exists at that path**. The runtime templates were reorganized into a `vllm/` subdirectory. The authoritative path is:

`config/runtimes/vllm/vllm-spyre-s390x-template.yaml`

The `kustomization.yaml` on `incubating` only references `vllm/` as a resource, not the root-level file. **These changes will have zero effect on deployed clusters.**

Please:
1. Rebase the branch onto `incubating`
2. Apply these changes to `config/runtimes/vllm/vllm-spyre-s390x-template.yaml`
3. Coordinate with your own PR #736 which also targets the same file on `incubating`

config/runtimes/vllm-spyre-s390x-template.yaml
```diff
@@ -34,37 +34,43 @@ objects:
         - image: $(vllm-spyre-s390x-image)
           name: kserve-container
```
mwaykole, line 35: **Stale runtime version** — `v0.10.2.0` was in the original file and has not been bumped. The authoritative `incubating` branch version of this template is already at `v0.21.0`. This annotation is directly surfaced in the RHOAI Dashboard's runtime selector — users will see the wrong version string, making support and debugging harder.

```yaml
# Change to:
opendatahub.io/runtime-version: 'v0.21.0'
```

config/runtimes/vllm-spyre-s390x-template.yaml
```diff
@@ -34,37 +34,43 @@ objects:
...
               protocol: TCP
+          volumeMounts:
+            - name: shm
+              mountPath: /dev/shm
       multiModel: false
       supportedModelFormats:
         - autoSelect: true
-          name: vLLM
\ No newline at end of file
+          name: vLLM
```
mwaykole, line 71: **🚨 Missing `schedulerName: spyre-scheduler`** — this is a critical functional gap.

The canonical `incubating` version of this template includes `schedulerName: spyre-scheduler` between `multiModel: false` and `supportedModelFormats`. Without it, Kubernetes uses the **default scheduler**, which will place pods on nodes that may not have Spyre hardware, causing the runtime to fail at startup.

```yaml
      multiModel: false
      schedulerName: spyre-scheduler   # <-- add this
      supportedModelFormats:
```

See `config/runtimes/vllm/vllm-spyre-s390x-template.yaml` on `incubating` for confirmation.

config/runtimes/vllm-spyre-s390x-template.yaml
```diff
@@ -34,37 +34,43 @@ objects:
         - image: $(vllm-spyre-s390x-image)
           name: kserve-container
           command:
-            - python3
-            - '-m'
-            - vllm_tgis_adapter
```
mwaykole, line 39: **Image parameter not declared — template will fail to resolve**

This template uses `$(vllm-spyre-s390x-image)` as the container image reference, but the file has **no `parameters:` section** at the end of the template (as required by OpenShift Template spec). At deploy time, the `$(vllm-spyre-s390x-image)` variable will not be substituted and the pod spec will have a blank image reference.

Options:
1. **Add a `parameters:` section** at the end of this file:
```yaml
parameters:
  - name: vllm-spyre-s390x-image
    value: 'quay.io/modh/vllm-spyre:latest'  # use the correct default image
```
2. **Align with the incubating approach** and switch to `$(vllm-spyre-image)` which is the shared parameter already declared in the canonical template.

Cross-reference: the `incubating` version uses `$(vllm-spyre-image)` (not architecture-specific). If there's a deliberate reason to maintain a separate s390x image stream, please document it and add the parameter declaration.

config/runtimes/vllm-spyre-s390x-template.yaml
```diff
@@ -34,37 +34,43 @@ objects:
         - image: $(vllm-spyre-s390x-image)
           name: kserve-container
           command:
-            - python3
-            - '-m'
```
mwaykole, line 38: **Missing `opendatahub.io/kserve-runtime: 'vllm'` spec annotation**

The canonical `incubating` version of this template includes a `spec.annotations` block with `opendatahub.io/kserve-runtime: 'vllm'`. This annotation is used by the odh-model-controller to identify the runtime type for reconciliation, metrics routing, and dashboard integration. Without it, the controller may not correctly classify this runtime.

```yaml
    spec:
      annotations:
        opendatahub.io/kserve-runtime: 'vllm'   # <-- add this
        prometheus.io/path: /metrics
        prometheus.io/port: '8080'
```

(The `prometheus.io` annotations are already present in the base file — only `opendatahub.io/kserve-runtime` is missing.)

config/runtimes/vllm-spyre-s390x-template.yaml
```diff
@@ -34,37 +34,43 @@ objects:
...
-            - vllm_tgis_adapter
+            - /bin/bash
+            - -c
+            - source /etc/profile.d/ibm-aiu-setup.sh && exec python3 -m vllm.entrypoints.openai.api_server "$@"
+            - --
           args:
-            - /mnt/models
+            - '--model=/mnt/models'
             - '--port=8000'
```
mwaykole, line 43: **Entrypoint: source failure will silently skip setup**

The pattern `source /etc/profile.d/ibm-aiu-setup.sh && exec python3 ...` is correct in intent, but if `ibm-aiu-setup.sh` doesn't exist in the container image (e.g., after a future image rebuild or if the wrong image tag is pulled), the `source` command will fail and `exec` will **never run** — the container will exit with a non-zero code and surface a cryptic error.

Consider making the source optional:
```yaml
- '[ -f /etc/profile.d/ibm-aiu-setup.sh ] && source /etc/profile.d/ibm-aiu-setup.sh; exec python3 -m vllm.entrypoints.openai.api_server "$@"'
```

This is resilient to missing setup scripts while still applying them when present. Also, the `--` as the 4th element of the `command` array correctly separates bash args from the forwarded `args:` list — that part is fine.

config/runtimes/vllm-spyre-s390x-template.yaml
```diff
@@ -34,37 +34,43 @@ objects:
...
-              value: '1024'
-            - name: VLLM_SPYRE_WARMUP_NEW_TOKENS
-              value: '256'
+            - name: VLLM_SPYRE_USE_CB
+              value: "1"
+            - name: VLLM_SPYRE_REQUIRE_PRECOMPILED_DECODERS
+              value: "1"
+            - name: TORCH_SENDNN_CACHE_ENABLE
+              value: "1"
           ports:
```
mwaykole, line 62: **Minor: Env var value quoting inconsistency**

The three new env vars use bare double-quoted YAML strings (`"1"`):
```yaml
value: "1"
```

But all pre-existing env vars in this file use single-quoted YAML strings (`'false'`, `'error'`, `'CRITICAL'`). Both are valid YAML, but mixing styles in a single file reduces readability and can trip up YAML linters.

Please standardize to match the rest of the file:
```yaml
- name: VLLM_SPYRE_USE_CB
  value: '1'
- name: VLLM_SPYRE_REQUIRE_PRECOMPILED_DECODERS
  value: '1'
- name: TORCH_SENDNN_CACHE_ENABLE
  value: '1'
```

PR #595, opened 2025-11-10
resolve race condition in raw crb when 2 isvcs using the same SA have…

VedantMahabaleshwarkar:
… different auth requirements

## Description

Fixes : https://issues.redhat.com/browse/RHOAIENG-35816 

Added logic to prevent the existing race condition 

## How Has This Been Tested?

Added unit tests 

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ClusterRoleBinding handling so bindings are created only when needed, preserved when multiple InferenceServices share a service account with differing auth requirements, and cleaned up safely.

* **Tests**
  * Added coverage for shared service-account scenarios validating CRB stability and updated tests to centralize the service-account name used across cases.

### DIFF
--- a/internal/controller/serving/reconcilers/kserve_raw_clusterrolebinding_reconciler.go
+++ b/internal/controller/serving/reconcilers/kserve_raw_clusterrolebinding_reconciler.go
@@ -50,7 +50,10 @@ func NewKserveRawClusterRoleBindingReconciler(client client.Client) *KserveRawCl
 func (r *KserveRawClusterRoleBindingReconciler) Reconcile(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService) error {
 	log.V(1).Info("Reconciling ClusterRoleBinding for InferenceService")
 	// Create Desired resource
-	desiredResource := r.createDesiredResource(isvc)
+	desiredResource, err := r.createDesiredResource(ctx, log, isvc)
+	if err != nil {
+		return err
+	}
 
 	// Get Existing resource
 	existingResource, err := r.getExistingResource(ctx, log, isvc)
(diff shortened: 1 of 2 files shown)


internal/controller/serving/reconcilers/kserve_raw_clusterrolebinding_reconciler.go
```diff
@@ -138,3 +138,31 @@ func (r *KserveRawClusterRoleBindingReconciler) Cleanup(_ context.Context, _ log
 	// NO OP
 	return nil
 }
+
+func getExistingIsvcsWithTheSameSA(ctx context.Context, cli client.Client, log logr.Logger, isvcSA string, namespace string) ([]kservev1beta1.InferenceService, error) {
+	var isvcList kservev1beta1.InferenceServiceList
+	var listOpts client.ListOptions
+	listOpts = client.ListOptions{Namespace: namespace}
```
brettmthompson: nit: the `:=` syntax can be used to declare an initialize `listOpts`

internal/controller/serving/reconcilers/kserve_raw_clusterrolebinding_reconciler.go
```diff
@@ -138,3 +138,31 @@ func (r *KserveRawClusterRoleBindingReconciler) Cleanup(_ context.Context, _ log
...
+	// List all inference services in the namespace
+	if err := cli.List(ctx, &isvcList, &listOpts); err != nil {
+		log.Error(err, "failed to list inference services")
+		return nil, err
+	}
+	var existingIsvcs []kservev1beta1.InferenceService
+	for _, svc := range isvcList.Items {
```
brettmthompson: should we also be checking if the isvc has auth enabled here?

internal/controller/serving/reconcilers/kserve_raw_clusterrolebinding_reconciler.go
```diff
@@ -107,18 +85,28 @@ func (r *KserveRawClusterRoleBindingReconciler) Delete(ctx context.Context, log
...
+		existingIsvcs, err := getExistingIsvcsWithTheSameSAandAuthEnabled(ctx, r.client, log, isvcSA, isvc.Namespace)
+		if err != nil {
+			return nil, err
+		}
+		if len(existingIsvcs) == 0 {
+			// This ISVC does not request auth and no other ISVC using the same SA requests auth, so no CRB is needed
+			return nil, nil
+		}
+		// if execution has reached here, it means at least one other ISVC using the same SA requests auth
+		// thus, CRB is needed
```
brettmthompson: Is this check required? The cluster role binding should already be created by the crb reconciliation triggered by the existing inference services with auth enabled.

I think the original state of this function makes sense. If auth is disabled for the isvc being reconciled then the cluster role binding does not need to be created for its service account. If any inference services exist with auth enabled, that cluster role binding will be created while those inference services are reconciled if it does not already exist.

internal/controller/serving/reconcilers/kserve_raw_clusterrolebinding_reconciler.go
```diff
@@ -138,3 +126,32 @@ func (r *KserveRawClusterRoleBindingReconciler) Cleanup(_ context.Context, _ log
...
+func getExistingIsvcsWithTheSameSAandAuthEnabled(ctx context.Context, cli client.Client, log logr.Logger, isvcSA string, namespace string) ([]kservev1beta1.InferenceService, error) {
+	var isvcList kservev1beta1.InferenceServiceList
+	listOpts := client.ListOptions{Namespace: namespace}
+	// List all inference services in the namespace
+	if err := cli.List(ctx, &isvcList, &listOpts); err != nil {
+		log.Error(err, "failed to list inference services")
+		return nil, err
+	}
+	var existingIsvcs []kservev1beta1.InferenceService
+	for _, svc := range isvcList.Items {
```
brettmthompson: If we changed this for loop to something like the following to lower complexity, would this cover the same scenarios?
```
for _, isvc := range isvcList.Items {
    if isvc.GetDeletionTimestamp() != nil {
        continue
    }
    if val, ok := isvc.Annotations[constants.EnableAuthODHAnnotation]; ok && val == "true" {
        if isvcSA == constants.KserveServiceAccountName && len(isvc.Spec.Predictor.SaName) == 0 {
            existingIsvcs = append(existingIsvcs, svc)
        } else if isvcSA == isvc.Spec.Predictor.ServiceAccountName {
            existingIsvcs = append(existingIsvcs, svc)
        }
    }
}
```

brettmthompson (approved):

PR #616, opened 2025-12-03
Update the namespace for the MaaS Config to be configurable

jland-redhat:
Setting the TENANT_NAMESPACE variable on the ODH Controller now allows the automation to work on a namespace other than maas-api.

**Note:** There is more to be done around the actual operator integration which is why I did not include documentation yet since this should really only be used internally for now. But that will be included in a future PR

## Summary by CodeRabbit

* **New Features**
  * MaaS namespace is now configurable at runtime via the MAAS_NAMESPACE environment variable; when unset it falls back to the previous default. Tier configuration lookups and loading now respect the configured namespace.

* **Chores**
  * Deployment configs updated to expose MAAS_NAMESPACE and include parameter mapping for easy customization.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/config/base/kustomization.yaml
+++ b/config/base/kustomization.yaml
@@ -159,6 +159,17 @@ replacements:
       kind: ConfigMap
       version: v1
       name: odh-model-controller-parameters
+      fieldPath: data.maas-namespace
+    targets:
+      - select:
+          kind: Deployment
+          name: odh-model-controller
+        fieldPaths:
+          - spec.template.spec.containers.[name=manager].env.[name=MAAS_NAMESPACE].value
+  - source:
+      kind: ConfigMap
+      version: v1
+      name: odh-model-controller-parameters
       fieldPath: metadata.namespace
     targets:
       - select:
--- a/config/base/params.env
+++ b/config/base/params.env
@@ -3,6 +3,7 @@ ray-tls-generator-image=registry.redhat.io/ubi9/ubi-minimal:latest
 nim-state=managed
 kserve-state=managed
 modelregistry-state=removed
+maas-namespace=maas-api
 tgis-image=quay.io/opendatahub/text-generation-inference:fast
 ovms-image=quay.io/opendatahub/openvino_model_server:2025.1-release
 vllm-cuda-image=quay.io/opendatahub/vllm:fast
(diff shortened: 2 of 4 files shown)


nerdalert (approved):
LGTM tested and work as described.

israel-hdez (commented):
@jland-redhat we are missing a change in kustomize manifests to be able to configure this from odh-operator.

This should require the [update in the deployment](https://github.com/opendatahub-io/odh-model-controller/blob/93a3c2b4111e7c2807c7be28681f5d5d3a5d0568/config/manager/manager.yaml#L74), and adding a [new parameter](https://github.com/opendatahub-io/odh-model-controller/blob/93a3c2b4111e7c2807c7be28681f5d5d3a5d0568/config/base/params.env) and [binding it](https://github.com/opendatahub-io/odh-model-controller/blob/93a3c2b4111e7c2807c7be28681f5d5d3a5d0568/config/base/kustomization.yaml#L14) properly. Otherwise, this won't be settable by odh-operator.

Nit probably: the naming of `TENANT_NAMESPACE` would make it hard to relate it to MaaS.

jland-redhat:
> @jland-redhat we are missing a change in kustomize manifests to be able to configure this from odh-operator.
> 
> This should require the [update in the deployment](https://github.com/opendatahub-io/odh-model-controller/blob/93a3c2b4111e7c2807c7be28681f5d5d3a5d0568/config/manager/manager.yaml#L74), and adding a [new parameter](https://github.com/opendatahub-io/odh-model-controller/blob/93a3c2b4111e7c2807c7be28681f5d5d3a5d0568/config/base/params.env) and [binding it](https://github.com/opendatahub-io/odh-model-controller/blob/93a3c2b4111e7c2807c7be28681f5d5d3a5d0568/config/base/kustomization.yaml#L14) properly. Otherwise, this won't be settable by odh-operator.
> 
> Nit probably: the naming of `TENANT_NAMESPACE` would make it hard to relate it to MaaS.

Sure any thoughts about what we should name it? MAAS_TENANT_NAMESPACE maybe?

israel-hdez (approved):
/lgtm

PR #680, opened 2026-02-05
fix(auth): decouples EnvoyFilter from LLMISvc lifecycle

bartoszmajsak:
Previously, the EnvoyFilter for TLS to Authorino and the Gateway AuthPolicy were only created when the first LLMInferenceService was deployed. This caused 500 errors on endpoints like `/v1/models` that should work even without models, because the TLS configuration for Authorino was missing.

Introduces a dedicated Gateway controller that creates these resources when a managed Gateway exists, independent of LLMInferenceService presence. The Gateway owns both resources.

Also simplifies AuthPolicy template selection to use explicit AuthType instead of inferring from target Kind, and skips platform gateways owned by GatewayConfig.

Ref: https://issues.redhat.com/browse/RHOAIENG-48760

## Summary by CodeRabbit

* **New Features**
  * Adds a Gateway controller that ensures AuthPolicy and EnvoyFilter resources for Gateways are created, updated, restored, or removed as needed.

* **Refactor**
  * Unified AuthPolicy naming and templating to support multiple target kinds (Gateway/HTTPRoute).
  * EnvoyFilter creation now respects bootstrap annotation, platform-owned gateways, and explicit managed/unmanaged status.

* **Tests**
  * New Gateway controller test suite and updated policy/fixture tests to reflect unified naming and loader behavior.


internal/controller/serving/llm/gateway_controller.go
```diff
@@ -0,0 +1,356 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if !gateway.GetDeletionTimestamp().IsZero() {
+		logger.V(1).Info("Gateway being deleted, skipping reconciliation")
+		return ctrl.Result{}, nil
+	}
+
+	shouldCreateEnvoyFilter := utils.ShouldCreateEnvoyFilterForGateway(gateway)
+	shouldCreateAuthPolicy := !utils.IsExplicitlyUnmanaged(gateway) && !utils.IsOwnedByPlatformController(gateway)
```
pierDipi: should we also exclude the ones that don't belong to llmisvc?

internal/controller/serving/llm/gateway_controller.go
```diff
@@ -0,0 +1,356 @@
...
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				if _, ok := e.Object.(*gatewayapiv1.Gateway); ok {
+					return false
+				}
+				return utils.IsManagedByOpenDataHub(e.Object)
+			},
+		}).
+		Named("gateway-auth-bootstrap").
+		Complete(r)
```
pierDipi: we might need to watch llmisvc for when a GW ref is added/removed (given the other question above)

bartoszmajsak: We could, but given my answer above about "chicken-egg" problem not sure that will help?

internal/controller/serving/llm/gateway_controller.go
bartoszmajsak: Just to capture what we discussed offline for posterity.

In order to limit blast radius and to not mess up with other gateways in the cluster that might not be part of ODH, the logic should be tighten to:

- reconcile only known gateways (default one from cfgmap or explicitly annotated)
- watch llmisvc for gw-ref changes and reconcile those gateways too

internal/controller/serving/llm/gateway_controller.go
```diff
@@ -0,0 +1,683 @@
...
+		authPolicyStore:   resources.NewClientAuthPolicyStore(client),
+		deltaProcessor:    processors.NewDeltaProcessor(),
+	}
+}
+
+// +kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways,verbs=get;list;watch
+// +kubebuilder:rbac:groups=networking.istio.io,resources=envoyfilters,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=kuadrant.io,resources=authpolicies,verbs=get;list;watch;create;update;patch;delete
+
+func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
pierDipi: when we return an error should emit an even on the Gateway so that it's visible what went wrong?

pierDipi (approved):
/lgtm

PR #686, opened 2026-02-09
chore: update regex for release tag validation

israel-hdez:
Because of the new pattern for early access releases.

## Summary by CodeRabbit

* **Chores**
  * Updated release workflow and image tag matching to accept an optional Early Access suffix (e.g., -EA1) on release and image tags across all related components. Validation and replacement now recognize EA-designated tags while preserving existing behavior and control flow.


spolti (commented):
/lgtm

PR #715, opened 2026-03-05
Namespace rbac

spolti:
## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Added namespace-scoped role binding management for the controller, enabling dynamic RBAC configuration per namespace.

* **Bug Fixes / Security**
  * Reduced and refined permissions across controllers to follow least-privilege principle, limiting write access to secrets, configmaps, and other resources.

* **Chores**
  * Updated Kubernetes RBAC configuration resources to support enhanced namespace-scoped access control.

### DIFF
--- a/internal/controller/serving/reconcilers/namespace_rbac_reconciler.go
+++ b/internal/controller/serving/reconcilers/namespace_rbac_reconciler.go
@@ -1,0 +1,217 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reconcilers
+
+import (
+	"context"
+	"os"
+
+	"github.com/go-logr/logr"
+	kservev1alpha1 "github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
+	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	NamespaceRoleBindingName = "odh-model-controller-namespace"
+	NamespaceClusterRoleName = "odh-model-controller-namespace"
+)
+
+// NamespaceRBACReconciler manages dynamic RoleBindings that grant the controller
+// write permissions in namespaces where InferenceServices or LLMInferenceServices exist.
+type NamespaceRBACReconciler struct {
+	client client.Client
+}
+
+func NewNamespaceRBACReconciler(client client.Client) *NamespaceRBACReconciler {
+	return &NamespaceRBACReconciler{
+		client: client,
+	}
+}
+
+// ReconcileNamespaceRBAC ensures a RoleBinding exists in the given namespace,
+// binding the controller's ServiceAccount to the odh-model-controller-namespace ClusterRole.
+func (r *NamespaceRBACReconciler) ReconcileNamespaceRBAC(ctx context.Context, log logr.Logger, namespace string) error {
+	controllerNamespace := os.Getenv("POD_NAMESPACE")
+	if controllerNamespace == "" {
+		controllerNamespace = "opendatahub"
+	}
+
+	// Skip if this is the controller's own namespace — it has a static Role/RoleBinding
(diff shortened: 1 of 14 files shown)


mwaykole (changes requested):
## PR Risk Assessment

**PR**: https://github.com/opendatahub-io/odh-model-controller/pull/715
**Title**: Namespace RBAC
**Author**: spolti
**Overall Risk**: 🔴 HIGH — Logic bug causes silent RoleBinding recreation failure; RBAC reduction needs careful verification against active write paths; security nuances in namespace-scoped role/rolebinding grants.

### Summary

This PR implements a solid architectural improvement: moving write permissions from a broad ClusterRole to dynamic, namespace-scoped RoleBindings created on-demand. This follows least-privilege principles correctly. However there are several concrete issues found during review:

1. **Logic Bug** — `ReconcileNamespaceRBAC` has a broken "fall through to create" path: after deleting a RoleBinding with a wrong RoleRef, the function exits without creating the replacement (returns `nil` silently).
2. **RBAC Reduction Risk** — `secret_controller.go` actively creates/updates/deletes secrets in user namespaces, but its kubebuilder annotation was reduced to `get;list;watch`. The dynamic RoleBinding covers this only if established first — and `SecretReconciler` doesn't call `ReconcileNamespaceRBAC` at all.
3. **NIM Template Regression** — `account_controller.go` uses `TemplateClient.Apply()` (which still uses the SA token) to write templates. The annotation was reduced to `get;list;watch` and `controller_namespace_rbac.yaml` doesn't include `template.openshift.io` write permissions.
4. **Two Separate Reconciler Instances** — `cmd/main.go` creates two independent `NamespaceRBACReconciler` instances (one for ISVC, one for LLM-ISVC). Both attempt to manage the same-named RoleBinding, creating a TOCTOU race on the delete+create path.
5. **Security** — `namespace_role.yaml` grants `create/delete/patch/update` on `rolebindings` and `roles` in user namespaces. This may be intentional but should be explicitly documented as accepted risk.
6. **Missing newline** — All 4 new YAML files are missing a trailing newline.
7. **No dedicated unit tests** for `NamespaceRBACReconciler` — the new reconciler has no `_test.go` file.

internal/controller/serving/reconcilers/namespace_rbac_reconciler.go
```diff
@@ -0,0 +1,217 @@
...
+				return err
+			}
+			// Fall through to create below
+		} else if !subjectsMatch(existing.Subjects, expectedSubjects) {
+			existing.Subjects = expectedSubjects
+			needsUpdate = true
+		}
+
+		if needsUpdate {
+			log.Info("Updating RoleBinding subjects", "namespace", namespace)
```
mwaykole, line 101: **🐛 Bug: "Fall through to create" doesn't actually fall through.**

After calling `r.client.Delete(ctx, existing)` on the wrong-RoleRef RoleBinding (line ~98), the code sets no flag and exits the `if err == nil` block. The outer guard below is `if !errors.IsNotFound(err)` where `err` is still `nil` from the original `Get` — so `errors.IsNotFound(nil)` returns `false`, `!false` is `true`, and the function **returns `nil` without ever creating the new RoleBinding**.

The "Fall through to create below" comment is misleading — no fall-through actually occurs.

**Suggested fix**: reset `err` after the delete so the create path is reached:
```go
if existing.RoleRef != expectedRoleRef {
    log.Info("Existing RoleBinding has wrong RoleRef, recreating", "namespace", namespace)
    if delErr := r.client.Delete(ctx, existing); delErr != nil && !errors.IsNotFound(delErr) {
        return delErr
    }
    // Reset err to NotFound so the create block below is reached
    err = apierrors.NewNotFound(rbacv1.Resource("rolebindings"), NamespaceRoleBindingName)
}
```

internal/controller/serving/reconcilers/namespace_rbac_reconciler.go
```diff
@@ -0,0 +1,217 @@
...
+// binding the controller's ServiceAccount to the odh-model-controller-namespace ClusterRole.
+func (r *NamespaceRBACReconciler) ReconcileNamespaceRBAC(ctx context.Context, log logr.Logger, namespace string) error {
+	controllerNamespace := os.Getenv("POD_NAMESPACE")
+	if controllerNamespace == "" {
+		controllerNamespace = "opendatahub"
+	}
+
+	// Skip if this is the controller's own namespace — it has a static Role/RoleBinding
+	if namespace == controllerNamespace {
+		return nil
```
mwaykole, line 60: **⚡ Performance + Correctness: `os.Getenv("POD_NAMESPACE")` called on every reconcile.**

This is called in both `ReconcileNamespaceRBAC` and `CleanupNamespaceRBAC` on every invocation. Issues:

1. **Performance**: `os.Getenv` is a syscall on every reconcile loop iteration — unnecessary overhead. Cache it once at construction time.
2. **Correctness**: The hardcoded fallback `"opendatahub"` will silently use the wrong namespace for deployments using non-standard namespaces (e.g., `redhat-ods-applications`), causing the controller-namespace skip check to fail.

**Recommended fix:**
```go
type NamespaceRBACReconciler struct {
    client              client.Client
    controllerNamespace string
}

func NewNamespaceRBACReconciler(client client.Client) *NamespaceRBACReconciler {
    ns := os.Getenv("POD_NAMESPACE")
    if ns == "" {
        ns = "opendatahub"
    }
    return &NamespaceRBACReconciler{client: client, controllerNamespace: ns}
}
```

PR #716, opened 2026-03-05
Revert "Reduce RBAC permissions scope for odh-model-controller"

spolti:
Reverts opendatahub-io/odh-model-controller#710

## Summary by CodeRabbit

* **Chores**
  * Consolidated role-based access control rules to streamline permission management across the system.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -10,6 +10,7 @@ rules:
   resources:
   - configmaps
   - secrets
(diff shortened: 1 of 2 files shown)


pierDipi:
Given it's runnings on the broken manifests, we need to

/override e2e-odh-kserve

pierDipi:
properly though :)

/override ci/prow/e2e-odh-kserve

PR #730, opened 2026-03-12
chore: update golangci-lint to v2

bartoszmajsak:
## Description

Upgrades golangci-lint from v1.64.8 to v2 and bumps Go to 1.25.

The v2 migration brings a cleaner config format - formatters (gofmt, goimports) are now a separate section, linter settings are properly nested, and deprecated options are replaced. The GH Actions workflow is updated to use the v7 action which supports v2.

All existing staticcheck quick-fix suggestions are resolved - mainly simplifying embedded field access patterns (e.g. `.ObjectMeta.Labels` -> `.Labels`, `.Client.Get` -> `.Get`) and cleaning up boolean expressions. Stale `golint` nolint directives are removed since that linter no longer exists in v2.

## How Has This Been Tested?

- `make lint` and `make lint-fix` pass with 0 issues
- GH Actions workflow aligned with v2

## Merge criteria:

## Release Notes

* **Chores**
  * Updated Go toolchain version specifications.
  * Refined linter configuration for improved code analysis.

* **Refactor**
  * Simplified internal resource comparison logic.
  * Updated client API calls for consistency across controllers.
  * Enhanced test utilities and fixture builders for API alignment.

* **Tests**
  * Improved test infrastructure and cleanup procedures.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,8 +1,6 @@
 version: "2"
-
 run:
   allow-parallel-runners: true
-
 linters:
   default: none
   enable:
@@ -45,11 +43,14 @@ linters:
           - dupl
           - lll
         path: internal/*
+      - linters:
+          - staticcheck
+        text: "ST1001"
+        path: "(test/|_test\\.go|internal/controller/testing/)"
     paths:
       - third_party$
       - builtin$
       - examples$
-
 formatters:
   enable:
     - gofmt
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/opendatahub-io/odh-model-controller
 
-go 1.25.7
+go 1.25
+
+toolchain go1.25.8
 
 require (
 	github.com/fsnotify/fsnotify v1.9.0
--- a/internal/controller/processors/deltaProcessor.go
+++ b/internal/controller/processors/deltaProcessor.go
@@ -38,7 +38,7 @@ func (d *deltaProcessor) ComputeDelta(comparator comparators.ResourceComparator,
 	var updated bool
 	var removed bool
 
-	if !(utils.IsNil(desiredResource) && utils.IsNil(existingResource)) {
+	if !utils.IsNil(desiredResource) || !utils.IsNil(existingResource) {
 		if utils.IsNotNil(desiredResource) && utils.IsNil(existingResource) {
 			added = true
 		} else if utils.IsNil(desiredResource) && utils.IsNotNil(existingResource) {
--- a/internal/controller/serving/inferencegraph_controller.go
+++ b/internal/controller/serving/inferencegraph_controller.go
@@ -57,7 +57,7 @@ func (r *InferenceGraphReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	logger := log.FromContext(ctx).WithValues("InferenceGraph", req.Name, "namespace", req.Namespace)
 
 	ig := &servingv1alpha1.InferenceGraph{}
-	err := r.Client.Get(ctx, req.NamespacedName, ig)
+	err := r.Get(ctx, req.NamespacedName, ig)
 	if err != nil && apierrs.IsNotFound(err) {
 		if apierrs.IsNotFound(err) {
 			return ctrl.Result{}, nil
--- a/internal/controller/serving/inferenceservice_controller.go
+++ b/internal/controller/serving/inferenceservice_controller.go
@@ -107,7 +107,7 @@ func (r *InferenceServiceReconciler) ReconcileServing(ctx context.Context, req c
 	// Get the InferenceService object when a reconciliation event is triggered (create,
 	// update, delete)
 	isvc := &kservev1beta1.InferenceService{}
-	err := r.Client.Get(ctx, req.NamespacedName, isvc)
+	err := r.Get(ctx, req.NamespacedName, isvc)
 	if err != nil && apierrs.IsNotFound(err) {
 		logger.Info("Stop InferenceService reconciliation")
 		// InferenceService not found, so we check for any other inference services that might be using Kserve
@@ -129,7 +129,7 @@ func (r *InferenceServiceReconciler) ReconcileServing(ctx context.Context, req c
 		logger.Info("Adding Finalizer")
 		if !controllerutil.ContainsFinalizer(isvc, constants.InferenceServiceODHFinalizerName) {
 			controllerutil.AddFinalizer(isvc, constants.InferenceServiceODHFinalizerName)
-			patchYaml := "metadata:\n  finalizers: [" + strings.Join(isvc.ObjectMeta.Finalizers, ",") + "]"
+			patchYaml := "metadata:\n  finalizers: [" + strings.Join(isvc.Finalizers, ",") + "]"
 			patchJson, _ := yaml.YAMLToJSON([]byte(patchYaml))
 			if err := r.Patch(ctx, isvc, client.RawPatch(types.MergePatchType, patchJson)); err != nil {
 				return reconcile.Result{}, err
@@ -149,7 +149,7 @@ func (r *InferenceServiceReconciler) ReconcileServing(ctx context.Context, req c
 				deleteErrors = multierror.Append(deleteErrors, err1)
 			}
 			controllerutil.RemoveFinalizer(isvc, constants.InferenceServiceODHFinalizerName)
-			patchYaml := "metadata:\n  finalizers: [" + strings.Join(isvc.ObjectMeta.Finalizers, ",") + "]"
+			patchYaml := "metadata:\n  finalizers: [" + strings.Join(isvc.Finalizers, ",") + "]"
 			patchJson, _ := yaml.YAMLToJSON([]byte(patchYaml))
 			if err := r.Patch(ctx, isvc, client.RawPatch(types.MergePatchType, patchJson)); err != nil {
 				return reconcile.Result{}, err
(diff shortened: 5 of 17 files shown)


internal/controller/serving/inferencegraph_controller.go
```diff
@@ -57,7 +57,7 @@ func (r *InferenceGraphReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	logger := log.FromContext(ctx).WithValues("InferenceGraph", req.Name, "namespace", req.Namespace)
 
 	ig := &servingv1alpha1.InferenceGraph{}
-	err := r.Client.Get(ctx, req.NamespacedName, ig)
+	err := r.Get(ctx, req.NamespacedName, ig)
 	if err != nil && apierrs.IsNotFound(err) {
 		if apierrs.IsNotFound(err) {
```
mwaykole, line 62: **Pre-existing bug exposed by this diff — worth fixing while you're here.**

The outer condition on line 61 is `if err != nil && apierrs.IsNotFound(err)`, which means the block is only entered when `err` IS a NotFound error. The inner `if apierrs.IsNotFound(err)` on line 62 is therefore always `true`, making the `logger.Error` on line 65 **dead code that can never be reached**.

Moreover, when `err != nil` but it is *not* a NotFound error (e.g., a transient API server timeout), the outer `if` is skipped entirely and the function falls through to `return ctrl.Result{}, nil` — silently swallowing the error with no log, no requeue.

Suggested fix:
```go
err := r.Get(ctx, req.NamespacedName, ig)
if err != nil {
    if apierrs.IsNotFound(err) {
        return ctrl.Result{}, nil
    }
    logger.Error(err, "Unable to get the InferenceGraph resource")
    return ctrl.Result{}, err
}
```
This properly handles the three cases: success, not-found, and real errors with a requeue.

internal/controller/serving/llm/llm_inferenceservice_controller_test.go
```diff
@@ -59,7 +59,7 @@ var _ = Describe("LLMInferenceService Controller", func() {
 		llmList := &kservev1alpha2.LLMInferenceServiceList{}
 		if err := envTest.Client.List(ctx, llmList, client.InNamespace(testNs)); err == nil {
 			for i := range llmList.Items {
-				_ = envTest.Client.Delete(ctx, &llmList.Items[i])
+				_ = envTest.Delete(ctx, &llmList.Items[i])
```
mwaykole, line 62: **Inconsistent access style in the same cleanup block.**

Line 60 still uses `envTest.Client.List(...)` (explicit field access), while this line was changed to `envTest.Delete(...)` (promoted method via embedded `client.Client`). Both are functionally identical at runtime since `pkgtest.Client` embeds `client.Client`, but the mixed style in the same `AfterEach` block is confusing — a reader might wonder if `envTest.Delete` and `envTest.Client.Delete` have different semantics.

Suggest making both consistent:
```go
// Either:
if err := envTest.List(ctx, llmList, client.InNamespace(testNs)); err == nil {
    for i := range llmList.Items {
        _ = envTest.Delete(ctx, &llmList.Items[i])
    }
}
// Or keep the .Client. prefix on both:
if err := envTest.Client.List(ctx, llmList, client.InNamespace(testNs)); err == nil {
    for i := range llmList.Items {
        _ = envTest.Client.Delete(ctx, &llmList.Items[i])
    }
}
```
The first form (no `.Client.` prefix) is more idiomatic since the PR's intent is to use the promoted methods.

internal/controller/serving/inferenceservice_controller.go
```diff
@@ -149,7 +149,7 @@ func (r *InferenceServiceReconciler) ReconcileServing(ctx context.Context, req c
 				deleteErrors = multierror.Append(deleteErrors, err1)
 			}
 			controllerutil.RemoveFinalizer(isvc, constants.InferenceServiceODHFinalizerName)
-			patchYaml := "metadata:\n  finalizers: [" + strings.Join(isvc.ObjectMeta.Finalizers, ",") + "]"
+			patchYaml := "metadata:\n  finalizers: [" + strings.Join(isvc.Finalizers, ",") + "]"
```
mwaykole, line 152: **Fragile YAML-by-string-interpolation for the finalizer patch (pre-existing, but touched by this PR).**

The pattern `"metadata:\n  finalizers: [" + strings.Join(isvc.Finalizers, ",") + "]"` produces syntactically invalid YAML/JSON if any finalizer name contains a `,` or `]`. While finalizer names in practice follow DNS subdomain naming (and therefore don't contain these chars), the approach is fragile and departs from the standard controller-runtime idiom.

The idiomatic approach that is safe and well-tested:
```go
// Use controllerutil + SSA patch, or simply:
if err := r.Update(ctx, isvc); err != nil {
    return reconcile.Result{}, err
}
```
Or if a patch is required, use `client.MergeFrom` with a deep copy:
```go
patch := client.MergeFrom(isvc.DeepCopy())
controllerutil.AddFinalizer(isvc, constants.InferenceServiceODHFinalizerName)
if err := r.Patch(ctx, isvc, patch); err != nil {
    return reconcile.Result{}, err
}
```
This is not introduced by this PR but is now visible. Consider a follow-up to clean it up.

internal/controller/utils/nim.go
```diff
@@ -132,7 +132,7 @@ func init() {
 			return true
 		},
 		func(a, b metav1.Time) bool {
-			return a.UTC() == b.UTC()
+			return a.UTC().Equal(b.UTC())
```
mwaykole, line 135: **Good correctness fix.** `a.UTC() == b.UTC()` compares `time.Time` structs using Go's `==` operator, which checks wall time, monotonic reading, and the `*Location` pointer identity. This can return `false` for times that represent the same instant if they were obtained differently (e.g., one from `time.Now()` and one deserialized from JSON). `a.UTC().Equal(b.UTC())` correctly compares only the time instant, ignoring the monotonic clock and location pointer. This is the right fix — nice catch by staticcheck.

internal/controller/processors/deltaProcessor.go
```diff
@@ -38,7 +38,7 @@ func (d *deltaProcessor) ComputeDelta(comparator comparators.ResourceComparator,
 	var updated bool
 	var removed bool
 
-	if !(utils.IsNil(desiredResource) && utils.IsNil(existingResource)) {
+	if !utils.IsNil(desiredResource) || !utils.IsNil(existingResource) {
```
mwaykole, line 41: **Correct and clearer.** The original `!(utils.IsNil(desired) && utils.IsNil(existing))` is equivalent to `!utils.IsNil(desired) || !utils.IsNil(existing)` by De Morgan's law — both mean "at least one of the two is not nil". The new form reads more naturally as a positive condition. No behavior change. ✅

.golangci.yml
```diff
@@ -45,11 +43,14 @@ linters:
           - dupl
           - lll
         path: internal/*
+      - linters:
+          - staticcheck
+        text: "ST1001"
+        path: "(test/|_test\\.go|internal/controller/testing/)"
```
mwaykole, line 49: **Minor: the path regex is not anchored, which may cause unexpected matches in the future.**

The pattern `(test/|_test\.go|internal/controller/testing/)` is matched as a substring of the file path. A file at, say, `mytest/something.go` or `unit_test.go` would match unintentionally. Consider anchoring or being more specific:
```yaml
path: "(^test/|_test\.go$|^internal/controller/testing/)"
```
This is low-risk today since the directory structure is stable, but worth tightening for correctness.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/opendatahub-io/odh-model-controller
 
-go 1.25.7
+go 1.25
```
mwaykole, line 3: **Informational note on the `go` / `toolchain` split.**

The base branch used `go 1.25.8` (fully pinned), which in Go 1.21+ means both the minimum version AND the toolchain version. Splitting into `go 1.25` + `toolchain go1.25.8` is the more expressive modern form: `go 1.25` declares the minimum compatible language version, while `toolchain go1.25.8` pins the exact toolchain for reproducible builds. This is a valid improvement — just documenting the intent for reviewers who may question why the `go` line appears to regress from `1.25.8` to `1.25`.

mwaykole (commented):
## PR Risk Assessment — chore: update golangci-lint to v2

**Overall Risk: LOW** — This is a well-scoped linter/toolchain upgrade. The code changes are purely mechanical staticcheck quick-fixes: promoted field access (`ObjectMeta.Labels → Labels`), `r.Client.Get → r.Get` via the embedded interface, and one genuine correctness fix in `nim.go`. No API contracts changed, no CRDs mutated, no RBAC modified.

### What's Good ✅
- `golangci.yml` v2 config format is correct; formatters section properly separated
- `r.Client.Get → r.Get` (and all similar) is idiomatic Go — the embedded `client.Client` interface means these are identical at runtime, just cleaner to read
- `ObjectMeta.Labels → Labels` / `ObjectMeta.Annotations → Annotations` are promoted field aliases — semantically identical, a genuine code quality improvement
- De Morgan's law rewrite in `deltaProcessor.go` (`!(A && B)` → `!A || !B`) is logically correct and clearer
- `clientset.DiscoveryClient.ServerGroupsAndResources()` → `clientset.ServerGroupsAndResources()` is correct since `kubernetes.Clientset` itself promotes the discovery methods
- GH Actions workflow correctly uses `golangci/golangci-lint-action@v7` for v2 support
- `go.mod` split of `go 1.25` + `toolchain go1.25.8` follows Go 1.21+ toolchain management best practice

### Issues Found

1. **`nim.go` — `time.UTC().Equal()` fix is correct but worth calling out** — `time.Time` comparison with `==` compares monotonic clock readings and pointer identity of the `Location`, making it unreliable across serialization boundaries. `.Equal()` is the right choice here.

2. **`inferencegraph_controller.go` — pre-existing double-IsNotFound check creates dead code and silently swallows non-NotFound errors** — this file was touched by the PR (the `r.Client.Get → r.Get` change), making it a good opportunity to fix it. Not introduced by this PR but now visible.

3. **`llm_inferenceservice_controller_test.go` — inconsistent access style introduced** — the PR changed `envTest.Client.Delete` → `envTest.Delete` on line 62 but left `envTest.Client.List` on line 60 unchanged, creating mixed style in the same cleanup block.

4. **`inferenceservice_controller.go` — YAML patch for finalizers uses string interpolation without quoting** — finalizer names are freeform strings; if one contained a `,` or `]`, the patch YAML would be malformed. This is pre-existing but the PR touches these lines.

5. **`.golangci.yml` — `ST1001` exclusion regex uses a non-anchored path match** — the pattern `(test/|_test\\.go|internal/controller/testing/)` would match any path containing those substrings. If a future path like `mytest/prod.go` exists, it would unexpectedly suppress ST1001 there.

No blocking issues — the PR is safe to merge as-is. The inline comments below are a mix of bugs worth fixing (inferencegraph dead code, test inconsistency) and informational notes.

PR #757, opened 2026-03-20
fix: Harden informer cache to prevent OOM from ConfigMap flooding

ugiordan:
## Summary

Hardens the controller-runtime informer cache to prevent out-of-memory crashes caused by cluster-wide ConfigMap caching.

**Problem**: The ConfigMap informer caches ALL ConfigMaps cluster-wide. Any cluster user with ConfigMap create permissions can create large ConfigMaps (~900KB each) across namespaces, causing unbounded memory growth that OOMKills the operator.

**Fix**:
- Add `opendatahub.io/managed=true` label selector to ConfigMap ByObject, restricting the informer to operator-managed ConfigMaps only
- Add `DisableFor` on ConfigMap so `client.Get()`/`client.List()` bypass the cache for direct API reads of external ConfigMaps (odh-trusted-ca-bundle, openshift-service-ca.crt)
- Add `cache.TransformStripManagedFields()` as DefaultTransform to reduce all cached object sizes
- Add `GOMEMLIMIT=1800MiB` (90% of 2Gi container limit) for GC pressure tuning

## Test Results

Tested on RHOAI 3.3 ROSA HCP cluster with 700 flood ConfigMaps (~630MB raw data) across 14 namespaces:

| Metric | Before Fix | After Fix |
|--------|-----------|-----------|
| Memory (512Mi limit) | OOMKilled in 12s | **17Mi stable** |
| OOM Restarts | Continuous | **0** |
| Memory Reduction | - | **~97%** |

## Ref

- [RHOAIENG-54558](https://redhat.atlassian.net/browse/RHOAIENG-54558)

## Summary by CodeRabbit

* **Chores**
  * Reads for ConfigMap and Secret now bypass the cache to fetch fresh data from the API.
  * Cached ConfigMap and Secret entries are sanitized to remove stored payloads and applied-configuration metadata.
  * Manager container adds a memory limit environment variable.
  * Secret-related reconcile logging verbosity reduced to a less prominent level.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -198,6 +198,36 @@ func setupNim(mgr manager.Manager, signalHandlerCtx context.Context,
 	}
 }
 
+// stripConfigMapData removes data payloads from cached ConfigMaps to reduce
+// memory consumption. Watch events still trigger reconciliation; actual data
+// is read via direct API calls (DisableFor).
+func stripConfigMapData(i interface{}) (interface{}, error) {
+	if cm, ok := i.(*corev1.ConfigMap); ok {
+		cm.Data = nil
+		cm.BinaryData = nil
+		if cm.Annotations != nil {
+			delete(cm.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
+		}
+		cm.SetManagedFields(nil)
+	}
+	return i, nil
+}
+
+// stripSecretData removes data payloads from cached Secrets to reduce
+// memory consumption. Watch events still trigger reconciliation; actual data
+// is read via direct API calls (DisableFor).
+func stripSecretData(i interface{}) (interface{}, error) {
+	if s, ok := i.(*corev1.Secret); ok {
+		s.Data = nil
+		s.StringData = nil
+		if s.Annotations != nil {
+			delete(s.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
+		}
+		s.SetManagedFields(nil)
+	}
+	return i, nil
+}
+
 func createManager(cfg *rest.Config, metricsAddr, probeAddr string,
 	enableLeaderElection, secureMetrics bool, tlsOpts []func(*tls.Config)) ctrl.Manager {
 	webhookServer := webhook.NewServer(webhook.Options{
@@ -245,14 +275,30 @@ func createManager(cfg *rest.Config, metricsAddr, probeAddr string,
 		// after the manager stops then its usage might be unsafe.
 		// LeaderElectionReleaseOnCancel: true,
 		Client: client.Options{
-			Cache: &client.CacheOptions{},
+			Cache: &client.CacheOptions{
+				// ConfigMap and Secret client reads bypass the cache
+				// entirely for direct API reads, since the informer
+				// caches stripped-down objects (no .data/.binaryData)
+				// to prevent unbounded memory growth.
+				DisableFor: []client.Object{
+					&corev1.ConfigMap{},
+					&corev1.Secret{},
+				},
+			},
 		},
 		Cache: cache.Options{
 			ByObject: map[client.Object]cache.ByObject{
+				// Strip data payloads from cached ConfigMaps and Secrets
+				// to prevent OOM from cluster-wide caching. The informer
+				// still delivers watch events for all objects (including
+				// external ones like odh-trusted-ca-bundle, NIM API key
+				// secrets, etc.), but cached objects are metadata-only.
+				// Actual data is read via direct API calls (DisableFor).
+				&corev1.ConfigMap{}: {
+					Transform: stripConfigMapData,
+				},
 				&corev1.Secret{}: {
-					Label: labels.SelectorFromSet(labels.Set{
-						"opendatahub.io/managed": "true",
-					}),
+					Transform: stripSecretData,
 				},
 				&corev1.Pod{}: {
 					Label: labels.SelectorFromSet(labels.Set{
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -203,14 +203,12 @@ func setupNim(mgr manager.Manager, signalHandlerCtx context.Context,
 // is read via direct API calls (DisableFor).
 func stripConfigMapData(i interface{}) (interface{}, error) {
 	if cm, ok := i.(*corev1.ConfigMap); ok {
-		cp := cm.DeepCopy()
-		cp.Data = nil
-		cp.BinaryData = nil
-		if cp.Annotations != nil {
-			delete(cp.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
+		cm.Data = nil
+		cm.BinaryData = nil
+		if cm.Annotations != nil {
+			delete(cm.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
 		}
-		cp.SetManagedFields(nil)
-		return cp, nil
+		cm.SetManagedFields(nil)
 	}
 	return i, nil
 }
@@ -220,14 +218,12 @@ func stripConfigMapData(i interface{}) (interface{}, error) {
 // is read via direct API calls (DisableFor).
 func stripSecretData(i interface{}) (interface{}, error) {
 	if s, ok := i.(*corev1.Secret); ok {
-		cp := s.DeepCopy()
-		cp.Data = nil
-		cp.StringData = nil
-		if cp.Annotations != nil {
-			delete(cp.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
+		s.Data = nil
+		s.StringData = nil
+		if s.Annotations != nil {
+			delete(s.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
 		}
-		cp.SetManagedFields(nil)
-		return cp, nil
+		s.SetManagedFields(nil)
 	}
 	return i, nil
 }
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -203,12 +203,14 @@ func setupNim(mgr manager.Manager, signalHandlerCtx context.Context,
 // is read via direct API calls (DisableFor).
 func stripConfigMapData(i interface{}) (interface{}, error) {
 	if cm, ok := i.(*corev1.ConfigMap); ok {
-		cm.Data = nil
-		cm.BinaryData = nil
-		if cm.Annotations != nil {
-			delete(cm.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
+		cp := cm.DeepCopy()
+		cp.Data = nil
+		cp.BinaryData = nil
+		if cp.Annotations != nil {
+			delete(cp.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
 		}
-		cm.SetManagedFields(nil)
+		cp.SetManagedFields(nil)
+		return cp, nil
 	}
 	return i, nil
 }
@@ -218,12 +220,14 @@ func stripConfigMapData(i interface{}) (interface{}, error) {
 // is read via direct API calls (DisableFor).
 func stripSecretData(i interface{}) (interface{}, error) {
 	if s, ok := i.(*corev1.Secret); ok {
-		s.Data = nil
-		s.StringData = nil
-		if s.Annotations != nil {
-			delete(s.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
+		cp := s.DeepCopy()
+		cp.Data = nil
+		cp.StringData = nil
+		if cp.Annotations != nil {
+			delete(cp.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
 		}
-		s.SetManagedFields(nil)
+		cp.SetManagedFields(nil)
+		return cp, nil
 	}
 	return i, nil
 }
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -72,6 +72,8 @@ spec:
             drop:
             - "ALL"
         env:
+          - name: GOMEMLIMIT
+            value: "1800MiB"
           - name: POD_NAMESPACE
             valueFrom:
               fieldRef:
--- a/internal/controller/serving/inferenceservice_controller.go
+++ b/internal/controller/serving/inferenceservice_controller.go
@@ -253,7 +253,7 @@ func (r *InferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager, setupLog
 					return []reconcile.Request{}
 				}
 				logger := log.FromContext(ctx)
-				logger.Info("Reconcile event triggered by Secret: " + o.GetName())
+				logger.V(1).Info("Reconcile event triggered by Secret: " + o.GetName())
 				isvc := &kservev1beta1.InferenceService{}
 				err := r.Client.Get(ctx, types.NamespacedName{Name: o.GetName(), Namespace: o.GetNamespace()}, isvc)
 				if err != nil {


config/manager/manager.yaml
```diff
@@ -72,6 +72,8 @@ spec:
             drop:
             - "ALL"
         env:
+          - name: GOMEMLIMIT
+            value: "1800MiB"
```
pierDipi: Is this necessary?

cmd/main.go
```diff
@@ -253,14 +275,30 @@ func createManager(cfg *rest.Config, metricsAddr, probeAddr string,
 		// after the manager stops then its usage might be unsafe.
 		// LeaderElectionReleaseOnCancel: true,
```
spolti, line 222: why removing this filter?

cmd/main.go
```diff
@@ -253,14 +275,30 @@ func createManager(cfg *rest.Config, metricsAddr, probeAddr string,
...
 		Cache: cache.Options{
 			ByObject: map[client.Object]cache.ByObject{
+				// Strip data payloads from cached ConfigMaps and Secrets
+				// to prevent OOM from cluster-wide caching. The informer
+				// still delivers watch events for all objects (including
+				// external ones like odh-trusted-ca-bundle, NIM API key
+				// secrets, etc.), but cached objects are metadata-only.
+				// Actual data is read via direct API calls (DisableFor).
+				&corev1.ConfigMap{}: {
+					Transform: stripConfigMapData,
```
spolti: We probably need to do a better job at defining what configMaps we will be caching. Probably define a label/annotation for all needed CMs.

spolti:
Hi, do we still want to merge it?

ugiordan:
@spolti I think so. Let's double-check with @bartoszmajsak and @pierDipi.

ugiordan:
@spolti are we good to merge this?

mwaykole (changes requested):
## PR Risk Assessment — `fix: Harden informer cache to prevent OOM from ConfigMap flooding`

**Overall Risk: MEDIUM-HIGH** — The approach is architecturally sound and the real-world test results are impressive (97% memory reduction), but there are two correctness issues worth addressing before merge: in-place mutation without `DeepCopy` in the transform functions (potential data race), and the removal of the Secret label selector that previously bounded the informer scope. The PR description also contains a misleading claim about a ConfigMap label selector that was never implemented.

### Positive aspects
- The `DisableFor` + cache `Transform` combo is the right pattern for this class of OOM problem.
- `GOMEMLIMIT` at the container level is a good defense-in-depth measure.
- Reducing the Secret reconcile log to `V(1)` is appropriate — it was very chatty at the informational level.
- `DisableFor` correctly ensures all `client.Get` / `client.List` calls in reconcilers go to the API server directly, so `ConfigMapReconciler.reconcileConfigMap`, `SecretReconciler.ReconcileStorageSecret`, and the NIM `APIKeyManager.loadSecret` all receive full objects with `.Data` intact. ✅

### Issues found
1. Transform functions mutate objects in-place without `DeepCopy` — risk of data race per controller-runtime docs.
2. Secret label selector (`opendatahub.io/managed=true`) was removed — informer scope regression.
3. No label selector added for ConfigMap informer — PR description claims this is added but the code doesn't implement it.
4. GOMEMLIMIT value doesn't match the stated 90% (actual: 87.9%).
5. No unit/integration test coverage for the new strip functions.

cmd/main.go
```diff
@@ -198,6 +198,36 @@ func setupNim(mgr manager.Manager, signalHandlerCtx context.Context,
 	}
 }
 
+// stripConfigMapData removes data payloads from cached ConfigMaps to reduce
+// memory consumption. Watch events still trigger reconciliation; actual data
+// is read via direct API calls (DisableFor).
+func stripConfigMapData(i interface{}) (interface{}, error) {
+	if cm, ok := i.(*corev1.ConfigMap); ok {
+		cm.Data = nil
```
mwaykole: **🔴 Data race risk: in-place mutation without `DeepCopy`**

The controller-runtime [`cache.TransformFunc`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/cache#TransformFunc) contract states that the transform function **must not modify the object in place**. When the SharedIndexInformer fires an `Update` event it keeps a reference to both the old and new objects for the `UpdateFunc` handler. Mutating `cm.Data = nil` directly on the passed-in pointer can corrupt the old-object reference and cause a data race under concurrent reconciliation.

**Suggested fix — return a deep copy with modifications:**
```go
func stripConfigMapData(i interface{}) (interface{}, error) {
    if cm, ok := i.(*corev1.ConfigMap); ok {
        cp := cm.DeepCopy()
        cp.Data = nil
        cp.BinaryData = nil
        if cp.Annotations != nil {
            delete(cp.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
        }
        cp.SetManagedFields(nil)
        return cp, nil
    }
    return i, nil
}
```
The same pattern applies to `stripSecretData` (line 221).

ugiordan: Fixed. Thanks for the feedback; applied the same to other PRs.

ugiordan: I reverted the DeepCopy. Since client-go v1.27, in-place mutation in TransformFunc is explicitly safe: https://github.com/kubernetes/client-go/blob/v0.35.2/tools/cache/delta_fifo.go#L148-L164.

This repo uses client-go v0.35.1, so the original code was correct. The DeepCopy added unnecessary allocation overhead.

cmd/main.go
```diff
@@ -176,6 +176,36 @@ func main() {
...
```
asanzgom: **[CONFIRMED] This breaks the gateway controller's ConfigMap predicate.**

Setting `cm.Data = nil` here causes the gateway controller's predicate at `gateway_controller.go:1077` to become permanently dead:

```go
return oldCM.Data["ingress"] != newCM.Data["ingress"]
```

Both `oldCM` and `newCM` come from the informer cache (Transform runs before predicate dispatch). With `Data` set to `nil`, both sides evaluate to `""` (nil map access returns zero value), so the comparison is always `false`. The gateway controller will never reconcile on `inferenceservice-config` changes.

`DisableFor` only affects `client.Get()`/`client.List()` — it does not affect watch events or predicates.

Consider: (1) always returning `true` for the ConfigMap name match (the Reconcile does a direct API read anyway), (2) using `oldCM.Generation != newCM.Generation`, or (3) excluding `inferenceservice-config` from the Transform.

cmd/main.go
```diff
@@ -176,6 +176,36 @@ func main() {
 	}
 }
 
+// stripConfigMapData removes data payloads from cached ConfigMaps to reduce
+// memory consumption. Watch events still trigger reconciliation; actual data
+// is read via direct API calls (DisableFor).
+func stripConfigMapData(i interface{}) (interface{}, error) {
```
asanzgom: **[CONFIRMED] Reimplements `cache.TransformStripManagedFields()`.**

controller-runtime v0.19.7 already provides `cache.TransformStripManagedFields()` which handles the managed-fields and `last-applied-configuration` stripping. The identical annotation deletion + `SetManagedFields(nil)` logic is duplicated between `stripConfigMapData` and `stripSecretData`.

Also, `DefaultTransform` is not set, so all other cached types (Pods, ISVCs, ServingRuntimes) still carry full managed fields. Setting `DefaultTransform: cache.TransformStripManagedFields()` would strip managed fields from **all** cached objects with a single line, providing additional memory savings.

Suggested approach: set `DefaultTransform` for managed-fields stripping across all types, and only use per-type Transforms for the Data/BinaryData stripping specific to Secrets and ConfigMaps.

cmd/main.go
```diff
@@ -221,14 +251,30 @@ func createManager(cfg *rest.Config, metricsAddr, probeAddr string,
 		// after the manager stops then its usage might be unsafe.
 		// LeaderElectionReleaseOnCancel: true,
 		Client: client.Options{
-			Cache: &client.CacheOptions{},
+			Cache: &client.CacheOptions{
+				// ConfigMap and Secret client reads bypass the cache
+				// entirely for direct API reads, since the informer
+				// caches stripped-down objects (no .data/.binaryData)
+				// to prevent unbounded memory growth.
+				DisableFor: []client.Object{
```
asanzgom, line 228: **[PLAUSIBLE] `DisableFor` shifts 20+ Secret/ConfigMap reads from cache to API server.**

There are ~14 `client.Get` call sites for Secrets and ~6 for ConfigMaps across the controllers. All of these now bypass the cache and hit the API server directly. The SecretReconciler alone makes 5 Gets per reconcile.

Under the ConfigMap flood scenario this PR targets, the flood generates many reconcile events. Each reconciliation now makes multiple direct API calls instead of cache hits. Combined with the broadened Secret event scope (finding below), API server request volume increases. Under sustained load, Kubernetes priority-and-fairness may throttle the controller with 429 responses.

This is a known tradeoff (OOM vs API server load), but worth documenting as a design decision and monitoring in production.

cmd/main.go
```diff
@@ -221,14 +251,30 @@ func createManager(cfg *rest.Config, metricsAddr, probeAddr string,
...
+				// Strip data payloads from cached ConfigMaps and Secrets
+				// to prevent OOM from cluster-wide caching. The informer
+				// still delivers watch events for all objects (including
+				// external ones like odh-trusted-ca-bundle, NIM API key
+				// secrets, etc.), but cached objects are metadata-only.
+				// Actual data is read via direct API calls (DisableFor).
+				&corev1.ConfigMap{}: {
+					Transform: stripConfigMapData,
+				},
 				&corev1.Secret{}: {
```
asanzgom, line 243: **[PLAUSIBLE] Removing the label selector causes cluster-wide Secret event processing.**

The old `opendatahub.io/managed=true` label selector meant the informer only watched matching Secrets. Now the informer watches **all** Secrets cluster-wide. Every Secret create/update/delete triggers the ISVC controller's `MapFunc` at `inferenceservice_controller.go:252`, which does a `client.Get` for a matching InferenceService — and with `DisableFor`, that Get now hits the API server directly.

On a production OpenShift cluster with thousands of Secrets and high churn (cert-manager rotations, SA token refresh), this generates continuous CPU overhead from processing irrelevant events.

Consider keeping a label selector or adding a namespace selector to limit the watch scope.

config/manager/manager.yaml
```diff
@@ -72,6 +72,8 @@ spec:
             drop:
             - "ALL"
         env:
+          - name: GOMEMLIMIT
```
asanzgom, line 76: **[PLAUSIBLE] `GOMEMLIMIT` is hardcoded independently from the container memory limit.**

`GOMEMLIMIT=1800MiB` while the container memory limit is `2Gi` (2048MiB). There's no mechanism to keep them in sync.

If a downstream overlay changes the memory limit to 1Gi without updating `GOMEMLIMIT`, the Go runtime targets 1800MiB heap before aggressive GC — but the container OOM-kills at 1Gi, so `GOMEMLIMIT` provides no protection. Conversely, increasing to 4Gi without updating `GOMEMLIMIT` wastes CPU on unnecessarily aggressive GC.

Consider adding a comment documenting the relationship (e.g., `# ~90% of memory limit (2Gi)`) or using the automemlimit library to derive it from cgroup limits at runtime.

internal/controller/serving/inferenceservice_controller.go
```diff
@@ -254,7 +254,7 @@ func (r *InferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager, setupLog
 					return []reconcile.Request{}
 				}
 				logger := log.FromContext(ctx)
-				logger.Info("Reconcile event triggered by Secret: " + o.GetName())
+				logger.V(1).Info("Reconcile event triggered by Secret: " + o.GetName())
```
asanzgom: **[PLAUSIBLE] `V(1)` masks the symptom of the broadened event scope.**

With the label selector removed, this log line now fires for **every** Secret event in the cluster. Changing it to `V(1)` hides the noise at default verbosity, but it also hides evidence of whether a Secret event was received at all.

When debugging a case where updating a data connection secret doesn't trigger ISVC reconciliation, operators would need to restart the controller with `-v=1` — which on a large cluster produces a flood of entries for every irrelevant system secret, burying the actionable log line.

Consider instead filtering the events at the source (e.g., adding a predicate to the Secret watch that only passes through secrets matching an ISVC name) rather than silencing the log.

cmd/main.go
ugiordan: Fair point. I'll set `DefaultTransform: cache.TransformStripManagedFields()` and remove the manual `SetManagedFields(nil)` + `last-applied-configuration` deletion from the per-type transforms, keeping them focused on Data/BinaryData stripping only.

config/manager/manager.yaml
ugiordan: I'll add a comment documenting the relationship. The automemlimit library is a nice idea but adds a runtime dependency for something a comment solves. If the container limit changes, the deployer needs to update GOMEMLIMIT too.

internal/controller/serving/inferenceservice_controller.go
ugiordan: Agreed that filtering at the source is better than silencing the log. The V(1) change was a stopgap for the noise from the broadened Secret scope. If we add a predicate to the Secret watch (per the suggestion on the label selector comment), the V(1) becomes appropriate since irrelevant events would already be filtered before reaching this code path.

cmd/main.go
ugiordan: The label selector was removed intentionally after @bartoszmajsak identified in an earlier review round (https://github.com/opendatahub-io/odh-model-controller/pull/757#discussion_r3001972392) that it broke reactions to external Secrets the controller needs to watch (NIM API keys, data connection secrets that don't carry the opendatahub.io/managed label). The tradeoff between scope filtering and functional correctness is documented in https://developers.redhat.com/articles/2026/07/06/5-anti-patterns-cause-kubernetes-operator-vulnerabilities (Anti-pattern 1).

cmd/main.go
ugiordan: This is a known tradeoff we documented in https://developers.redhat.com/articles/2026/07/06/5-anti-patterns-cause-kubernetes-operator-vulnerabilities (Pattern A section). The alternative is unbounded memory growth from cluster-wide caching, which is worse than the latency increase from direct reads (~50ms vs ~1ms). Under normal operation, the load is modest since predicates filter most events before reconciliation. I'll add a comment documenting this as a design decision.

cmd/main.go
ugiordan: Fixed.

cmd/main.go
asanzgom: Sounds good. Cleaner to let `DefaultTransform` handle managedFields globally and keep the per-type transforms focused on data stripping only.

config/manager/manager.yaml
asanzgom: Agree, a comment is sufficient. No need for a runtime dependency here.

internal/controller/serving/inferenceservice_controller.go
asanzgom: Makes sense as a follow-up plan. Just want to flag that without the predicate in this PR, the V(1) change reduces visibility for a problem (noisy Secret events from the broadened scope) that this PR introduces. If the Secret watch predicate is a separate PR, consider leaving the log at Info until then so the noise is visible and motivates adding it.

cmd/main.go
asanzgom: Understood — the label selector removal is justified for correctness, and the article documents the tradeoff well. Worth considering a name-based or owner-reference predicate on the Secret watch in a follow-up to reduce event processing volume, but that can be a separate PR.

cmd/main.go
asanzgom: Agree, the tradeoff is sound. A documenting comment is a good idea since future contributors might wonder why ConfigMap/Secret reads are slower than expected.

cmd/main.go
asanzgom: Thanks!

internal/controller/serving/llm/gateway_controller.go
```diff
@@ -1072,9 +1072,10 @@ func (r *GatewayReconciler) SetupWithManager(mgr ctrl.Manager, setupLog logr.Log
 					if podNS := os.Getenv("POD_NAMESPACE"); podNS != "" && e.ObjectNew.GetNamespace() != podNS {
 						return false
 					}
-					oldCM := e.ObjectOld.(*corev1.ConfigMap)
-					newCM := e.ObjectNew.(*corev1.ConfigMap)
-					return oldCM.Data["ingress"] != newCM.Data["ingress"]
+					// Always reconcile on generation change. Cannot compare
+					// .Data here because the cache transform strips it to
+					// reduce memory (see stripConfigMapData in cmd/main.go).
+					return e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration()
```
asanzgom: Correcting my own suggestion here — option (2) from my review (`GetGeneration()`) does not work for ConfigMaps. The Kubernetes API server only increments `metadata.generation` for resources with a spec/status split. ConfigMaps have no `/status` subresource, so their generation stays at `1` for the lifetime of the object regardless of data changes. This predicate will always return `false` and the gateway controller will never reconcile on `inferenceservice-config` updates.

The simplest correct fix is option (1) — just `return true` after the namespace check:

```go
return true
```

The reconciler already does a direct API read via `DisableFor`, so the extra reconciliation cost is negligible for a single ConfigMap in the controller namespace. The comment can note that data comparison is not possible due to cache stripping.

ugiordan: Done!

cmd/main.go
```diff
@@ -176,6 +176,28 @@ func main() {
 	}
 }
 
+// stripConfigMapData removes data payloads from cached ConfigMaps to reduce
+// memory consumption. Watch events still trigger reconciliation; actual data
+// is read via direct API calls (DisableFor).
+func stripConfigMapData(i interface{}) (interface{}, error) {
+	if cm, ok := i.(*corev1.ConfigMap); ok {
```
mwaykole: The transform functions mutate the input directly. While `DisableFor` ensures reconcilers read from the API server, predicates still receive these stripped objects from the informer.

This is fine as long as no predicate logic depends on `.Data` — and the gateway controller predicate was already fixed to handle this.

**Confirm:** Are there any other predicates in the codebase that access `ConfigMap.Data` or `Secret.Data`? A quick grep would be prudent before merge.

internal/controller/serving/llm/gateway_controller.go
```diff
@@ -1072,9 +1072,11 @@ func (r *GatewayReconciler) SetupWithManager(mgr ctrl.Manager, setupLog logr.Log
 					if podNS := os.Getenv("POD_NAMESPACE"); podNS != "" && e.ObjectNew.GetNamespace() != podNS {
 						return false
 					}
-					oldCM := e.ObjectOld.(*corev1.ConfigMap)
-					newCM := e.ObjectNew.(*corev1.ConfigMap)
-					return oldCM.Data["ingress"] != newCM.Data["ingress"]
+					// ConfigMaps have no spec/status split so metadata.generation
+					// is never incremented. Return true to reconcile on any update
```
mwaykole, line 1076: The fix is correct. ConfigMaps have no spec/status split, so `metadata.generation` stays at 1 forever — the old generation-based comparison would never pass.

Since the name (`InferenceServiceConfigMapName`) and namespace guards already scope this to a single ConfigMap, returning `true` is appropriate. The Reconcile function does a direct API read anyway.

internal/controller/serving/inferenceservice_controller.go
mwaykole: Per @asanzgom's earlier comment: with the label selector removed, this log now fires for every Secret event cluster-wide. `V(1)` hides the noise but also reduces debuggability.

Consider leaving this at `Info` until a name-based or owner-reference predicate is added to the Secret watch (raised as a follow-up by @asanzgom). That way the noise motivates adding the predicate rather than being silently suppressed.

mwaykole (commented):
## Follow-up Review

CI is green and the major issues from previous rounds have been addressed (gateway predicate fix, DefaultTransform for managed fields). The implementation is sound.

**Minor items remaining:**

1. **PR description is stale** — The description still mentions "Add `opendatahub.io/managed=true` label selector to ConfigMap ByObject" which was never implemented. The actual approach is Transform + DisableFor. Consider updating for clarity.

2. **Test suite cache topology divergence** — Noted by @asanzgom, the test suites (`suite_test.go`) still use the old cache config. This is acceptable as a follow-up but worth tracking.

cmd/main.go
asanzgom: Grepped the full codebase — the only predicate that accessed `ConfigMap.Data` was the gateway controller at line 1077, and it's already fixed in this PR (replaced the `Data["ingress"]` comparison with an unconditional `return true` since ConfigMaps don't increment metadata.generation).

No other predicate in the codebase accesses ConfigMap.Data, ConfigMap.BinaryData, Secret.Data, or Secret.StringData. All `.Data` accesses are in reconciler logic, which reads via the API reader (DisableFor), not the informer cache.

internal/controller/serving/inferenceservice_controller.go
asanzgom: +1 on keeping this at Info until a Secret watch predicate is added. The noise is a useful signal that the scope is broader than it needs to be — suppressing it with V(1) removes the incentive to fix it properly.

@ugiordan can you revert the V(1) change on this line? The follow-up predicate PR can downgrade it then.

asanzgom:
All review items from the latest round are addressed or acknowledged:

  - **Transform + DisableFor approach** — confirmed safe. Grepped the full codebase: the only predicate that accessed `ConfigMap.Data` was the gateway controller (line 1077), and it's already fixed in this PR. No other predicate accesses `.Data` or `.BinaryData`. All data reads happen in reconciler logic via the API reader, not the informer cache.
  - **Gateway controller predicate fix** — correct. `return true` is appropriate since ConfigMaps don't increment `metadata.generation`, and the name/namespace guards already scope to a single ConfigMap.
  - **Secret event log level** — agree with @mwaykole and my earlier comment. @ugiordan please revert the `V(1) `change on the Secret event log (line 257 in `inferenceservice_controller.go`) back to Info. The noise is a useful signal until a Secret watch predicate is added as a follow-up.
  - **Stale PR description** — @ugiordan the description still references the label selector approach. Quick update to reflect the actual Transform + DisableFor implementation would help reviewers.

With those two minor fixes (log level revert + description update), this is ready for serving team approval. This is a Blocker for the Developer Sandbox RHOAI 3.x upgrade ([RHOAIENG-54558](https://redhat.atlassian.net/browse/RHOAIENG-54558) (https://redhat.atlassian.net/browse/RHOAIENG-54558)) — @mwaykole or serving team PTAL.

[RHOAIENG-54558]: https://redhat.atlassian.net/browse/RHOAIENG-54558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

asanzgom:
The xKS Smoke Test failure **is not caused by this PR**. It's a pre-existing issue: the base kustomization now includes new `ClusterServingRuntime` resources (vllm-spyre CSRs merged recently), but the xKS test setup script (`test/scripts/kind/setup-odh-model-controller-xks.sh`) doesn't install the `ClusterServingRuntime` CRD. The xKS overlay uses `$patch: delete` to strip CSRs, but kustomize still needs the CRD registered to process them.

Fix: `add the serving.kserve.io_clusterservingruntimes.yaml` CRD to `test/crds/` and install it in the setup script alongside the LLMInferenceService CRD (line 52). This is independent of the informer cache changes in this PR.

cmd/main.go
spolti: Same here, I am not sure if there is a win on doing it, it will remove cache in size but increase api calls.

cmd/main.go
```diff
@@ -176,6 +176,28 @@ func main() {
...
+		cm.Data = nil
+		cm.BinaryData = nil
+	}
+	return i, nil
+}
+
+// stripSecretData removes data payloads from cached Secrets to reduce
+// memory consumption. Watch events still trigger reconciliation; actual data
+// is read via direct API calls (DisableFor).
+func stripSecretData(i interface{}) (interface{}, error) {
```
spolti: The big gain of caching is quick read it from the cache and don't hit the api, why we need it?

internal/controller/serving/inferenceservice_controller.go
ugiordan: I've added the full label predicate in this PR. @mwaykole, could you please check?

config/base/kustomization.yaml
```diff
@@ -38,12 +38,7 @@ replacements:
           kind: Template
           name: ovms
         fieldPaths:
-          - objects.0.spec.containers.0.image
```
spolti, line 41: are these changes related?

config/manager/manager.yaml
spolti: it sets the soft limit of the go runtime to be more aggressive in the garbage collection, setting to the pod limit may lead to have OOM anyways as it will hit the pod limit before getting more aggressive with GC

spolti (changes requested):

bartoszmajsak (commented):
If we want the OOM fix to land for the people hitting it now, maybe split the Secret-related changes into their own PR? They're two changes with opposite risk profiles - one shrinks the cache, the other widens the watch scope, and the second one is where most of the review round trips came from.

@ugiordan can we have a minimal set for the ConfigMap OOM fix for now:

- `GOMEMLIMIT` back to 1800MiB (~88%, matches the other RHOAI operators). Better knobs later if we want them.
- `StripConfigMapData` only, with `ManagedFields = nil` and no `DeepCopy`
- ConfigMap `Transform` + `DisableFor`. Keep `DefaultTransform` - `Secrets` and `Pods` then get `managedFields` stripping for free since they have no per-type Transform, and ConfigMap is the only one needing the manual nil
- Gateway predicate fix - without it the transform silently kills gateway reconciliation on `inferenceservice-config` changes. We are stripping data so there's nothing to compare against anymore.
- `strip_test.go` covering the above
 
Since reads now bypass the cache, did we get any stress test numbers on `rest_client_rate_limiter_duration_seconds` and `workqueue_queue_duration_seconds`?

cmd/main.go
```diff
@@ -221,14 +222,32 @@ func createManager(cfg *rest.Config, metricsAddr, probeAddr string,
...
+				// external ones like odh-trusted-ca-bundle, NIM API key
+				// secrets, etc.), but cached objects are metadata-only.
+				// Actual data is read via direct API calls (DisableFor).
+				// See: https://developers.redhat.com/articles/2026/07/06/5-anti-patterns-cause-kubernetes-operator-vulnerabilities
+				&corev1.ConfigMap{}: {
+					Transform: informercache.StripConfigMapData,
+				},
 				&corev1.Secret{}: {
-					Label: labels.SelectorFromSet(labels.Set{
-						"opendatahub.io/managed": "true",
```
spolti, line 230: this is the label to filter our resources. Why remove it?

ugiordan: Fair point — restored in `b73ed1beb`. This PR is now ConfigMap OOM only:

- **Secret** informer keeps `opendatahub.io/managed=true` (unchanged from `incubating`)
- **ConfigMap** uses strip transform + `DisableFor` so we still watch external CAs without caching `.data`

Secret strip / predicate changes are dropped from this PR.

bartoszmajsak:
@coderabbitai full review

bartoszmajsak (approved):
/lgtm

Few non-blocking observations:

1. `AGENTS.md:151` still only documents the label-filtered caches. Worth a sentence saying ConfigMaps are cached metadata-only and that reads bypass the cache
2.  The gateway predicate comment blames `metadata.generation` - I think it's better to point out that  `.Data` is nil in the cache, so the old comparison was always `false`. Adding for posterity will definitely be appreciated.

cmd/main.go
bartoszmajsak: @ugiordan are you going to follow-up with Secret-related changes in a new PR?

ugiordan:
Addressed review #4965084113 in `9a1254d06` (AGENTS.md + gateway predicate comment).

Secret follow-up: #923 — keeps the label filter, adds strip + DisableFor + ISVC predicate only. Depends on #757.

cmd/main.go
ugiordan: https://github.com/opendatahub-io/odh-model-controller/pull/923

PR #767, opened 2026-03-30
fix: provision get RBAC for LLMInferenceService inference access

amasolov:
## Summary

The Gateway-level `AuthPolicy` (`maas-default-gateway-authn`) created by `KserveAuthPolicyReconciler` performs a `SubjectAccessReview` with `verb: get` on each `LLMInferenceService` to authorise inference requests ([template source](internal/controller/resources/template/authpolicy_llm_isvc_userdefined.yaml#L36-L43)). However, no component ever creates the corresponding `Role`/`RoleBinding`, so **every inference call returns `403 Forbidden`** regardless of any other auth configuration.

### Root cause

`maas-default-gateway-authn` overrides `gateway-auth-policy` (the legacy tier-based policy) at the Gateway level. The old tier system created `post` RBAC for tier SAs, but `maas-default-gateway-authn` checks `get` — and nothing provisions `get` RBAC for end users.

```
HTTP/1.1 403 Forbidden
x-ext-auth-reason: not authorized: unknown reason
```

### Fix

Add `KserveRBACReconciler` as a new `LLMSubResourceReconciler` that creates per-`LLMInferenceService`:

- **`Role`** — grants `get` on the specific `LLMInferenceService` (scoped by `resourceNames`)
- **`RoleBinding`** — binds to `system:authenticated`

The `system:authenticated` default is intentional: the actual per-model authorisation is already enforced at the HTTPRoute level by `MaaSAuthPolicy`/subscription policies. The gateway-level SAR only needs to verify the token is a real authenticated cluster identity. Platform operators who want stricter per-model access control can delete the generated `RoleBinding` and create their own.

Both resources carry an `OwnerReference` to the `LLMInferenceService` and are garbage-collected automatically on deletion.

## How to reproduce

1. Deploy an `LLMInferenceService` (model becomes ready)
2. Call the inference endpoint with a valid SA token
3. Without this fix: `403 Forbidden` with `x-ext-auth-reason: not authorized`
4. With this fix: request passes the SAR check and reaches the model

## Testing

- Unit: `KserveRBACReconciler` follows the same patterns as `KserveAuthPolicyReconciler` and `KserveEnvoyFilterReconciler`
- Integration: verified manually on OCP 4.17 + RHOAI 3.3 cluster — inference calls succeed after reconciler provisions the RBAC

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * LLM Inference Service now provisions a namespace-scoped Role and RoleBinding so authenticated users gain access to inference endpoints.

* **Behavior**
  * Controller automatically creates and reconciles RBAC rules/subjects, restores drifted changes, and recreates bindings when Role references are immutable; resources labeled unmanaged are left unchanged.

* **Tests**
  * Added tests for RBAC creation, restoration, immutable-reference handling, owner references, and opt-out behavior.

### DIFF
--- a/internal/controller/serving/llm/fixture/rbac_helpers.go
+++ b/internal/controller/serving/llm/fixture/rbac_helpers.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fixture
+
+import (
+	"context"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const inferenceAccessSuffix = "-llmisvc-inference-access"
+
+func GetInferenceAccessRoleName(llmisvcName string) string {
+	return llmisvcName + inferenceAccessSuffix
+}
+
+func VerifyInferenceAccessRoleExists(ctx context.Context, c client.Client, namespace, llmisvcName string) {
+	VerifyResourceExists(ctx, c, namespace, GetInferenceAccessRoleName(llmisvcName), &rbacv1.Role{})
+}
+
+func VerifyInferenceAccessRoleBindingExists(ctx context.Context, c client.Client, namespace, llmisvcName string) {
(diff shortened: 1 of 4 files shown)


bartoszmajsak:
@amasolov I think the issue is due to misconfiguration of the gateway and LLMInferenceService. New MaaS solution attaches their own policies that are evaluated through subscription model. `maas-default-gateway-authn` is the policy indetended for "regular inference flow", not when MaaS machinery is involved and thus it shouldn't be attached to maas gateway. 

Perhaps you are missing a label (`opendatahub.io/managed: "false"`) on the gateway to opt-out for managing it and avoid policy conflicts.

mwaykole (commented):
Test review check

bartoszmajsak:
I think setting RBAC is user responsibility. 

FWIW we do have some aggregation logic for these https://github.com/opendatahub-io/kserve/blob/76da12e77e42af9c46d347a0a3b6eaba7afc8546/config/overlays/odh/rbac/user-cluster-roles.yaml

Should we close it @amasolov?

PR #827, opened 2026-05-27
feat(webhook): migrate ConnectionsAPI injection to odh-model-controller

israel-hdez:
## Description

Migrates the ConnectionsAPI mutating webhook from `opendatahub-operator` to `odh-model-controller`.

- Introduces an internal `connectionapi` package with shared utilities for resolving `opendatahub.io/connections` annotations and injecting credentials into inference resources (S3, URI, OCI types).
- Extends `InferenceServiceCustomDefaulter` (v1beta1) to handle all connection types on CREATE and UPDATE operations (inject, remove, replace, and no-op actions), including S3 path priority, dry-run support, and backward compatibility with the deprecated `connection-type-ref` annotation.
- Adds a new `LLMInferenceServiceCustomDefaulter` (v1alpha2) with the same connection-type coverage, plus a hybrid unstructured approach to cleanly remove `spec.model` on cleanup.
- Registers the new `LLMInferenceService` mutating webhook in `cmd/main.go` and updates the generated `config/webhook/manifests.yaml`.
- Adds 38 Ginkgo specs for the ISVC defaulter (`v1beta1`) and 43 Ginkgo specs for the LLMInferenceService defaulter (`v1alpha2`), both using `controller-runtime` fake clients.

Related Jira: https://redhat.atlassian.net/browse/RHOAIENG-62537

## How Has This Been Tested?

- Unit tests for the ISVC defaulter run under the existing envtest-based Ginkgo suite in `internal/webhook/serving/v1beta1/`; the new specs use fake clients and do not interact with envtest infrastructure.
- Unit tests for the LLMInferenceService defaulter run under a new minimal Ginkgo suite in `internal/webhook/serving/v1alpha2/` with no envtest dependency: `go test ./internal/webhook/serving/v1alpha2/...`
- Full test suite can be run via `make test` once envtest binaries are available.

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Added ConnectionsAPI mutating webhook for LLMInferenceService to auto-inject and clean up credentials; supports S3, URI, and OCI connection types and respects dry-run.

* **Chores**
  * Webhook registrations, manifests, and kustomization updated; mutating webhooks set dry-run sideEffects.
  * InferenceService/LLMInferenceService defaulters enhanced to handle update/cleanup and use API reader for prior-state logic.
  * Added connection utilities and test secret builders.

* **Tests**
  * Extensive suites covering create/update/dry-run flows, edge cases, and backward compatibility.

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -48,14 +48,33 @@ webhooks:
     service:
       name: webhook-service
       namespace: system
-      path: /mutate-serving-kserve-io-v1alpha2-llminferenceservice
+      path: /mutate-serving-kserve-io-v1alpha1-llminferenceservice
   failurePolicy: Fail
-  name: connection-llmisvc.odh-model-controller.opendatahub.io
+  name: connection-llmisvc-v1alpha1.odh-model-controller.opendatahub.io
   rules:
   - apiGroups:
     - serving.kserve.io
     apiVersions:
     - v1alpha1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - llminferenceservices
+  sideEffects: NoneOnDryRun
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /mutate-serving-kserve-io-v1alpha2-llminferenceservice
+  failurePolicy: Fail
+  name: connection-llmisvc-v1alpha2.odh-model-controller.opendatahub.io
+  rules:
+  - apiGroups:
+    - serving.kserve.io
+    apiVersions:
     - v1alpha2
     operations:
     - CREATE
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -48,26 +48,6 @@ webhooks:
     service:
       name: webhook-service
       namespace: system
-      path: /mutate-serving-kserve-io-v1beta1-inferenceservice
-  failurePolicy: Fail
-  name: minferenceservice-v1beta1.odh-model-controller.opendatahub.io
-  rules:
-  - apiGroups:
-    - serving.kserve.io
-    apiVersions:
-    - v1beta1
-    operations:
-    - CREATE
-    - UPDATE
-    resources:
-    - inferenceservices
-  sideEffects: NoneOnDryRun
-- admissionReviewVersions:
-  - v1
-  clientConfig:
-    service:
-      name: webhook-service
-      namespace: system
       path: /mutate-serving-kserve-io-v1alpha2-llminferenceservice
   failurePolicy: Fail
   name: connection-llmisvc.odh-model-controller.opendatahub.io
@@ -82,6 +62,26 @@ webhooks:
     - UPDATE
     resources:
     - llminferenceservices
+  sideEffects: NoneOnDryRun
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /mutate-serving-kserve-io-v1beta1-inferenceservice
+  failurePolicy: Fail
+  name: minferenceservice-v1beta1.odh-model-controller.opendatahub.io
+  rules:
+  - apiGroups:
+    - serving.kserve.io
+    apiVersions:
+    - v1beta1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - inferenceservices
   sideEffects: NoneOnDryRun
 ---
 apiVersion: admissionregistration.k8s.io/v1
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -58,9 +58,31 @@ webhooks:
     - v1beta1
     operations:
     - CREATE
+    - UPDATE
     resources:
     - inferenceservices
-  sideEffects: None
+  sideEffects: NoneOnDryRun
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /mutate-serving-kserve-io-v1alpha2-llminferenceservice
+  failurePolicy: Fail
+  name: connection-llmisvc.odh-model-controller.opendatahub.io
+  rules:
+  - apiGroups:
+    - serving.kserve.io
+    apiVersions:
+    - v1alpha1
+    - v1alpha2
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - llminferenceservices
+  sideEffects: NoneOnDryRun
 ---
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
--- a/internal/webhook/connectionapi/utils.go
+++ b/internal/webhook/connectionapi/utils.go
@@ -1,0 +1,395 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package connectionapi
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"slices"
+
+	corev1 "k8s.io/api/core/v1"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// ConnectionInfo holds connection-related information extracted from resource annotations and the referenced Secret.
+type ConnectionInfo struct {
+	// SecretName is the name of the Secret from the opendatahub.io/connections annotation.
+	SecretName string
+	// Type is the connection type from the connection-type-protocol or connection-type-ref annotation on the Secret.
+	Type string
+	// Path is the S3 sub-path from the opendatahub.io/connection-path annotation on the resource.
+	Path string
+}
+
+// ConnectionAction represents the mutation action to perform on a resource during admission.
+type ConnectionAction string
+
+const (
+	// ConnectionActionInject injects connection credentials into the resource.
+	ConnectionActionInject ConnectionAction = "inject"
+	// ConnectionActionRemove removes previously injected connection credentials.
+	ConnectionActionRemove ConnectionAction = "remove"
+	// ConnectionActionReplace removes old credentials and injects new ones.
+	ConnectionActionReplace ConnectionAction = "replace"
+	// ConnectionActionNone indicates no mutation is needed.
+	ConnectionActionNone ConnectionAction = "none"
+)
+
+// ConnectionType is the canonical identifier for a connection protocol.
+type ConnectionType string
+
+const (
+	// ConnectionTypeProtocolS3 identifies S3-compatible object storage connections.
+	ConnectionTypeProtocolS3 ConnectionType = "s3"
+	// ConnectionTypeProtocolURI identifies generic URI connections.
+	ConnectionTypeProtocolURI ConnectionType = "uri"
+	// ConnectionTypeProtocolOCI identifies OCI container image connections.
+	ConnectionTypeProtocolOCI ConnectionType = "oci"
+
+	// ConnectionTypeRefS3 is the deprecated ref value for S3 connections.
+	ConnectionTypeRefS3 ConnectionType = "s3"
+	// ConnectionTypeRefURI is the deprecated ref value for URI connections.
+	ConnectionTypeRefURI ConnectionType = "uri-v1"
+	// ConnectionTypeRefOCI is the deprecated ref value for OCI connections.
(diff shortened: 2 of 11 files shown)


internal/webhook/serving/v1alpha2/llminferenceservice_webhook_test.go
```diff
@@ -0,0 +1,905 @@
...
+			})
+			newLLM := buildLLMISVC(nil)
+			newLLM.Spec.Template = &corev1.PodSpec{ServiceAccountName: llmS3SecretName + "-sa"}
+			// Model.URI will be removed by cleanup
+			admCtx := llmAdmissionCtx(admissionv1.Update, oldLLM, false)
+
+			Expect(d.Default(admCtx, newLLM)).To(Succeed())
+			Expect(newLLM.Spec.Template.ServiceAccountName).To(BeEmpty())
+			// After cleanup, Model.URI is zero
+			Expect(newLLM.Spec.Model.URI.String()).To(BeEmpty())
```
israel-hdez, line 409: This check looks buggy, since the URI is not set in the original `newLLM`.

internal/webhook/serving/v1alpha2/llminferenceservice_webhook_test.go
```diff
@@ -0,0 +1,905 @@
...
+			cli := newLLMFakeClient(secret)
+			d := newLLMDefaulter(cli)
+			oldLLM := buildLLMISVC(map[string]string{
+				connectionapi.AnnotationConnections: llmURISecretName,
+			})
+			newLLM := buildLLMISVC(nil)
+			admCtx := llmAdmissionCtx(admissionv1.Update, oldLLM, false)
+
+			Expect(d.Default(admCtx, newLLM)).To(Succeed())
+			Expect(newLLM.Spec.Model.URI.String()).To(BeEmpty())
```
israel-hdez, line 427: Also here, check looks buggy, since the URI is not set in the original `newLLM`.

internal/webhook/serving/v1alpha2/llminferenceservice_webhook_test.go
```diff
@@ -0,0 +1,905 @@
...
+			d := newLLMDefaulter(cli)
+			oldLLM := buildLLMISVC(map[string]string{
+				connectionapi.AnnotationConnections: llmURISecretName,
+			})
+			newLLM := buildLLMISVC(nil)
+			admCtx := llmAdmissionCtx(admissionv1.Update, oldLLM, false)
+
+			Expect(d.Default(admCtx, newLLM)).To(Succeed())
+			// After removeModelSpec, Model.URI is zero-valued
+			Expect(newLLM.Spec.Model.URI.String()).To(BeEmpty())
```
israel-hdez, line 558: Same here, check looks buggy, since the URI is not set in the original `newLLM`.

internal/webhook/serving/v1beta1/inferenceservice_connectionsapi_test.go
```diff
@@ -0,0 +1,621 @@
...
+// buildSecretWithRef creates a Secret with the deprecated connection-type-ref annotation.
+func buildSecretWithRef(name, ns, refType string, data map[string][]byte) *corev1.Secret {
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name, Namespace: ns,
+			Annotations: map[string]string{connectionapi.AnnotationConnectionTypeRef: refType},
+		},
+		Data: data,
+	}
+}
```
israel-hdez: Check if these functions can be extracted and reused, since this looks like duplicated code with `buildLLMSecret` and `buildLLMSecretWithRef`.

internal/webhook/serving/v1beta1/inferenceservice_connectionsapi_test.go
```diff
@@ -0,0 +1,621 @@
...
+				newISVC := buildISVC(map[string]string{
+					connectionapi.AnnotationConnections:    defaultS3Secret,
+					connectionapi.AnnotationConnectionPath: "models/v1",
+				})
+				newISVC.Spec.Predictor.Model = &kservev1beta1.ModelSpec{}
+
+				Expect(newISVCDefaulter(cli).Default(defaultCtx(admissionv1.Update, oldISVC, false), newISVC)).To(Succeed())
+				Expect(newISVC.Spec.Predictor.ServiceAccountName).To(Equal(defaultS3SA))
+				Expect(newISVC.Spec.Predictor.Model.Storage.StorageKey).To(HaveValue(Equal(defaultS3Secret)))
+				Expect(newISVC.Spec.Predictor.Model.Storage.Path).To(HaveValue(Equal("models/v1")))
```
israel-hdez, line 236: There is no check to ensure the SA is created.

internal/webhook/connectionapi/utils.go
```diff
@@ -0,0 +1,395 @@
...
+// Parameters:
+//   - ctx: context for API calls and logging
+//   - cli: write-capable client for ServiceAccount creation
+//   - secretName: name of the connection Secret (SA will be named secretName+"-sa")
+//   - connType: connection type string
+//   - namespace: namespace for the ServiceAccount
+//   - isDryRun: when true, skips actual SA creation
+//
+// Returns any error from ServiceAccount creation.
+func ServiceAccountCreation(ctx context.Context, cli client.Client, secretName, connType, namespace string, isDryRun bool) error {
```
israel-hdez: Rename to `CreateServiceAccountIfNeeded`.

internal/webhook/serving/v1alpha2/llminferenceservice_webhook.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	// Phase 2: remove spec.model when spec.model.uri is set.
+	// Uses the hybrid unstructured approach because Model is a non-pointer non-omitempty struct
+	// and its URI field type (apis.URL) cannot be zeroed without failing CRD validation.
+	if llmisvc.Spec.Model.URI.String() != "" {
+		if err := removeModelSpec(llmisvc); err != nil {
+			return fmt.Errorf("failed to remove spec.model: %w", err)
+		}
+	}
```
israel-hdez: I don't think we need to do this processing, because when Unmarshalling back to a typed var, the field will be defaulted to a zero value. Let's simplify and set spec.model to a zero value instance.

config/webhook/manifests.yaml
```diff
@@ -42,6 +42,27 @@ webhooks:
...
+      namespace: system
+      path: /mutate-serving-kserve-io-v1alpha2-llminferenceservice
+  failurePolicy: Fail
+  name: connection-llmisvc.odh-model-controller.opendatahub.io
+  rules:
+  - apiGroups:
+    - serving.kserve.io
+    apiVersions:
+    - v1alpha1
+    - v1alpha2
```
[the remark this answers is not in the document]
Jooho: I think this is right comment. do we need to support v1alpha1 now? then we need to remove this version from manifests.yaml
or if we need v1alpha1 then we should add it to webhook file.

internal/webhook/connectionapi/utils.go
```diff
@@ -0,0 +1,395 @@
...
+				Path:       oldObj.GetAnnotations()[AnnotationConnectionPath],
+			}, nil
+		}
+		return ConnectionInfo{}, fmt.Errorf("failed to get old secret metadata: %w", err)
+	}
+
+	connType := secretMeta.GetAnnotations()[AnnotationConnectionTypeProtocol]
+	if connType == "" {
+		connType = secretMeta.GetAnnotations()[AnnotationConnectionTypeRef]
+	}
```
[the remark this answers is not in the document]
israel-hdez: Won't to, rationale:

This does not fix a TOCTOU. The only practical difference is that unrecognized type strings now fall through to the full-cleanup path explicitly instead of silently hitting no switch branch. However, the existing performISVCCleanup / performLLMISVCCleanup already handle the empty-type case as a full cleanup (the "" branch). The suggested change is a no-op in practice. 

Additionally, this is how opendatahub-operator works currently.

internal/webhook/connectionapi/utils.go
```diff
@@ -0,0 +1,395 @@
...
+
+// createSA creates a ServiceAccount linking the given secret. AlreadyExists is treated as success.
+func createSA(ctx context.Context, cli client.Client, secretName, namespace string) error {
+	sa := &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName + "-sa",
+			Namespace: namespace,
+		},
+		Secrets: []corev1.ObjectReference{{Name: secretName}},
+	}
```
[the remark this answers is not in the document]
israel-hdez: Won't do. Rationale: adding owner references is a deliberate design choice excluded from scope to preserve the behavior of the existing opendatahub-operator implementation.

internal/webhook/serving/v1alpha2/llminferenceservice_webhook.go
```diff
@@ -0,0 +1,310 @@
...
+			if oldConn.SecretName != "" {
+				connectionapi.CleanupOCIImagePullSecrets(&(*template).ImagePullSecrets, oldConn.SecretName)
+			} else {
+				(*template).ImagePullSecrets = nil
+			}
+		}
+	}
+
+	// Phase 2: zero spec.model.uri if it was previously set.
+	if modelURI.String() != "" {
```
Jooho, line 307: This seems different from the original logic. Previously, it removed the model field, whereas now it sets spec.model.uri to an empty string ("").

I'm not sure whether this behavioral difference is intentional. Is that expected?

https://github.com/opendatahub-io/opendatahub-operator/blob/580a7c65/internal/webhook/serving/mutating_llmisvc.go#L303-L309

Jooho (commented):
From what I can tell, the implementation was mostly moved over from the existing opendatahub-operator, so I don’t see any major migration concerns with the code itself.

However, there’s one thing I’m worried about, and I think members of the platform(@StevenTobin ) team or the LLMD squad(@pierDipi) might know the answer. LLMISVC is supported on XKS, but odh-model-controller is not deployed there. That means the `Connection API functionality` being added here would not be available on XKS.

Previously, this code lived in `opendatahub-operator`, so the Connection API was available on XKS as well. With this change, it seems that XKS would lose that capability.

Before we proceed, I think we should confirm whether that is actually the intended approach. If the Connection API is still required on XKS(or was it not supported already?), then it seems like this logic should live in the KServe module instead. What do you think?

valdar:
> From what I can tell, the implementation was mostly moved over from the existing opendatahub-operator, so I don’t see any major migration concerns with the code itself.
> 
> However, there’s one thing I’m worried about, and I think members of the platform(@StevenTobin ) team or the LLMD squad(@pierDipi) might know the answer. LLMISVC is supported on XKS, but odh-model-controller is not deployed there. That means the `Connection API functionality` being added here would not be available on XKS.
> 
> Previously, this code lived in `opendatahub-operator`, so the Connection API was available on XKS as well. With this change, it seems that XKS would lose that capability.
> 
> Before we proceed, I think we should confirm whether that is actually the intended approach. If the Connection API is still required on XKS(or was it not supported already?), then it seems like this logic should live in the KServe module instead. What do you think?

@pierDipi you might have some more context here?

internal/webhook/serving/v1alpha2/llminferenceservice_webhook.go
israel-hdez: Yes, it is different but IMO it is quite equivalent from controller perspective.

In the typed structs, these are not pointers, so these cannot be set to nil and would default to zero value. Thus, emptying is what made sense to me here when using the types.

pierDipi:
For context, currently connection API is not necessarily needed on xKS and we're not documenting it there.

Since there is an effort to bring the UI on xKS slowly (@andyatmiami), can we keep it as it was before on the module operator side?

israel-hdez:
@pierDipi I talked with Jooho and we will keep the webhook here. We will create a new Jira to add adapt odh-model-controller to work in xKS.

internal/webhook/connectionapi/utils.go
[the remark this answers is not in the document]
spolti: @israel-hdez so what happens when the llmisvc or isvc is deleted?

internal/webhook/serving/v1alpha2/llminferenceservice_webhook_test.go
```diff
@@ -0,0 +1,889 @@
...
+limitations under the License.
+*/
+
+package v1alpha2
+
+import (
+	"context"
+	"encoding/json"
+
+	kservev1alpha2 "github.com/kserve/kserve/pkg/apis/serving/v1alpha2"
```
spolti, line 24: shouldn't we also add one test for v1alpha1?

israel-hdez: Merely for sanity, added one basic test in commit a66894a3

pierDipi:
> @pierDipi I talked with Jooho and we will keep the webhook here. We will create a new Jira to add adapt odh-model-controller to work in xKS.

can you expand why we're moving it? the PR description and messages don't have any reasoning for this change.
architecture changes should be discussed separately from modularization

pierDipi:
Dashboard is doing xKS PoCs on the old architecture, this is a breaking change and creating additional work that we would need to discuss and work on.

israel-hdez:
> can you expand why we're moving it? the PR description and messages don't have any reasoning for this change.
architecture changes should be discussed separately from modularization

It is a separation of concerns, since operators are not supposed to have non-setup tasks.

> Dashboard is doing xKS PoCs on the old architecture, this is a breaking change and creating additional work that we would need to discuss and work on.

I'm not sure I agree that this is a breaking change. It will work identically as before.
The only additional work is to make odh-model-controller compatible with xKS. I just created the Jira for that: https://redhat.atlassian.net/browse/RHOAIENG-69082

pierDipi:
> I'm not sure I agree that this is a breaking change. It will work identically as before.
> The only additional work is to make odh-model-controller compatible with xKS. I just created the Jira for that: https://redhat.atlassian.net/browse/RHOAIENG-69082

well, that task is there and it's unclear when it will be done, dashboard team is already working on running dashboard on xKS

internal/webhook/serving/v1alpha2/llminferenceservice_webhook.go
Jooho: fair enough.

PR #841, opened 2026-06-15
feat: integrate with cluster TLS security profile

ugiordan:
## Summary

- Read the cluster-wide TLS security profile from the OpenShift APIServer CR and apply it to the webhook and metrics servers
- TLS resolution code extracted into `pkg/tls/` for reusability and testability
- `pkg/tls.Resolve()` uses unstructured access (zero new dependencies, compatible with controller-runtime v0.22.5)
- On non-OpenShift clusters or when the profile is unavailable, falls back to hardened defaults (TLS 1.2, Intermediate ciphers, ALPN h2/http1.1)
- Handles all profile types: Intermediate (default with 6 ECDHE ciphers), Modern (TLS 1.3), Old (TLS 1.0), and Custom (user-defined ciphers/version)
- Fails closed on unexpected errors (RBAC failures, transport errors)
- Replaces the old `disableHTTP2` / `enable-http2` flag with proper ALPN negotiation (`h2`, `http/1.1`)
- Adds RBAC rule for `config.openshift.io/apiservers` (get/list/watch, resourceNames=cluster)

## Motivation

OCP 5.0 (GA October 2026) requires all components to honor the centralized TLS profile ([OCPSTRAT-2611](https://redhat.atlassian.net/browse/OCPSTRAT-2611)).

Reference: [openshift/cluster-machine-approver #286](https://github.com/openshift/cluster-machine-approver/pull/286)

## Files changed

- `cmd/main.go`: Replaced ~150 lines of inline TLS code with 4-line call to `pkgtls.Resolve()`
- `pkg/tls/tls.go`: Reusable TLS profile resolution (unstructured APIServer read, cipher mapping, Intermediate defaults)
- `pkg/tls/tls_test.go`: Table-driven tests for all profile types
- `Containerfile`: Added `COPY pkg/ pkg/` for container builds

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./pkg/tls/... -v` passes (12 table-driven tests)
- [ ] Deploy on OpenShift cluster, verify TLS profile is read from APIServer CR
- [ ] Deploy on vanilla k8s, verify fallback to hardened defaults (TLS 1.2)

Ref: [RHOAIENG-61065](https://redhat.atlassian.net/browse/RHOAIENG-61065)

[OCPSTRAT-2611]: https://redhat.atlassian.net/browse/OCPSTRAT-2611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/main.go
```diff
@@ -397,3 +384,157 @@ func setupGatewayReconciler(mgr ctrl.Manager) error {
...
+		}
+		tlsOpts = append(tlsOpts, func(c *tls.Config) { c.MinVersion = tls.VersionTLS12 })
+	} else {
+		minVersion, ciphers := parseTLSProfile(apiServer)
+		setupLog.Info("Applying cluster TLS profile", "minVersion", minVersion, "ciphers", len(ciphers))
+		tlsOpts = append(tlsOpts, func(c *tls.Config) {
+			c.MinVersion = minVersion
+			if len(ciphers) > 0 {
+				c.CipherSuites = ciphers
+			}
```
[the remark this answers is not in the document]
ugiordan: @coderabbitai Addressed with a different (stronger) approach.
Instead of the boolean flag approach, we:

  1. Changed `parseTLSProfile` to use `ciphers := make([]uint16, 0, len(cipherNames))` in the Custom case, so it returns a non-nil empty slice (distinguishable from nil "no restriction") when all ciphers are unsupported.

  2. The caller checks `ciphers != nil && len(ciphers) == 0` and calls `os.Exit(1)`, refusing to start entirely. This is stronger than setting an empty CipherSuites (which Go treats as "use all defaults").

  3. Regression test added: "Custom with all unsupported ciphers returns empty slice" verifies `parseTLSProfile` returns the empty non-nil slice for this scenario.

  Regarding TLS 1.3 cipher filtering: our `openSSLToGoCipher` map only contains TLS 1.2 constants. TLS 1.3 cipher names (TLS_AES_128_GCM_SHA256 etc.) use different OpenSSL names than TLS 1.2 ciphers, so they don't match any map entry and
  get dropped with a log message. No additional filtering needed.

cmd/main.go
pierDipi, line 297: why is this removed?

pkg/tls/tls.go
```diff
@@ -0,0 +1,183 @@
...
+
+	k8sClient, err := client.New(cfg, client.Options{})
+	if err != nil {
+		return result, fmt.Errorf("creating bootstrap client for TLS profile: %w", err)
+	}
+
+	apiServer := &unstructured.Unstructured{}
+	apiServer.SetGroupVersionKind(schema.GroupVersionKind{
+		Group: "config.openshift.io", Version: "v1", Kind: "APIServer",
+	})
```
pierDipi: is there a reason we're using unstructured instead of typed objects?

cmd/main.go
ugiordan: Accidental removal during the refactor, sorry about that. Restored!

pkg/tls/tls.go
ugiordan: fixed

PR #850, opened 2026-06-22
fix(release): add go build smoke test after kserve dep update

jlost:
## Summary
- Add `go build ./...` after `go mod tidy` in the release workflow's kserve dependency update step
- Catches type mismatches (e.g. renamed API types) before commit/tag/push
- Restructure dry run to validate full tidy+build pipeline instead of exiting early

## Context
The release workflow updates the kserve go.mod replace directive but previously didn't verify that odh-model-controller still compiles against the new version. This caused silent build failures when kserve API types changed (e.g. `UntypedObjectReference` → `GatewayObjectReference`), only caught later during Konflux container builds.

## Test plan
- [ ] Trigger workflow with `dry_run: true` targeting `odh-model-controller` — verify build check runs and passes
- [ ] Trigger workflow with a known-incompatible kserve tag — verify build check fails and blocks release

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **Chores**
  * Improved the dependency update workflow by ensuring the updated module file is written in both dry-run and non-dry-run modes.
  * Added a validation sequence that runs a module tidy followed by a full build smoke test, with clearer failure/timeout behavior.
  * Enhanced logging to confirm the resolved kserve version after a successful update.
  * Updated dry-run completion messaging to reflect “changes not committed.”

### DIFF
--- a/.github/workflows/odh-release.yaml
+++ b/.github/workflows/odh-release.yaml
@@ -384,13 +384,21 @@ jobs:
             f.write(new_content)
 
           print("Running go mod tidy...")
-          result = subprocess.run(['go', 'mod', 'tidy'], capture_output=True, text=True)
+          try:
+            result = subprocess.run(['go', 'mod', 'tidy'], capture_output=True, text=True, timeout=300)
+          except subprocess.TimeoutExpired:
+            print(f"::error::go mod tidy timed out after 300s")
+            raise SystemExit(1)
           if result.returncode != 0:
             print(f"::error::go mod tidy failed:\n{result.stderr}")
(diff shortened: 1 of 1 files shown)


vivekk16 (commented):
Worth addressing the CodeRabbit timeout comment before merge and apply the same pattern to the `go mod tidy` call above it for consistency. A stalled module proxy on either can hang the job indefinitely.

vivekk16 (approved):
/lgtm
/approve

pierDipi:
Great follow up, thanks!

PR #854, opened 2026-06-23
test(e2e): verify AuthPolicy injects flow control headers on inference paths

### LINKED ISSUE #853: test(e2e): verify AuthPolicy injects flow control headers for authenticated and anonymous requests
## What

Add e2e tests verifying that the AuthPolicy templates inject `x-gateway-inference-fairness-id`
and `x-gateway-inference-objective` headers on inference paths.

## Tests

1. **SA request**: `fairness-id` = cluster issuer, `objective` = SA namespace (CEL extraction)
2. **Cross-namespace**: two SAs from different namespaces get different objective values

## Why

If the headers aren't injected correctly, all tenants get default priority 0 and fairness ID
`default-flow`, silently disabling multi-tenant isolation in flow control.

## References

- `internal/controller/resources/template/FLOW_CONTROL.md`
- `internal/controller/resources/template/authpolicy_llm_isvc_userdefined.yaml`
- Pattern: `internal/controller/test/e2e/batch_authpolicy_test.go`

RishabhSaini:
## Summary

Verify that the user-defined AuthPolicy injects `x-gateway-inference-fairness-id` and
`x-gateway-inference-objective` headers on inference paths. Two tests:

- **TestFlowControlHeadersSA**: SA gets `fairness-id=https://kubernetes.default.svc`,
  `objective={SA namespace}` (CEL extraction from `system:serviceaccount:{ns}:{name}`)
- **TestFlowControlHeadersCrossNamespace**: two SAs from different namespaces get
  different objective values, confirming tenant isolation

Uses inference paths only. Independent of batch. Follows `batch_authpolicy_test.go` pattern.
Requires `make test-e2e-controller` (controller + gateway + Authorino deployed).

Closes #853

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end tests validating flow-control header injection on gateway inference responses
  * Added coverage for single-namespace behavior with expected fairness and objective values
  * Added cross-namespace scenarios to confirm each caller receives the correct tenant objective based on the requesting namespace

### DIFF
--- a/internal/controller/test/e2e/flowcontrol_authpolicy_test.go
+++ b/internal/controller/test/e2e/flowcontrol_authpolicy_test.go
@@ -3,11 +3,28 @@
 package e2e
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
-	"strings"
 	"testing"
 )
+
+// echoResponse captures the subset of the ealen/echo-server JSON response
+// needed to inspect reflected request headers.
+type echoResponse struct {
+	Request struct {
+		Headers map[string]string `json:"headers"`
+	} `json:"request"`
+}
+
+func parseEchoHeaders(t *testing.T, body []byte) map[string]string {
+	t.Helper()
+	var resp echoResponse
+	if err := json.Unmarshal(body, &resp); err != nil {
+		t.Fatalf("failed to parse echo server response: %v\nbody: %s", err, body)
+	}
+	return resp.Request.Headers
+}
 
 // TestFlowControlHeadersSA verifies that the AuthPolicy injects the correct
 // flow control headers for an authenticated ServiceAccount on an inference path.
@@ -30,18 +47,18 @@ func TestFlowControlHeadersSA(t *testing.T) {
 		t.Fatalf("expected 200, got %d", resp.StatusCode)
 	}
 
-	bodyStr := string(body)
+	headers := parseEchoHeaders(t, body)
 
-	if !strings.Contains(bodyStr, "x-gateway-inference-fairness-id") {
-		t.Error("response does not contain x-gateway-inference-fairness-id header")
-	}
-	if !strings.Contains(bodyStr, "https://kubernetes.default.svc") {
-		t.Error("fairness-id should be the cluster issuer (https://kubernetes.default.svc)")
+	if val, ok := headers["x-gateway-inference-fairness-id"]; !ok {
+		t.Error("x-gateway-inference-fairness-id header not injected")
+	} else if val != "https://kubernetes.default.svc" {
+		t.Errorf("fairness-id = %q, want %q", val, "https://kubernetes.default.svc")
 	}
 
-	expectedObjective := fmt.Sprintf("x-gateway-inference-objective\":%q", ns)
-	if !strings.Contains(bodyStr, expectedObjective) && !strings.Contains(bodyStr, fmt.Sprintf("x-gateway-inference-objective\": %q", ns)) {
-		t.Errorf("objective header value should be %q, body: %s", ns, bodyStr)
+	if val, ok := headers["x-gateway-inference-objective"]; !ok {
+		t.Error("x-gateway-inference-objective header not injected")
+	} else if val != ns {
+		t.Errorf("objective = %q, want namespace %q", val, ns)
 	}
 }
 
@@ -70,19 +87,21 @@ func TestFlowControlHeadersCrossNamespace(t *testing.T) {
 	if respA.StatusCode != http.StatusOK {
 		t.Fatalf("tenant A: expected 200, got %d", respA.StatusCode)
 	}
-	expectedA := fmt.Sprintf("x-gateway-inference-objective\":%q", nsA)
-	bodyAStr := string(bodyA)
-	if !strings.Contains(bodyAStr, expectedA) && !strings.Contains(bodyAStr, fmt.Sprintf("x-gateway-inference-objective\": %q", nsA)) {
-		t.Errorf("tenant A objective header value should be %q, body: %s", nsA, bodyAStr)
+	headersA := parseEchoHeaders(t, bodyA)
+	if val, ok := headersA["x-gateway-inference-objective"]; !ok {
+		t.Error("tenant A: x-gateway-inference-objective header not injected")
+	} else if val != nsA {
+		t.Errorf("tenant A: objective = %q, want %q", val, nsA)
 	}
 
 	respB, bodyB := batchEnv.gatewayGet(t, path, tokenB, nil)
 	if respB.StatusCode != http.StatusOK {
 		t.Fatalf("tenant B: expected 200, got %d", respB.StatusCode)
 	}
-	expectedB := fmt.Sprintf("x-gateway-inference-objective\":%q", nsB)
-	bodyBStr := string(bodyB)
-	if !strings.Contains(bodyBStr, expectedB) && !strings.Contains(bodyBStr, fmt.Sprintf("x-gateway-inference-objective\": %q", nsB)) {
-		t.Errorf("tenant B objective header value should be %q, body: %s", nsB, bodyBStr)
+	headersB := parseEchoHeaders(t, bodyB)
+	if val, ok := headersB["x-gateway-inference-objective"]; !ok {
+		t.Error("tenant B: x-gateway-inference-objective header not injected")
+	} else if val != nsB {
+		t.Errorf("tenant B: objective = %q, want %q", val, nsB)
 	}
 }
--- a/internal/controller/test/e2e/flowcontrol_authpolicy_test.go
+++ b/internal/controller/test/e2e/flowcontrol_authpolicy_test.go
@@ -13,17 +13,31 @@ import (
 // needed to inspect reflected request headers.
 type echoResponse struct {
 	Request struct {
-		Headers map[string]string `json:"headers"`
+		Headers map[string]json.RawMessage `json:"headers"`
 	} `json:"request"`
 }
 
-func parseEchoHeaders(t *testing.T, body []byte) map[string]string {
+// parseEchoHeader extracts a single header value from the echo server response.
+// Handles both string ("value") and array (["value"]) formats.
+func parseEchoHeader(t *testing.T, body []byte, header string) (string, bool) {
 	t.Helper()
 	var resp echoResponse
 	if err := json.Unmarshal(body, &resp); err != nil {
 		t.Fatalf("failed to parse echo server response: %v\nbody: %s", err, body)
 	}
-	return resp.Request.Headers
+	raw, ok := resp.Request.Headers[header]
+	if !ok {
+		return "", false
+	}
+	var s string
+	if json.Unmarshal(raw, &s) == nil {
+		return s, true
+	}
+	var arr []string
+	if json.Unmarshal(raw, &arr) == nil && len(arr) > 0 {
+		return arr[0], true
+	}
+	return "", false
 }
 
 // TestFlowControlHeadersSA verifies that the AuthPolicy injects the correct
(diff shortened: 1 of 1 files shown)


internal/controller/test/e2e/flowcontrol_authpolicy_test.go
```diff
@@ -0,0 +1,80 @@
...
+	if !strings.Contains(bodyStr, "https://kubernetes.default.svc") {
+		t.Error("fairness-id should be the cluster issuer (https://kubernetes.default.svc)")
+	}
+
+	if !strings.Contains(bodyStr, "x-gateway-inference-objective") {
+		t.Error("response does not contain x-gateway-inference-objective header")
+	}
+	if !strings.Contains(bodyStr, ns) {
+		t.Errorf("objective should be the SA namespace %q", ns)
+	}
```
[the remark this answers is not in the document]
RishabhSaini: follows existing batch pattern

internal/controller/test/e2e/flowcontrol_authpolicy_test.go
```diff
@@ -0,0 +1,86 @@
...
+		t.Error("response does not contain x-gateway-inference-fairness-id header")
+	}
+	if !strings.Contains(bodyStr, "https://kubernetes.default.svc") {
+		t.Error("fairness-id should be the cluster issuer (https://kubernetes.default.svc)")
+	}
+
+	if !strings.Contains(bodyStr, "x-gateway-inference-objective") {
+		t.Error("response does not contain x-gateway-inference-objective header")
+	}
+	if !strings.Contains(bodyStr, ns) {
```
mwaykole: **🐛 False positive — this assertion will ALWAYS pass, even if the header is not injected.**

The `ealen/echo-server` reflects the full HTTP request (including the request path) back in its JSON response body. Since the request path is `/{ns}/echo-server/v1/chat/completions`, the namespace string `ns` is already present in `bodyStr` from the URL — making `strings.Contains(bodyStr, ns)` trivially true whether or not `x-gateway-inference-objective` was ever set by the AuthPolicy.

This is the same pattern the existing batch tests avoid — they check for the full SA identity string (`system:serviceaccount:{ns}:{name}`) which is unique and cannot appear in the request path or other reflected fields.

To make this assertion meaningful, check for the full CEL-derived value that only appears in the header, such as:
```go
// The CEL expression yields only the namespace (split(':')[2]) so check for a
// pattern that uniquely identifies the header value vs the reflected path.
if !strings.Contains(bodyStr, `"x-gateway-inference-objective":"` + ns + `"`) {
    t.Errorf("x-gateway-inference-objective header value should be namespace %q", ns)
}
```
Or, alternatively, parse the JSON body and check the header map directly.

internal/controller/test/e2e/flowcontrol_authpolicy_test.go
```diff
@@ -0,0 +1,86 @@
...
+	batchEnv.grantInferenceAccess(t, nsA, nsB, "tenant-user")
+	tokenB := batchEnv.requestToken(t, nsB, "tenant-user")
+
+	path := fmt.Sprintf("/%s/echo-server/v1/chat/completions", nsA)
+
+	respA, bodyA := batchEnv.gatewayGet(t, path, tokenA, nil)
+	if respA.StatusCode != http.StatusOK {
+		t.Fatalf("tenant A: expected 200, got %d", respA.StatusCode)
+	}
+	if !strings.Contains(string(bodyA), nsA) {
```
mwaykole: **🐛 Same false-positive issue as line 45 in `TestFlowControlHeadersSA`.**

The request path is `/{nsA}/echo-server/v1/chat/completions`. The echo server reflects this path in `bodyA`, so `strings.Contains(string(bodyA), nsA)` is always true — it does not confirm that the `x-gateway-inference-objective` header contains `nsA`.

Note that `bodyB` (line 83) does NOT have this problem: `nsA` is in the path but `nsB` is not, so `strings.Contains(string(bodyB), nsB)` only passes if the header actually carries `nsB`. This asymmetry means the test only verifies isolation for tenant B — tenant A's assertion is a no-op.

Fix by checking the header value in the same way suggested for line 45, e.g.:
```go
if !strings.Contains(string(bodyA), `"x-gateway-inference-objective":"` + nsA + `"`) {
    t.Errorf("tenant A objective should contain namespace %q in header value", nsA)
}
```

internal/controller/test/e2e/flowcontrol_authpolicy_test.go
```diff
@@ -0,0 +1,86 @@
...
+	if resp.StatusCode != http.StatusOK {
+		t.Fatalf("expected 200, got %d", resp.StatusCode)
+	}
+
+	bodyStr := string(body)
+
+	if !strings.Contains(bodyStr, "x-gateway-inference-fairness-id") {
+		t.Error("response does not contain x-gateway-inference-fairness-id header")
+	}
+	if !strings.Contains(bodyStr, "https://kubernetes.default.svc") {
```
mwaykole: **⚠️ Hardcoded cluster issuer URL — environment-specific assumption.**

The string `"https://kubernetes.default.svc"` is hardcoded from the AuthPolicy template's `fairness.value` override. While this matches the current template, it's a cluster-environment assumption: custom clusters may configure a different issuer URL. If this ever changes (e.g., in an OIDC-configured cluster), this assertion will silently fail in CI.

Consider reading the expected value from an environment variable with the hardcoded value as a default:
```go
clusterIssuer := os.Getenv("CLUSTER_ISSUER_URL")
if clusterIssuer == "" {
    clusterIssuer = "https://kubernetes.default.svc"
}
if !strings.Contains(bodyStr, clusterIssuer) {
    t.Errorf("fairness-id should be the cluster issuer %q", clusterIssuer)
}
```
Alternatively, add a comment explicitly documenting this is the audience value set by the AuthPolicy template's `fairness.value` field so reviewers know where it comes from.

internal/controller/test/e2e/flowcontrol_authpolicy_test.go
```diff
@@ -0,0 +1,86 @@
...
+	batchEnv.createHTTPRoute(t, nsA, "echo-inference", fmt.Sprintf("/%s/echo-server", nsA))
+	batchEnv.createServiceAccount(t, nsA, "tenant-user")
+	batchEnv.grantInferenceAccess(t, nsA, nsA, "tenant-user")
+	tokenA := batchEnv.requestToken(t, nsA, "tenant-user")
+	batchEnv.waitForGatewayRoute(t, fmt.Sprintf("/%s/echo-server/test", nsA), tokenA)
+
+	nsB := batchEnv.createNamespace(t, "e2e-fc", nil)
+	batchEnv.createServiceAccount(t, nsB, "tenant-user")
+	batchEnv.grantInferenceAccess(t, nsA, nsB, "tenant-user")
+	tokenB := batchEnv.requestToken(t, nsB, "tenant-user")
```
mwaykole, line 94: **⚠️ Missing readiness gate for tenant B before sending a request.**

For tenant A, `waitForGatewayRoute` (line 62) is called after obtaining `tokenA`, ensuring the HTTPRoute is accepting traffic before the actual assertion request is made. Tenant B (`tokenB`) has no equivalent wait.

While tenant B's request targets the same `nsA` path (already confirmed reachable via `tokenA`'s wait), the RBAC grant at line 66 (`grantInferenceAccess(t, nsA, nsB, "tenant-user")`) adds a new RoleBinding to `nsA` immediately before the request. The Kubernetes RBAC cache may not have propagated this binding by the time Authorino's SAR check runs, which could cause a transient 403 and a flaky test.

Consider adding an explicit wait or small sleep, or use `waitForGatewayRoute` with `tokenB` to confirm the new binding is effective before asserting on the response:
```go
batchEnv.waitForGatewayRoute(t, fmt.Sprintf("/%s/echo-server/test", nsA), tokenB)
```

internal/controller/test/e2e/flowcontrol_authpolicy_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	path := fmt.Sprintf("/%s/echo-server/v1/chat/completions", ns)
+	resp, body := batchEnv.gatewayGet(t, path, token, nil)
+	if resp.StatusCode != http.StatusOK {
+		t.Fatalf("expected 200, got %d", resp.StatusCode)
+	}
+
+	bodyStr := string(body)
+
+	if !strings.Contains(bodyStr, "x-gateway-inference-fairness-id") {
```
mwaykole: **💡 Suggestion: co-locate header name + value checks to make failure messages more actionable.**

Currently, lines 35–46 check the header NAME and its VALUE in separate `if` blocks. If the header is present but has the wrong value (e.g., Authorino CEL expression changed), you get two failures without clear causality. Combining them or at least printing the actual body section makes it easier to diagnose:

```go
if !strings.Contains(bodyStr, "x-gateway-inference-fairness-id") {
    t.Errorf("response does not contain x-gateway-inference-fairness-id header; body: %s", bodyStr)
} else if !strings.Contains(bodyStr, clusterIssuer) {
    t.Errorf("x-gateway-inference-fairness-id value should be %q; body: %s", clusterIssuer, bodyStr)
}
```
This also avoids the second check running when the header itself is absent, which would give a misleading extra error.

internal/controller/test/e2e/flowcontrol_authpolicy_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	if !strings.Contains(string(bodyA), nsA) {
+		t.Errorf("tenant A objective should contain namespace %q", nsA)
+	}
+
+	respB, bodyB := batchEnv.gatewayGet(t, path, tokenB, nil)
+	if respB.StatusCode != http.StatusOK {
+		t.Fatalf("tenant B: expected 200, got %d", respB.StatusCode)
+	}
+	if !strings.Contains(string(bodyB), nsB) {
```
mwaykole: **💡 Suggestion: add a negative isolation check on `bodyA` for tenant B's namespace.**

The current cross-namespace test only verifies that `bodyB` contains `nsB`. It doesn't verify that `bodyA` does NOT contain `nsB` as the objective value — i.e., that tenant A's objective was correctly set to `nsA` (not `nsB`). Adding this explicit negative check would strengthen the isolation guarantee:

```go
// Confirm tenant A does not get tenant B's namespace as its objective.
if strings.Contains(string(bodyA), `"x-gateway-inference-objective":"` + nsB + `"`) {
    t.Errorf("tenant A objective should NOT be nsB %q", nsB)
}
```
This catches cross-contamination bugs where the CEL expression might return the wrong namespace.

mwaykole (changes requested):
## PR Risk Assessment

**PR**: https://github.com/opendatahub-io/odh-model-controller/pull/854
**Author**: RishabhSaini
**Changed Files**: 1 (`internal/controller/test/e2e/flowcontrol_authpolicy_test.go`)
**Overall Risk**: 🟡 Medium — test-only change, no production code touched; however there are **two high-severity false-positive bugs** that would make the test pass even when the AuthPolicy is NOT correctly injecting headers.

### Summary

The PR adds two e2e tests verifying that the user-defined `AuthPolicy` injects `x-gateway-inference-fairness-id` and `x-gateway-inference-objective` headers on inference paths. The structure and setup pattern follows `batch_authpolicy_test.go` well — the test orchestration, SA lifecycle, RBAC grants, and retry logic are all correct.

However, I found **two critical false-positive assertions** (lines 45 and 75) where the test would pass even if the header value is **never injected**. The echo server (`ealen/echo-server`) reflects the entire request — including the URL path — back in its JSON response body. Since the request path is `/{ns}/echo-server/v1/chat/completions`, the namespace string `ns` is always present in the response body, regardless of whether the `x-gateway-inference-objective` header was actually set. This means these assertions provide **no actual signal** about the header value.

There is also a hardcoded cluster issuer assumption and a missing readiness wait for tenant B in the cross-namespace test.

**Recommendation**: REQUEST_CHANGES on the false-positive assertions — these need to be fixed for the tests to be meaningful. The other issues are improvements to address alongside.

mwaykole:
@RishabhSaini can u please check the failures

PR #858, opened 2026-06-25
fix: update endpoint in docs

pierDipi:
Correct endpoint is `/api/v1/samples/llm-d`

https://github.com/opendatahub-io/odh-model-controller/blob/e047cd303f806d4a05e0f3577203a0279f28b4d3/server/server.go#L27


bartoszmajsak (approved):
\o/

PR #860, opened 2026-07-02
[RHOAIENG-63480] Add canary traffic splitting via Route alternateBackends

maskarb:
## Summary
- Add canary traffic splitting for OpenShift Routes via `alternateBackends`
- Validate canary count <= 3 (OpenShift Route limit)
- Update Route comparator to detect `AlternateBackends` changes

## Changes
- **`kserve_raw_route_reconciler.go`**: Read `isvc.Spec.Canary`, set stable weight, add `alternateBackends` per canary service, resolve numeric `targetPort` for cross-backend compatibility
- **`route_comparator.go`**: Include `AlternateBackends` in diff detection
- **`inferenceservice_webhook.go`**: Reject > 3 canaries (Route alternateBackends limit)

## How it works
When canaries are present, the Route gets:
- Stable backend weight = `100 - sum(canary weights)`
- One `alternateBackend` per canary with its `trafficPercent` as weight
- Numeric `targetPort` (resolved from stable service) instead of named port, since service port names differ between stable and canary

When canaries are removed, the Route reverts to weight 100 on stable with no alternateBackends.

## Dependencies
- kserve/kserve#5672 (CanarySpec API types + reconciler)

## Test plan
- [x] Manual testing on CRC: 50/50 traffic split verified via pod access logs (100 requests → 50 stable, 50 canary)
- [x] Existing reconciler tests pass (43/43)
- [ ] Unit tests for canary route logic
- [ ] E2E tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added canary traffic splitting for raw InferenceService routes, including weighted stable and canary backends.
  * Added validation to limit the number of configured canary revisions.
  * InferenceService updates are now validated by the admission webhook.

* **Bug Fixes**
  * Route comparisons now detect changes to alternate backends.
  * Canary traffic percentages are validated to prevent totals exceeding 100%.

* **Chores**
  * Updated Kubernetes and related library versions.


mwaykole (commented):
Test review - checking permissions

internal/webhook/serving/v1beta1/inferenceservice_webhook.go
```diff
@@ -104,6 +104,11 @@ func (v *InferenceServiceCustomValidator) ValidateCreate(ctx context.Context, ob
 	}
 
 	logger.Info("Namespace is not protected")
+
+	if err := validateCanaryCount(inferenceservice); err != nil {
+		return nil, err
```
mwaykole, line 109: **🔴 Bug: Webhook will never intercept UPDATE operations — `ValidateUpdate` canary validation is a no-op in production.**

The `+kubebuilder:webhook` marker at line 62 of this file has `verbs=create` only. When `controller-gen` regenerates the `ValidatingWebhookConfiguration`, the API server will route only CREATE admission requests to this webhook. UPDATE requests bypass it entirely, so a user can exceed the 3-canary limit by:
1. Creating an ISVC with ≤3 canaries (webhook fires, passes)
2. Patching it to add more canaries (webhook is never called, no rejection)

Fix — add `update` to the marker at line 62:
```go
// +kubebuilder:webhook:path=/validate-serving-kserve-io-v1beta1-inferenceservice,mutating=false,failurePolicy=fail,sideEffects=None,groups=serving.kserve.io,resources=inferenceservices,verbs=create;update,...
```
This is the same pattern already used for the mutating webhook (`verbs=create;update`).

internal/webhook/serving/v1beta1/inferenceservice_webhook.go
```diff
@@ -137,6 +144,19 @@ func (v *InferenceServiceCustomValidator) ValidateDelete(ctx context.Context, ob
 	return nil, nil
 }
 
+// maxCanaryCount is the maximum number of canary deployments allowed per
+// InferenceService. OpenShift Routes support up to 3 alternateBackends
+// (4 total backends including the primary stable service).
+const maxCanaryCount = 3
+
```
mwaykole, line 149: **🔴 Missing validation: `sum(canary.TrafficPercent)` must be ≤ 100.**

`validateCanaryCount` checks the *number* of canaries (≤3) but not whether their weights are sane. If a user sets two canaries at 60% + 60%, `stableWeight` in the reconciler becomes `100 - 120 = -20`. OpenShift Route weights must be in `[0, 256]` — a negative weight causes an API server rejection, and the reconciler enters a permanent error loop on every reconcile tick.

Suggested addition to this function:
```go
var total int32
for _, c := range isvc.Spec.Canary {
    total += c.TrafficPercent
}
if total > 100 {
    return fmt.Errorf("total canary traffic percent %d exceeds 100; stable backend weight would be negative", total)
}
```

internal/controller/serving/reconcilers/kserve_raw_route_reconciler.go
```diff
@@ -181,6 +181,26 @@ func (r *KserveRawRouteReconciler) createDesiredResource(ctx context.Context, lo
 		return nil, err
 	}
 
+	if len(isvc.Spec.Canary) > 0 {
+		// Use numeric targetPort when canary is active — named ports differ
+		// per service (e.g. "canary-test-predictor" vs "canary-test-v2-predictor")
+		// and alternateBackends requires a port name that exists on all backends.
+		// Resolve the named port chosen by setRouteTargetPort to its numeric value.
+		if targetPort.Type == intstr.String {
+			for _, p := range targetService.Spec.Ports {
```
mwaykole, line 191: **🟡 Silent failure: if the named port is not found in `targetService.Spec.Ports`, the loop completes without updating `desiredRoute.Spec.Port`.**

The route is left with the *string* `targetPort` (e.g. `"http"`). The comment above correctly explains that named ports are incompatible across canary services — but if the resolution silently does nothing, the route still carries a named port and the OpenShift Router will try to resolve `"http"` on each alternateBackend independently. Canary services (e.g. `"myisvc-v2-predictor"`) may have no port named `"http"`, causing routing failures for a fraction of requests with no error visible in reconciler logs.

Replace the silent fallthrough with an explicit error:
```go
if targetPort.Type == intstr.String {
    found := false
    for _, p := range targetService.Spec.Ports {
        if p.Name == targetPort.StrVal {
            desiredRoute.Spec.Port = &v1.RoutePort{TargetPort: p.TargetPort}
            found = true
            break
        }
    }
    if !found {
        return nil, fmt.Errorf("port %q not found on service %q; cannot resolve numeric targetPort for canary route",
            targetPort.StrVal, targetService.Name)
    }
}
```

internal/controller/serving/reconcilers/kserve_raw_route_reconciler.go
```diff
@@ -280,3 +300,33 @@ func (r *KserveRawRouteReconciler) processDelta(ctx context.Context, log logr.Lo
 	}
 	return nil
 }
+
+func (r *KserveRawRouteReconciler) applyCanaryTrafficSplits(ctx context.Context, log logr.Logger, isvc *kservev1beta1.InferenceService, route *v1.Route) error {
+	var totalCanaryPercent int32
+	for _, canary := range isvc.Spec.Canary {
+		totalCanaryPercent += canary.TrafficPercent
+	}
+	stableWeight := int32(100) - totalCanaryPercent
```
mwaykole, line 313: **🔴 Defensive guard missing: `stableWeight` can be negative if webhook validation is bypassed (or not wired for UPDATE).**

When `totalCanaryPercent > 100`, `stableWeight = int32(100) - totalCanaryPercent` is negative. `r.client.Create/Update` will then fail with an OpenShift API validation error, causing the reconciler to log an error and requeue indefinitely.

Add a guard after computing `stableWeight`:
```go
stableWeight := int32(100) - totalCanaryPercent
if stableWeight < 0 {
    return fmt.Errorf("canary traffic percents sum to %d which exceeds 100; stable backend weight would be %d",
        totalCanaryPercent, stableWeight)
}
route.Spec.To.Weight = &stableWeight
```

internal/controller/serving/reconcilers/kserve_raw_route_reconciler.go
```diff
@@ -280,3 +300,33 @@ func (r *KserveRawRouteReconciler) processDelta(ctx context.Context, log logr.Lo
...
+	route.Spec.To.Weight = &stableWeight
+
+	var alternateBackends []v1.RouteTargetReference
+	for _, canary := range isvc.Spec.Canary {
+		canaryServiceName := fmt.Sprintf("%s-%s-predictor", isvc.Name, canary.Predictor.Name)
+
+		svc := &corev1.Service{}
```
mwaykole, line 327: **🟡 Clarity: `svc` is fetched but its fields are never read after the existence check.**

This is functionally correct — OpenShift Routes apply `route.Spec.Port` uniformly to all backends, so the canary service's port doesn't need to be inspected per-backend. However, future readers will reasonably wonder why `svc` was fetched and populated. Add a comment explaining the intent:
```go
// Verify the canary Service exists before registering it as an alternateBackend.
// Port resolution is not needed here — route.Spec.Port (resolved to a numeric value
// above) is applied uniformly by the OpenShift Router across all backends.
svc := &corev1.Service{}
if err := r.client.Get(ctx, ..., svc); err != nil { ... }
```

internal/controller/comparators/route_comparator.go
```diff
@@ -31,6 +31,7 @@ func GetKServeRouteComparator() ResourceComparator {
 			reflect.DeepEqual(deployedRoute.Spec.Port, requestedRoute.Spec.Port) &&
 			reflect.DeepEqual(deployedRoute.Spec.TLS, requestedRoute.Spec.TLS) &&
 			reflect.DeepEqual(deployedRoute.Spec.WildcardPolicy, requestedRoute.Spec.WildcardPolicy) &&
+			reflect.DeepEqual(deployedRoute.Spec.AlternateBackends, requestedRoute.Spec.AlternateBackends) &&
 			reflect.DeepEqual(deployedRoute.ObjectMeta.Labels, requestedRoute.ObjectMeta.Labels) &&
```
mwaykole, line 35: ✅ Correct and necessary addition — without this, a canary weight change (e.g. `50%→30%`) wouldn't trigger a Route update delta, leaving the live Route out of sync. Note that `Spec.To` on the line above already covers the primary backend's `Weight` pointer via `reflect.DeepEqual`, so there's no duplication concern between the two fields.

mwaykole (changes requested):
## PR Risk Assessment — [RHOAIENG-63480] Canary Traffic Splitting via Route `alternateBackends`

**Overall Risk: Medium-High** — The feature direction is correct and the approach of using OpenShift Route `alternateBackends` is sound, but there are two real bugs: (1) a webhook misconfiguration meaning canary-count validation silently never runs on UPDATE, and (2) a missing traffic-sum guard that can produce a negative stable weight, causing the reconciler to error-loop. There is also a silent failure path in port resolution that will cause hard-to-debug partial routing breakage.

The PR is also self-labeled `do-not-merge/work-in-progress` with unit tests explicitly unchecked in the test plan — those need to land before merge.

### Summary of Findings
| # | Severity | Location | Issue |
|---|----------|----------|---------|
| 1 | 🔴 Bug | `inferenceservice_webhook.go` | Webhook marker only has `verbs=create` — `ValidateUpdate` canary validation never fires |
| 2 | 🔴 Bug | `kserve_raw_route_reconciler.go` | No guard that sum(canary weights) ≤ 100; `stableWeight` can go negative → OCP Route API rejects it |
| 3 | 🟡 Logic | `kserve_raw_route_reconciler.go` | Silent fallthrough when named port not found — no error, no log, route silently keeps wrong port |
| 4 | 🟡 Clarity | `kserve_raw_route_reconciler.go` | Canary `svc` is fetched but its port data is never read — misleading to future readers |
| 5 | 🟡 Tests | reconciler test file | Zero test coverage for the new `applyCanaryTrafficSplits` logic (explicitly unchecked in PR description) |

Great start overall — the port-name-to-numeric-port resolution rationale is well-documented in comments, the comparator update is correct, and the 50/50 manual test verification is appreciated.

internal/webhook/serving/v1beta1/inferenceservice_webhook.go
maskarb: This is a completely valid concern, however, this is handled by the kserve validation:
https://github.com/kserve/kserve/pull/5672/changes#diff-e3dceb9e78041cab47196b7b8b38332a61106f4d66009f89cd5ed793dec8377bR249-R251

internal/controller/serving/reconcilers/kserve_raw_route_reconciler.go
maskarb: Guard was added.

internal/controller/serving/reconcilers/kserve_raw_route_reconciler.go
maskarb: Replaced the fall through with an error.

internal/webhook/serving/v1beta1/inferenceservice_webhook.go
maskarb: Added the `update` verb.

internal/controller/serving/reconcilers/kserve_raw_route_reconciler.go
maskarb: A clarifying comment was added.

Jooho (approved):

PR #862, opened 2026-07-06
Add default ClusterServingRuntime manifests for Phase 1

rpancham:
## Summary

Adds 3 default ClusterServingRuntime manifests to enable CLI/GitOps model deployment workflows without Dashboard dependency.

Part of Phase 1 CSR migration to decouple serving runtime management from OpenShift Templates.

  ## Changes

  **New Files:**
  - `config/runtimes/csr-kserve-mlserver-v1.yaml` - MLServer (sklearn, xgboost, lightgbm, onnx)
  - `config/runtimes/csr-kserve-ovms-v1.yaml` - OpenVINO Model Server
  - `config/runtimes/csr-kserve-vllm-v1.yaml` - vLLM for LLMs

  **Modified:**
  - `config/runtimes/kustomization.yaml` - Added CSR resources

  ## Key Features

  - Cluster-scoped runtimes available to all namespaces
  - Prefix-based naming (`csr-kserve-*-v1`) prevents conflicts
  - Version tracking via `opendatahub.io/version` annotations
  - Backward compatible - existing Template workflow unchanged

  ## Related

  - JIRA: [RHOAIENG-71868](https://redhat.atlassian.net/browse/RHOAIENG-71868)
  - [Phase 1 Backend Plan](https://docs.google.com/document/d/16exEcAbTXI7c0ylJmqFZmPLUTdXMiNMzefs1Eljra6M/edit)

## Summary by CodeRabbit

* **New Features**
  * Added model-serving runtime templates for MLServer, OpenVINO Model Server (OVMS), and vLLM.
  * MLServer supports scikit-learn, XGBoost, LightGBM, and ONNX models with health checks, metrics, and hardened non-root execution.
  * OVMS supports configured model-serving protocols and exposes a metrics port.
  * vLLM provides an OpenAI-compatible API with Prometheus metrics and GPU acceleration support.
  * All three runtimes are included in the default runtime configuration.

[RHOAIENG-71868]: https://redhat.atlassian.net/browse/RHOAIENG-71868?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/runtimes/csr-kserve-mlserver-v1.yaml
+++ b/config/runtimes/csr-kserve-mlserver-v1.yaml
@@ -1,0 +1,82 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: csr-kserve-mlserver-v1
+  labels:
+    opendatahub.io/managed: "true"
+  annotations:
+    opendatahub.io/version: "1.0.0"
+    opendatahub.io/template-display-name: "MLServer"
+    opendatahub.io/runtime-version: "1.7.1"
+    openshift.io/display-name: MLServer ServingRuntime for KServe
+    serving.kserve.io/server-type: mlserver
+spec:
+  annotations:
+    opendatahub.io/kserve-runtime: "mlserver"
+    prometheus.io/port: "8082"
+    prometheus.io/path: /metrics
+    monitoring.opendatahub.io/scrape: "true"
+  containers:
+    - name: kserve-container
+      image: quay.io/opendatahub/mlserver:fast
+      env:
+        - name: MLSERVER_MODEL_IMPLEMENTATION
+          value: "{{.Labels.modelClass}}"
+        - name: MLSERVER_HTTP_PORT
+          value: "8080"
+        - name: MLSERVER_MODELS_DIR
+          value: /mnt/models
+      startupProbe:
+        initialDelaySeconds: 1
+        periodSeconds: 1
+        failureThreshold: 1
+        exec:
+          command:
+            - /bin/sh
+            - -c
+            - |
+              [ -n "$(ls -A /mnt/models 2>/dev/null)" ]
+      readinessProbe:
+        initialDelaySeconds: 5
+        periodSeconds: 5
+        failureThreshold: 3
+        timeoutSeconds: 5
+        httpGet:
+          path: /v2/models/{{.Name}}/ready
+          port: 8080
+      livenessProbe:
+        initialDelaySeconds: 20
+        periodSeconds: 10
+        failureThreshold: 6
+        timeoutSeconds: 5
+        httpGet:
+          path: /v2/models/{{.Name}}/ready
+          port: 8080
+      ports:
+        - containerPort: 8080
+          protocol: TCP
+      securityContext:
+        allowPrivilegeEscalation: false
+        capabilities:
+          drop:
+            - ALL
+        privileged: false
+        runAsNonRoot: true
+  protocolVersions:
+    - v2
+  multiModel: false
+  supportedModelFormats:
+    - name: sklearn
+      version: "0"
+    - name: sklearn
+      version: "1"
+    - name: xgboost
+      version: "1"
+    - name: xgboost
+      version: "2"
+    - name: lightgbm
+      version: "3"
+    - name: lightgbm
+      version: "4"
+    - name: onnx
+      version: "1"
--- a/config/runtimes/csr-kserve-mlserver-v1.yaml
+++ b/config/runtimes/csr-kserve-mlserver-v1.yaml
@@ -18,7 +18,7 @@ spec:
     monitoring.opendatahub.io/scrape: "true"
   containers:
     - name: kserve-container
-      image: quay.io/opendatahub/mlserver:fast
+      image: $(mlserver-image)
       env:
         - name: MLSERVER_MODEL_IMPLEMENTATION
           value: "{{.Labels.modelClass}}"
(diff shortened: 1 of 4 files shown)


rpancham:
These ClusterServingRuntime manifests are direct spec-for-spec conversions of the existing Template-wrapped ServingRuntimes in this repo. Every issue flagged here also applies to the source Templates already shipping in production.
Image tags (mlserver:fast, vllm-cuda:latest): Same pattern as the Templates — they use kustomize variable placeholders ($(mlserver-image), $(vllm-cuda-image)) resolved from config/base/params.env. These tags are build-time defaults that get replaced the same way.
No resource limits (all 3 runtimes):Existing Templates don't set resource limits either. Resource requests/limits are specified at InferenceService level by users when deploying models, not in the ServingRuntime spec.

No securityContext on OVMS and vLLM:
The source Templates (ovms-kserve-template.yaml, vllm-cuda-template.yaml) also have no securityContext. MLServer's CSR includes it because the MLServer Template has it. CSRs intentionally match their source Templates.
 
readOnlyRootFilesystem:
Not set in any existing Templates. For vLLM specifically, it writes to HF_HOME=/tmp/hf_home, so enabling this would break the runtime without additional volume mounts.

GPU resource requests (vLLM):
Existing vLLM Template doesn't request GPU resources either. GPU allocation is handled at InferenceService deployment time.
 Security hardening improvements are valid follow-up work but should cover both the existing Templates and the new CSRs together in a separate PR.

brettmthompson (approved):

config/runtimes/csr-kserve-mlserver-v1.yaml
```diff
@@ -0,0 +1,82 @@
...
+      ports:
+        - containerPort: 8080
+          protocol: TCP
+      securityContext:
+        allowPrivilegeEscalation: false
+        capabilities:
+          drop:
+            - ALL
+        privileged: false
+        runAsNonRoot: true
```
mwaykole, line 71: This is missing the `resources` block from the source template (`mlserver-template.yaml` has requests cpu:1/memory:2Gi and limits cpu:2/memory:4Gi). The reply that Templates don't set limits isn't true for MLServer — please keep CSR parity with the Template here.

config/runtimes/csr-kserve-mlserver-v1.yaml
```diff
@@ -0,0 +1,82 @@
...
+    serving.kserve.io/server-type: mlserver
+spec:
+  annotations:
+    opendatahub.io/kserve-runtime: "mlserver"
+    prometheus.io/port: "8082"
+    prometheus.io/path: /metrics
+    monitoring.opendatahub.io/scrape: "true"
+  containers:
+    - name: kserve-container
+      image: quay.io/opendatahub/mlserver:fast
```
mwaykole: Templates use `$(mlserver-image)` / `$(ovms-image)` / `$(vllm-cuda-image)` and get overwritten from `params.env` via kustomize replacements in `config/base/kustomization.yaml` (selecting `kind: Template`). These CSRs hardcode the default tags and have no matching `ClusterServingRuntime` replacement targets, so release digest pins / RELATED_IMAGE updates will land on Templates but not on these CSRs.

Please either:
1. Use a placeholder matching the Template pattern and add replacement entries for `csr-kserve-mlserver-v1` / `csr-kserve-ovms-v1` / `csr-kserve-vllm-v1`, or
2. Confirm in this PR that image rewriting for CSRs is handled elsewhere (and link where).

Same concern for the OVMS and vLLM CSR images.

mwaykole (commented):
Looked at these against the source Templates. Spec parity is mostly good (OVMS/vLLM match; CodeRabbit hardening asks are follow-ups that apply to Templates too).

Two things that should land before merge — left inline:
1. MLServer CSR dropped `resources` that exist on `mlserver-template.yaml`
2. CSR images are hardcoded and not wired into the `params.env` kustomize replacements that Templates use

mwaykole (approved):

Snomaan6846:
Missing kustomize replacement targets for ClusterServingRuntime image refs

The CSR manifests correctly use kustomize variable placeholders for their images:

- csr-kserve-mlserver-v1.yaml → $(mlserver-image)
- csr-kserve-ovms-v1.yaml → $(ovms-image)
- csr-kserve-vllm-v1.yaml → $(vllm-cuda-image)

However, `config/base/kustomization.yaml` has no replacements targets for `kind: ClusterServingRuntime`. All existing replacement targets only select `kind: Template`. This means kustomize will leave the literal strings `$(mlserver-image), $(ovms-image), and $(vllm-cuda-image)` in the rendered output — these are not valid image references and will cause ImagePullBackOff at runtime.

The following replacement targets need to be added to config/base/kustomization.yaml (appended to the existing source blocks for each image):
```
# Add to the existing data.ovms-image source block's targets:
      - select:
          kind: ClusterServingRuntime
          name: csr-kserve-ovms-v1
        fieldPaths:
          - spec.containers.0.image
# Add to the existing data.mlserver-image source block's targets:
      - select:
          kind: ClusterServingRuntime
          name: csr-kserve-mlserver-v1
        fieldPaths:
          - spec.containers.0.image
# Add to the existing data.vllm-cuda-image source block's targets:
      - select:
          kind: ClusterServingRuntime
          name: csr-kserve-vllm-v1
        fieldPaths:
          - spec.containers.0.image
```
Note the fieldPaths difference: Templates use `objects.0.spec.containers.0.image` (ServingRuntime nested inside the Template's objects array), while ClusterServingRuntimes are top-level resources and need `spec.containers.0.image`.

PR #871, opened 2026-07-09
Create Gateway PodMonitor to scrape Istio/Envoy inference metrics

pierDipi:
Add a PodMonitor for gateway pods to collect Istio traffic metrics (request rate, latency, request/response sizes) used by LLM inference dashboards. The monitor selects pods by the standard Gateway API label and uses relabeling to tag metrics with the gateway name and an llm_isvc_gateway sentinel for dashboard filtering.

The metrics endpoint (port 15020 "metrics") is always plain HTTP regardless of whether the gateway listeners use TLS or not, so the scheme is set explicitly to avoid issues with cluster-level monitoring defaults.

## Summary by CodeRabbit

* **New Features**
  * Gateways can now automatically create and reconcile a Prometheus `PodMonitor` to expose metrics when observability is enabled and the gateway is actively referenced.
  * The `PodMonitor` is removed when observability is disabled or the gateway is no longer referenced, and it is recreated if deleted.
  * Added support for the Prometheus `PodMonitor` CustomResourceDefinition.
* **Tests**
  * Expanded controller test coverage for `PodMonitor` lifecycle and ownership behavior.

### DIFF
--- a/internal/controller/serving/llm/fixture/podmonitor_helpers.go
+++ b/internal/controller/serving/llm/fixture/podmonitor_helpers.go
@@ -1,0 +1,50 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fixture
+
+import (
+	"context"
+
+	"github.com/onsi/gomega"
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/opendatahub-io/odh-model-controller/internal/controller/constants"
+)
+
+func VerifyGatewayPodMonitorExists(ctx context.Context, c client.Client, gatewayNamespace, gatewayName string) {
+	VerifyResourceExists(ctx, c, gatewayNamespace, constants.GetGatewayPodMonitorName(gatewayName), &monitoringv1.PodMonitor{})
+}
+
+func VerifyGatewayPodMonitorNotExist(ctx context.Context, c client.Client, gatewayNamespace, gatewayName string) {
+	VerifyResourceNotExist(ctx, c, gatewayNamespace, constants.GetGatewayPodMonitorName(gatewayName), &monitoringv1.PodMonitor{})
+}
+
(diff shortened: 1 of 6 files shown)


internal/controller/serving/llm/fixture/podmonitor_helpers.go
```diff
@@ -0,0 +1,50 @@
...
+func VerifyGatewayPodMonitorExists(ctx context.Context, c client.Client, gatewayNamespace, gatewayName string) {
+	VerifyResourceExists(ctx, c, gatewayNamespace, constants.GetGatewayPodMonitorName(gatewayName), &monitoringv1.PodMonitor{})
+}
+
+func VerifyGatewayPodMonitorNotExist(ctx context.Context, c client.Client, gatewayNamespace, gatewayName string) {
+	VerifyResourceNotExist(ctx, c, gatewayNamespace, constants.GetGatewayPodMonitorName(gatewayName), &monitoringv1.PodMonitor{})
+}
+
+func VerifyGatewayPodMonitorOwnerRef(ctx context.Context, c client.Client, gatewayNamespace, gatewayName string) {
+	gomega.Eventually(func() error {
```
bartoszmajsak: We should use scoped `func(g Gomega)`  for eventually.

ref https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode

internal/controller/serving/llm/gateway_controller.go
```diff
@@ -321,6 +335,142 @@ func (r *GatewayReconciler) deleteAuthPolicyIfManaged(ctx context.Context, logge
...
+		logger.Info("Creating PodMonitor", "name", podMonitorName)
+		if err := controllerutil.SetControllerReference(gateway, desired, r.Scheme); err != nil {
+			return fmt.Errorf("failed to set controller reference: %w", err)
+		}
+		if err := r.Client.Create(ctx, desired); err != nil {
+			return fmt.Errorf("failed to create PodMonitor: %w", err)
+		}
+	} else if delta.IsUpdated() {
+		logger.Info("Updating PodMonitor", "name", podMonitorName)
+		if err := controllerutil.SetControllerReference(gateway, desired, r.Scheme); err != nil {
```
bartoszmajsak: using `desired` here and then `rp` is a bit confusing. Maybe remove it entirely, as it's already set while creating, or use `rp` to enforce?

internal/controller/serving/llm/gateway_controller.go
```diff
@@ -139,6 +141,18 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		}
 	}
 
+	if scrape := gateway.GetLabels()[constants.RhoaiObservabilityLabel]; referencedByLLMService && scrape != "false" {
```
bartoszmajsak, line 144: Should we add this label as predicate to trigger reconcile?

internal/controller/serving/llm/fixture/podmonitor_helpers.go
pierDipi: I wish the API wouldn't allow it :)

bartoszmajsak (approved):
/lgtm

PR #872, opened 2026-07-13
[RHOAIENG-69082] Add xKS webhook-only mode for OMC

mholder6:
## Description

Adds xKS compatibility for odh-model-controller ([RHOAIENG-69082](https://redhat.atlassian.net/browse/RHOAIENG-69082)).

When `ODH_PLATFORM_TYPE=xKS`, OMC runs a webhook-only profile:
- **Active:** LLMISVC mutating webhooks (v1alpha1 + v1alpha2) for ConnectionsAPI and HardwareProfile injection
- **Disabled:** All reconcilers, NIM setup, MaaS RBAC cleanup, validating admission webhooks, and ISVC webhooks

Also adds `config/overlays/xks/` — a slim deployment overlay with minimal RBAC, two LLMISVC mutating webhooks, and `ODH_PLATFORM_TYPE=xKS`.

Production xKS install still requires Helm/build-config wiring (`RELATED_IMAGE_*` params in `RHOAI-Build-Config/helm/xks-values-patch.yaml`) to be registered separately.

JIRA: https://redhat.atlassian.net/browse/RHOAIENG-69082

## How Has This Been Tested?

- `go build ./...`
- `go test ./cmd/... ./internal/platform/...`
- `kubectl kustomize config/overlays/xks` (overlay builds cleanly)
- Kind smoke test (local): verified webhook-only startup, 2 LLMISVC mutating webhooks, no validating webhooks, LLMISVC create admitted

## Merge criteria:

[RHOAIENG-69082]: https://redhat.atlassian.net/browse/RHOAIENG-69082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added support for deploying the controller in xKS environments.
  * xKS deployments now provide LLM inference service webhooks while omitting unsupported serving, metrics, validation, and platform-specific resources.
  * Added an optimized xKS configuration with appropriate environment settings and resource limits.

* **Tests**
  * Added automated KinD smoke testing to verify deployment readiness, webhook behavior, and LLM inference service creation.
  * Added coverage for platform detection and webhook registration modes.


pierDipi (commented):
@mholder6 can you add test workflow on top of or similar to https://github.com/opendatahub-io/kserve/blob/master/.github/workflows/e2e-test-odh-xks-kind.yml?

mholder6:
> @mholder6 can you add test workflow on top of or similar to https://github.com/opendatahub-io/kserve/blob/master/.github/workflows/e2e-test-odh-xks-kind.yml?

@pierDipi Added an xKS Kind smoke test workflow (similar to KServe’s). It deploys the xKS overlay on Kind and checks webhook-only mode: 3 mutating webhooks, no validating webhooks, and ISVC create works.

config/overlays/xks/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: opendatahub
+
+resources:
+- ../../manager
+- webhook
+- rbac
```
pierDipi: can we follow the same pattern as Kserve where the the xks overlay starts on top of the odh overlay (in this case I think it's "default") and then we remove what's not necessary for xks? it will allow re-using as much as possible without duplicating resources multiple times, including image injection, etc

mholder6: Addressed: config/overlays/xks now inherits from config/base and removes xKS-unneeded resources (runtimes, model-serving-api, prometheus, NIM, validating webhooks), trims mutating webhooks to 3, and reuses base image injection. Separate overlay deployed via kustomize (not Helm); Kind smoke test in CI validates it.

cmd/main.go
```diff
@@ -305,6 +304,41 @@ func setupWebhooks(mgr ctrl.Manager, setupLog logr.Logger) error {
...
+func webhookSetupsForMode(xksMode bool) []struct {
+	name    string
+	setupFn func(ctrl.Manager) error
+} {
+	if xksMode {
+		return []struct {
+			name    string
+			setupFn func(ctrl.Manager) error
+		}{
+			{"InferenceService", webhookservingv1beta1.SetupInferenceServiceMutatingWebhookWithManager},
```
pierDipi: this is not needed on xKS (and related manifests)

mholder6: Dropped ISVC webhook registration, overlay entry, and smoke test

internal/platform/platform.go
```diff
@@ -0,0 +1,15 @@
...
+
+const (
+	// PlatformTypeEnv is the environment variable used to select the deployment platform.
+	PlatformTypeEnv = "ODH_PLATFORM_TYPE"
+	// XKS is the platform type for vanilla Kubernetes (extended KServe) deployments.
+	XKS = "xKS"
+)
+
+// IsXKS reports whether the controller should run in xKS webhook-only mode.
+func IsXKS() bool {
```
israel-hdez, line 13: nit: I think it would had been better to place in the existing `internal/controller/utils.go` file

config/overlays/xks/kustomization.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- ../../base
+- namespace.yaml
```
israel-hdez: The resultant manifest is so different, that IMO it is better to do the overlay "from scratch". It may be simpler to not rely on `base` and, instead, include the specific folders we need.

As it it, it removes almost everything but the deployment, some RBAC, and the webhook configurations.

config/overlays/xks/patches/mutating-webhooks-xks.yaml
```diff
@@ -0,0 +1,44 @@
+# Keep only LLMISVC ConnectionsAPI and HardwareProfile mutating webhooks for xKS.
+- op: replace
+  path: /webhooks
+  value:
```
israel-hdez, line 4: This is almost like re-creating the resource. Maybe, we should have a dedicated one for xKS rather than patching?

config/overlays/xks/kustomization.yaml
israel-hdez: @pierDipi This is basically the opposite of what you suggested.

Maybe, to reuse existing YAML, we could include only what we need? Some thing like:
```
resources:
- ../rbac
- ../manger
- ../webhook
- (etc)
```

and add patches to refine? Just to balance the amount of reusing vs new YAML. As it is, little is being reused, and this is adding quite a lot of `patch: $delete`.

test/scripts/kind/setup-odh-model-controller-xks.sh
```diff
@@ -0,0 +1,109 @@
...
+  name: xks-smoke-llmisvc
+  namespace: ${NAMESPACE}
+spec:
+  model:
+    name: test-model
+    uri: s3://test-bucket/test-model
+EOF
+
+kubectl -n "${NAMESPACE}" get llminferenceservice xks-smoke-llmisvc >/dev/null 2>&1 \
+  || fail "LLMInferenceService was not created; webhook admission may have failed"
```
israel-hdez, line 123: To ensure the webhook is running, I'd use a negative test:

Create an LLMIsvc with ConnectionAPI annotations referring to a non-existing connection:
```yaml
metadata:
  annotations:
    opendatahub.io/connections: dummy
```

The webhook should reject creation.

Similarly for HWP:
```yaml
metadata:
  annotations:
    opendatahub.io/hardware-profile-name: dummy
```

The webhook should reject creation.

That would ensure the webhooks are active, while the current smoke test I don't think proves much.

config/overlays/xks/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: opendatahub
```
pierDipi: why do we have this namespace? the namespace is still created by the operator

config/overlays/xks/patches/remove-runtimes.yaml
pierDipi, line 1: is there a way of removing all Template without listing them all? 

maybe something like this in the kustomization.yaml?
```yaml
patches:
  - target:
      kind: Template
    patch: |-
      $patch: delete
      apiVersion: template.openshift.io/v1
      kind: Template
      metadata:
        name: ".*" # 2. Dummy name; the target selector overrides this
```

config/overlays/xks/kustomization.yaml
pierDipi: my only problem is repeating the params.env injection

pierDipi: I also initially commented because we were copying resources altogether

pierDipi: Now that we control module operator, perhaps there could be even other/better solutions

config/overlays/xks/kustomization.yaml
```diff
@@ -0,0 +1,48 @@
...
+  patch: |-
+    - op: remove
+      path: /metadata/annotations/service.beta.openshift.io~1serving-cert-secret-name
+- target:
+    group: admissionregistration.k8s.io
+    kind: MutatingWebhookConfiguration
+    name: mutating.odh-model-controller.opendatahub.io
+  patch: |-
+    - op: remove
+      path: /metadata/annotations/service.beta.openshift.io~1inject-cabundle
```
pierDipi, line 141: in Kserve, I think we're injecting cert-manager annotations to get webhooks to work over TLS, can you double check that?

config/overlays/xks/patches/remove-runtimes.yaml
pierDipi: This is applicable to config/overlays/xks/patches/remove-prometheus.yaml too

config/overlays/xks/kustomization.yaml
israel-hdez: > I also initially commented because we were copying resources altogether

:+1: Yes, I wouldn't recommend replicating.

> my only problem is repeating the params.env injection
> Now that we control module operator, perhaps there could be even other/better solutions

This anyways needs a complementary PR to pick the right overlay, given the platform. If we want to maintain a single `params.env` file, I think what we can do is the operator to copy the file either at runtime or at container build -- I'd prefer the later.

Thus, @mholder6 @Jooho be aware that complementary PR is needed on the modular-operator side.

pierDipi: I'm not too concerned about params.env as you said we can copy it, more like the transformation logic to get those params in the right place

internal/webhook/serving/v1beta1/inferenceservice_webhook.go
```diff
@@ -50,13 +50,17 @@ var inferenceservicelog = logf.Log.WithName("inferenceservice-resource")
 func SetupInferenceServiceWebhookWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewWebhookManagedBy(mgr).For(&servingv1beta1.InferenceService{}).
 		WithValidator(&InferenceServiceCustomValidator{client: mgr.GetClient()}).
-		WithDefaulter(&InferenceServiceCustomDefaulter{
-			client:    mgr.GetClient(),
-			apiReader: mgr.GetAPIReader(),
-		}).
+		WithDefaulter(newInferenceServiceCustomDefaulter(mgr)).
```
Jooho: why we need this extraction?

config/overlays/xks/patches/remove-nim.yaml
```diff
@@ -0,0 +1,19 @@
...
+kind: CustomResourceDefinition
+metadata:
+  name: accounts.nim.opendatahub.io
+---
+$patch: delete
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: account-editor-role
+---
```
Jooho: Should we need to remove openshift related role too? 

example)
```
  - template.openshift.io (Templates CRUD)
  - route.openshift.io (Routes CRUD)
  - config.openshift.io (Authentications)
```

config/overlays/xks/kustomization.yaml
```diff
@@ -0,0 +1,48 @@
...
+- ../../base
+- namespace.yaml
+
+configMapGenerator:
+- name: odh-model-controller-parameters
+  behavior: merge
+  envs:
+  - params.env
+
+patchesStrategicMerge:
```
Jooho: this is deprecated so please use patches

cmd/main.go
```diff
@@ -305,6 +304,40 @@ func setupWebhooks(mgr ctrl.Manager, setupLog logr.Logger) error {
...
+func webhookSetupNames(xksMode bool) []string {
+	setups := webhookSetupsForMode(xksMode)
+	names := make([]string, len(setups))
+	for i, setup := range setups {
+		names[i] = setup.name
+	}
+	return names
+}
+
+func webhookSetupsForMode(xksMode bool) []struct {
```
Jooho: The anonymous struct is repeated 3 times so extracting a webhookSetup type would clean this up.

config/overlays/xks/manager_xks_env_patch.yaml
```diff
@@ -0,0 +1,23 @@
...
+    spec:
+      containers:
+      - name: manager
+        env:
+        - name: ODH_PLATFORM_TYPE
+          value: xKS
+        - name: NIM_STATE
+          value: removed
+        - name: MODELREGISTRY_STATE
+          value: removed
```
Jooho: nit: NIM_STATE,  MODELREGISTRY_STATE are already set by the base replacements from params.env values so I think we can remove it

config/overlays/xks/kustomization.yaml
mholder6: @pierDipi Checked KServe odh-xks and you’re right, it uses cert-manager (cert-manager.io/inject-ca-from) instead of OpenShift serving certs. Our xKS overlay strips the OCP annotations but doesn’t add cert-manager yet; the Kind test patches CA manually. Planning to wire cert-manager in the module-operator / production xKS follow-up, same pattern as KServe. If we want it in this PR, I'll be happy to add it now

pierDipi: let's follow the same pattern as Kserve? if Kserve is not using module operator, let's not use module operator and make changes here, I prefer a raw manifest approach as it's easier to test and reason about without too many components and build involved to get a dev setup working

config/overlays/xks/patches/remove-nim.yaml
mholder6: yes you're right those permissions leak through from config/base. I'll add a patch to trim the OpenShift-specific rules from the xKS ClusterRole.

config/overlays/xks/patches/remove-openshift-rbac.yaml
```diff
@@ -0,0 +1,9 @@
+# Remove OpenShift-specific ClusterRole rules not needed on vanilla Kubernetes.
```
israel-hdez, line 1: Coderabbit comment (which weirdly is marked as resolved) is very, very valid. I think this is quite brittle.
We should think how to enhance it.

PR #875, opened 2026-07-13
fix: correct JSON field path for HardwareProfile scheduling spec lookup

israel-hdez:
## Description

Fix nodeSelector, toleration and Kueue queue name injection of HardwareProfile.
The resolver was reading `spec.schedulingSpec.*` from the unstructured CR
object, but the actual CRD field is `spec.scheduling.*`.

Fixes https://redhat.atlassian.net/browse/RHOAIENG-76305

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed HardwareProfile scheduling parsing to read the correct `scheduling` fields, including Kueue queue configuration, node selectors, and tolerations.
  * Ensures scheduling settings from the current HardwareProfile specification structure are applied when resolving profiles.
* **Tests**
  * Updated webhook and test-fixture payloads to use the corrected `scheduling` key layout for HardwareProfile scenarios.

### DIFF
--- a/internal/webhook/serving/hardwareprofile/resolver.go
+++ b/internal/webhook/serving/hardwareprofile/resolver.go
@@ -57,7 +57,7 @@ type ResolvedProfile struct {
 	NodeSelector map[string]string
 	// Tolerations to prepend to the workload spec.
 	Tolerations []corev1.Toleration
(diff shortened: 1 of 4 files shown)


VedantMahabaleshwarkar (approved):
/lgtm

PR #878, opened 2026-07-14
fix(webhook): reject InferenceService with connection Secret missing connection-type-protocol

israel-hdez:
## Description

When an InferenceService or LLMInferenceService referenced a connection
Secret that had no opendatahub.io/connection-type-protocol annotation,
the ConnectionsAPI webhook silently skipped injection and admitted the
resource unchanged. This caused models using S3 connections created via
YAML, GitOps, or tutorials (which typically omit the type annotation
from the Secret) to be admitted without spec.predictor.model.storage
populated, so the model would never load from S3.

Additionally, on UPDATE of such an already-broken resource would trigger an
unintended cleanup of fields that were never injected.

The webhook now returns an error in both cases.

Workaround for existing YAML-created Secrets:
  oc annotate secret <name> opendatahub.io/connection-type-protocol=s3

Fixes: https://redhat.atlassian.net/browse/RHOAIENG-69754

## How Has This Been Tested?

Running unit tests.

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Connection configuration now fails validation when a referenced Secret is missing connection-type metadata.
  * Unsupported connection types are reported as errors instead of being silently skipped.
  * Error messages identify the missing or invalid connection-type information to help resolve configuration issues.

### DIFF
--- a/internal/webhook/connectionapi/utils.go
+++ b/internal/webhook/connectionapi/utils.go
@@ -85,9 +85,9 @@ func (ct ConnectionType) String() string {
 //
 // Fetches the referenced Secret once and validates the connection type from its annotations.
 // Returns populated ConnectionInfo and the fetched Secret when a valid connection is found.
-// Returns empty ConnectionInfo and a nil Secret (no error) when no injection should be performed
-// (annotation absent, unknown type, or no type annotation on the Secret).
-// Returns an error when validation fails (e.g., Secret not found, API error).
+// Returns empty ConnectionInfo and a nil Secret (no error) when no annotation is present.
+// Returns an error when validation fails — including when the Secret is not found, the Secret
+// lacks a connection-type-protocol annotation, or the type value is not a supported one.
 //
 // The returned Secret is non-nil only when injection should proceed; callers must pass it to
 // injection helpers (GetURIValue, BuildS3URI) to avoid a second fetch.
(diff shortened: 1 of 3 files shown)


internal/webhook/connectionapi/utils.go
```diff
@@ -119,12 +119,14 @@ func ValidateConnectionAnnotation(ctx context.Context, apiReader client.Reader,
...
-		return ConnectionInfo{}, nil, nil
+		return ConnectionInfo{}, nil, fmt.Errorf(
+			"connection %q referenced by annotation %q must have a %q annotation on the Secret",
+			secretName, AnnotationConnections, AnnotationConnectionTypeProtocol)
 	}
 	if !isValid {
-		log.Info("unknown connection type, skipping injection", "connectionType", connType, "connection", secretName)
-		return ConnectionInfo{}, nil, nil
+		return ConnectionInfo{}, nil, fmt.Errorf(
+			"connection %q has unsupported connection type %q; supported types are s3, uri, oci",
```
maskarb, line 128: nitpick: If we ever added more types to the list of supported types, then we have to remember to update this string. It would be nice if this string was constructed from the list of supported types rather then being a static string.

maskarb (approved):
/lgtm

Looks good to me, just one non-blocking nitpick:

PR #881, opened 2026-07-15
feat: add publisher-path authorization rules and cross-tenant deny to gateway AuthPolicy

bartoszmajsak:
## Description

The gateway-level AuthPolicy only had `inference-access` rules that work for per-participant paths but don't cover [publisher paths](https://github.com/kserve/kserve/pull/5822) (`/publishers/<ns>/models/<name>/...`), which need model-level authorization against `serving.opendatahub.io/models`.

Adds `model-access-path` and `model-access-path-delegate` authorization rules for direct publisher-path traffic. Supports multi-segment model names (e.g. `facebook/opt-125m`) via `split('/models/')[1].split('/v1/')[0]` extraction. Model names containing a literal `/v1/` segment are not supported - same structural limitation as the kserve HTTPRoute template.

Adds a `deny-misrouted-model-header` rule that blocks requests where the routing identity (model header) and authorization identity (request path) diverge. Without this, a per-participant path with a valid model routing header would be authorized against the path (tenant-A) but routed via the header to a different tenant's backend (tenant-B). The deny rule uses Authorino's `patternMatching` with `predicate: "false"` for an immediate local 403 with no API server round-trip.

The model routing header name is read from the `inferenceservice-config` ConfigMap (`modelBasedRoutingHeaderName`), lowercased to match Authorino's key normalization, and validated as an HTTP token. The GatewayReconciler watches the ConfigMap's `ingress` key for changes.

BBR support (`resolvedPath` normalization for `/v1/` + header traffic) is a planned follow-up.

## How Has This Been Tested?

- Unit tests verify template rendering: 5 authorization rules, overrides
- E2E tests (13 publisher-path + 20 batch = 33 total) against a kind cluster with Kuadrant: publisher-path access, delegation, anti-spoofing, namespace collisions, cross-tenant deny, multi-segment model names
- Kind-based CI workflow triggered on auth-related file changes
- All existing batch AuthPolicy e2e tests pass

## Merge criteria:

- [ ] JIRA(s) are linked in the PR description
- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

## Summary by CodeRabbit

* **New Features**
  * Added configurable model-routing header support, including validation and a default header.
  * Added authorization for publisher model paths, multi-segment model names, and delegated access.
  * Added safeguards against cross-tenant model misrouting and identity spoofing.
[...]

PR #884, opened 2026-07-16
fix: extend deny-misrouted rule to cover /v1/ paths besides batch

bartoszmajsak:
## Description

Follow-up to #881. The `deny-misrouted-model-header` rule excluded all `/v1/` paths from its when-condition, which was intended for the BBR follow-up where `resolvedPath` normalization would handle `/v1/` + header authorization. Without that normalization in place, `/v1/` inference endpoints with a valid model routing header had no authorization rule covering them.

Restores the original batch-path bypass (`/v1/files`, `/v1/batches`) as the deny rule's exclusion instead of the broad `/v1/` exclusion. Batch paths remain authn-only as designed.

Also adds publisher-path test coverage for delegation edge cases (forwarded user without RBAC, forwarded user with only delegate access), authentication layer (no token on publisher path), and response header scoping (batch headers not injected on publisher paths).

Part of [RHOAIENG-69638](https://redhat.atlassian.net/browse/RHOAIENG-69638)

## Merge criteria:

[RHOAIENG-69638]: https://redhat.atlassian.net/browse/RHOAIENG-69638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Bug Fixes**
  - Strengthened authorization for publisher model routes to prevent cross-tenant access and block misrouted model headers.
  - Corrected deny vs auth-only handling across chat, files, batches, models, and health endpoints (including proper 403 behavior when applicable).
  - Ensured publisher-path requests without a token return `401` and do not include batch-specific response headers.
  - Fixed namespace-collision and delegated-access cases where required model permissions were missing.
- **Documentation**
  - Updated publisher-path authorization guidance with clarified request-flow behaviors and known limitations.
- **Tests**
  - Expanded end-to-end coverage for the new routing, header, delegation, and namespace collision scenarios.

### DIFF
--- a/internal/controller/test/e2e/publisher_path_authpolicy_test.go
+++ b/internal/controller/test/e2e/publisher_path_authpolicy_test.go
@@ -3,9 +3,15 @@
 package e2e
 
 import (
+	"context"
 	"fmt"
 	"net/http"
+	"strings"
 	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // publisherFixture holds per-test resources for publisher-path authorization tests.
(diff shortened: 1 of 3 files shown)


internal/controller/test/e2e/publisher_path_authpolicy_test.go
```diff
@@ -211,6 +217,109 @@ func TestEmptyModelHeaderAuthnOnlyFallthrough(t *testing.T) {
...
+		context.Background(),
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: ns}},
+		metav1.CreateOptions{},
+	)
+	if err != nil {
+		t.Logf("namespace v1 may already exist: %v", err)
+	}
+	t.Cleanup(func() {
+		_ = authEnv.clientset.CoreV1().Namespaces().Delete(context.Background(), ns, metav1.DeleteOptions{})
+	})
```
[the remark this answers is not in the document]
bartoszmajsak: It's test code, I think we can live with this trade-off considering emphemeral nature of test clusters?

internal/controller/test/e2e/publisher_path_authpolicy_test.go
```diff
@@ -211,6 +217,109 @@ func TestEmptyModelHeaderAuthnOnlyFallthrough(t *testing.T) {
...
+	headers := map[string]string{
+		"x-gateway-model-name": fmt.Sprintf("publishers/%s/models/echo-server", f.ns),
+	}
+	resp, _ := authEnv.gatewayGet(t, "/v1/chat/completions", f.noAccessToken, headers)
+	if resp.StatusCode != http.StatusForbidden {
+		t.Fatalf("expected 403 (/v1/ + model header should be denied without BBR), got %d", resp.StatusCode)
+	}
+}
+
+func TestV1PathWithGarbageModelHeaderDenied(t *testing.T) {
```
KillianGolds: **Tiny naming nit:** the name says `Denied` but the test asserts 200 (documenting the authn-only fallthrough, which the comment explains well). Maybe something like `TestV1PathWithNonPublisherHeaderAuthnOnly` at first glance I thought the test assertion was inverted, again a tiny nit.

internal/controller/test/e2e/publisher_path_authpolicy_test.go
```diff
@@ -211,6 +217,109 @@ func TestEmptyModelHeaderAuthnOnlyFallthrough(t *testing.T) {
...
+	}
+	resp, _ := authEnv.gatewayGet(t, "/v1/chat/completions", f.noAccessToken, headers)
+	if resp.StatusCode != http.StatusForbidden {
+		t.Fatalf("expected 403 (/v1/ + cross-tenant model header should be denied), got %d", resp.StatusCode)
+	}
+}
+
+// --- ns=v1 collision: user with RBAC for garbage SAR identity ---
+
+func TestV1NamespaceCollisionWithRBAC(t *testing.T) {
```
KillianGolds, line 273: Couple of small robustness thoughts, non-blocking:

- If the create at line 279 fails because ns `v1` already exists, the `t.Cleanup` at line 287 would still delete it right? Might be worth guarding the cleanup on this test having actually created the namespace.
- Back-to-back runs could flake: namespace deletion is async, so a rerun might hit `v1` in Terminating, the create fails, and `createServiceAccount` at line 292 then `Fatalf`s. Waiting for the namespace to be active (or tolerating Terminating with a retry) might help.

Also wondering if the first request at line 307 could assert the expected 200 rather than only logging it, just to pin down the documented authn-only behavior.

KillianGolds (commented):
The fix looks right to me. With the broad `/v1/` exclusion from #881, `/v1/` inference endpoints with a valid model routing header seemed to fall through to authn-only, so restoring the narrow batch bypass makes the deny rule fail-closed again. I checked the pre-existing boundary tests (no-header `/v1/`, empty header, participant path + header) and they all seem consistent with the new predicates.

A few small things I noticed, happy to be corrected on any of them:

1. **gofmt**: `internal/controller/test/e2e/publisher_path_authpolicy_test.go` fails `gofmt -l` for me (import ordering at lines 12-14, `corev1` should sort before `rbacv1`). I don't think lint CI catches it since golangci-lint doesn't run with the `e2e` build tag, but probably worth a quick fix to keep the tree gofmt-clean.

2. **Publisher path + mismatched header**: if I'm reading the deny rule's first predicate right (`authpolicy_llm_isvc_userdefined.yaml` lines 107-108), publisher paths are excluded entirely, so `/publishers/A/models/x/...` carrying header `publishers/B/models/y` would be authorized against the *path* (A/x) while routing could act on the *header*. Is routing on publisher paths fully path-pinned after the URLRewrite? If so this is a non-issue, just wanted to check whether the deny rule should also require header == path-derived identity on publisher paths until BBR normalization lands.

3. **Query strings**: as far as I can tell, Envoy's `request.path` includes the query string, so `GET /v1/files?purpose=batch` plus a model header would miss the batch exclusion (`authpolicy_llm_isvc_userdefined.yaml` lines 109-111) and get denied. That is over-deny and fail-closed, so probably fine in practice. But the same pattern on the pre-existing `x-maas-user`/`x-maas-groups` response header predicates (lines 181 and 187) might mean batch identity headers are not injected on `GET /v1/files?purpose=...`. Would `request.url_path` be the better attribute here? Pre-existing either way, so fine as a follow-up if it's a real issue.

internal/controller/test/e2e/publisher_path_authpolicy_test.go
bartoszmajsak: Good catch - the name is misleading. Will rename to TestV1PathWithNonPublisherHeaderAuthnOnly.

internal/controller/test/e2e/publisher_path_authpolicy_test.go
bartoszmajsak: I think coderabbit shared the same concern: https://github.com/opendatahub-io/odh-model-controller/pull/884#discussion_r3602667509

WRT back-to-back flakiness - valid concern. These currently run on ephemeral kind clusters (fresh per CI run), so the namespace collision doesn't happen in practice. If we move toward running these against shared/long-lived clusters, the retry logic would be worth adding. @coderabbitai capture this as a follow-up

> Assert 200 on first request instead of just logging 

Good point. The test logs the result of `/v1/chat/completions` without header but doesn't assert it. Should assert 200 to pin the documented behavior.

KillianGolds: Agh yes of course, makes sense, ephemeral kind clusters do sidestep it in practice, and capturing the retry logic as a follow-up for the shared-cluster case works for me. With the 200 assertion added I'm happy here.

internal/controller/test/e2e/publisher_path_authpolicy_test.go
KillianGolds: Thanks! Happy to resolve once the rename lands.

bartoszmajsak:
Thanks @KillianGolds, a bit more context:

> 1. **gofmt**: `internal/controller/test/e2e/publisher_path_authpolicy_test.go` fails `gofmt -l` for me (import ordering at lines 12-14, `corev1` should sort before `rbacv1`). I don't think lint CI catches it since golangci-lint doesn't run with the `e2e` build tag, but probably worth a quick fix to keep the tree gofmt-clean.

We should fix linter then, @coderabbitai open issue for this to follow-up.

> 2. **Publisher path + mismatched header**: if I'm reading the deny rule's first predicate right (`authpolicy_llm_isvc_userdefined.yaml` lines 107-108), publisher paths are excluded entirely, so `/publishers/A/models/x/...` carrying header `publishers/B/models/y` would be authorized against the _path_ (A/x) while routing could act on the _header_. Is routing on publisher paths fully path-pinned after the URLRewrite? If so this is a non-issue, just wanted to check whether the deny rule should also require header == path-derived identity on publisher paths until BBR normalization lands.

Good question. Publisher path HTTPRoutes use PathPrefix matching, which should take precedence over the header-only v1-catch-all-model-routing per Gateway API spec. Publisher path routing is path-pinned and the model-access-path rule authorizes against `request.path`, not the header, so a mismatched header doesn't affect the authorization decision. The BBR follow-up's normalization (descoped from this PR) aims to formalize this - it only fires for /v1/ paths, publisher paths pass through unchanged. I hope this makes sense.

> 3. **Query strings**: as far as I can tell, Envoy's `request.path` includes the query string, so `GET /v1/files?purpose=batch` plus a model header would miss the batch exclusion (`authpolicy_llm_isvc_userdefined.yaml` lines 109-111) and get denied. That is over-deny and fail-closed, so probably fine in practice. But the same pattern on the pre-existing `x-maas-user`/`x-maas-groups` response header predicates (lines 181 and 187) might mean batch identity headers are not injected on `GET /v1/files?purpose=...`. Would `request.url_path` be the better attribute here? Pre-existing either way, so fine as a follow-up if it's a real issue.

Sharp eye, it seems to be pre-existing (original template uses the same `request.path` predicates). I just checked and OpenAI files API does use query params (i.e. `?purpose=batch`), so it seems real. Happy to follow up with a switch to `request.url_path` - @coderabbitai do your job.

KillianGolds (approved):
Thanks for the quick turnaround - rename, import fix, the 200 assertion, and the gist link all addressed, and #887/#888 track the follow-ups. The path-pinning explanation on publisher routes resolved my main concern.

bartoszmajsak:
/override ci/prow/e2e-kserve-llmisvc

autoscaling failures seem unrelated

PR #901, opened 2026-07-30
fix: flaky auth posture and ReconcileError event tests

bartoszmajsak:
## Description

Eliminates intermittent failures in auth posture and ReconcileError event tests
caused by a race condition between the test assertions and controller-runtime's
async event recorder.

The real event recorder writes events through a buffered channel that drains to
the API server asynchronously. Tests that snapshot the event count right after
fixing a mismatch condition hit a race window - late-draining events from prior
reconcile runs increment the count after the snapshot, causing `Consistently`
assertions to fail unpredictably.

Switches the LLM test suite to `record.FakeRecorder`, so events land on a
channel the tests read directly. No API server write delay, no snapshot races.
The assertion pattern becomes straightforward drain-and-check - `Eventually`
for presence, `Consistently` for absence.

Net result: -27 lines, deterministic event assertions, zero flake window.

## How Has This Been Tested?

- Full LLM test suite passes (56/56 specs)
- Previously-flaky tests (`should stop emitting events when annotation is fixed
  to match` and `should exclude terminating members from mismatch detection`)
  run 3x consecutively with no failures
- `go vet` clean

## Merge criteria:

### DIFF
--- a/internal/controller/serving/llm/auth_posture_test.go
+++ b/internal/controller/serving/llm/auth_posture_test.go
@@ -17,13 +17,12 @@ limitations under the License.
 package llm_test
 
 import (
-	"context"
+	"strings"
 	"time"
 
 	kservev1alpha2 "github.com/kserve/kserve/pkg/apis/serving/v1alpha2"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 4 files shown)


pierDipi (approved):
/lgtm

